PR #16, opened 2020-09-28
Rework the logging infrastructure.

oshogbo:
For know we will use instrumentation function to log the entry and return of the
function. This reduce the size of code for over 10%.

### DIFF
--- a/src/modules/database/JUMP_LABEL/p_arch_jump_label_transform/p_arch_jump_label_transform.c
+++ b/src/modules/database/JUMP_LABEL/p_arch_jump_label_transform/p_arch_jump_label_transform.c
@@ -47,8 +47,6 @@ int p_arch_jump_label_transform_entry(struct kretprobe_instance *p_ri, struct pt
    struct module *p_module = NULL;
 
    p_debug_kprobe_log(
-          "Entering function <p_arch_jump_label_transform_entry>\n");
-   p_debug_kprobe_log(
           "p_arch_jump_label_transform_entry: comm[%s] Pid:%d\n",current->comm,current->pid);
 
    p_print_log(P_LKRG_INFO,
@@ -84,9 +82,6 @@ int p_arch_jump_label_transform_entry(struct kretprobe_instance *p_ri, struct pt
                   "[JUMP_LABEL] <Entry> I should never be here!\n");
    }
 
-   p_debug_kprobe_log(
-          "Leaving function <p_arch_jump_label_transform_entry>\n");
-
    /* A dump_stack() here will give a stack backtrace */
    return 0x0;
 }
@@ -96,10 +91,6 @@ int p_arch_jump_label_transform_ret(struct kretprobe_instance *ri, struct pt_reg
 
    unsigned int p_tmp,p_tmp2;
    unsigned char p_flag = 0x0;
-
-   p_debug_kprobe_log(
-          "Entering function <p_arch_jump_label_transform_ret>\n");
-
 
    switch (p_db.p_jump_label.p_state) {
 
@@ -188,8 +179,6 @@ int p_arch_jump_label_transform_ret(struct kretprobe_instance *ri, struct pt_reg
 
    p_db.p_jump_label.p_state = P_JUMP_LABEL_NONE;
 
-   p_debug_kprobe_log(
-          "Entering function <p_arch_jump_label_transform_ret>\n");
    return 0x0;
 }
 
@@ -198,38 +187,22 @@ int p_install_arch_jump_label_transform_hook(void) {
 
    int p_ret;
 
-// STRONG_DEBUG
-   p_debug_log(P_LKRG_STRONG_DBG,
-          "Entering function <p_install_arch_jump_label_transform_hook>\n");
-
    if ( (p_ret = register_kretprobe(&p_arch_jump_label_transform_kretprobe)) < 0) {
       p_print_log(P_LKRG_ERR, "[kretprobe] register_kretprobe() for <%s> failed! [err=%d]\n",
                   p_arch_jump_label_transform_kretprobe.kp.symbol_name,
                   p_ret);
-      goto p_install_arch_jump_label_transform_hook_out;
+      return p_ret;
    }
    p_print_log(P_LKRG_INFO, "Planted [kretprobe] <%s> at: 0x%lx\n",
                p_arch_jump_label_transform_kretprobe.kp.symbol_name,
                (unsigned long)p_arch_jump_label_transform_kretprobe.kp.addr);
    p_arch_jump_label_transform_kretprobe_state = 0x1;
 
-//   p_ret = 0x0; <- should be 0x0 anyway...
-
-p_install_arch_jump_label_transform_hook_out:
-
-// STRONG_DEBUG
-   p_debug_log(P_LKRG_STRONG_DBG,
-          "Leaving function <p_install_arch_jump_label_transform_hook> (p_ret => %d)\n",p_ret);
-
    return p_ret;
 }
 
 
 void p_uninstall_arch_jump_label_transform_hook(void) {
-
-// STRONG_DEBUG
-   p_debug_log(P_LKRG_STRONG_DBG,
-          "Entering function <p_uninstall_arch_jump_label_transform_hook>\n");
 
    if (!p_arch_jump_label_transform_kretprobe_state) {
       p_print_log(P_LKRG_INFO, "[kretprobe] <%s> at 0x%lx is NOT installed\n",
@@ -243,8 +216,4 @@ void p_uninstall_arch_jump_label_transform_hook(void) {
                   p_arch_jump_label_transform_kretprobe.nmissed);
       p_arch_jump_label_transform_kretprobe_state = 0x0;
    }
-
-// STRONG_DEBUG
-   p_debug_log(P_LKRG_STRONG_DBG,
-          "Leaving function <p_uninstall_arch_jump_label_transform_hook>\n");
 }
--- a/src/modules/database/JUMP_LABEL/p_arch_jump_label_transform_apply/p_arch_jump_label_transform_apply.c
+++ b/src/modules/database/JUMP_LABEL/p_arch_jump_label_transform_apply/p_arch_jump_label_transform_apply.c
@@ -54,8 +54,6 @@ int p_arch_jump_label_transform_apply_entry(struct kretprobe_instance *p_ri, str
    p_text_poke_loc *p_tmp;
 
    p_debug_kprobe_log(
-          "Entering function <p_arch_jump_label_transform_apply_entry>\n");
-   p_debug_kprobe_log(
           "p_arch_jump_label_transform_apply_entry: comm[%s] Pid:%d\n",current->comm,current->pid);
 
    p_print_log(P_LKRG_INFO,
@@ -86,9 +84,6 @@ int p_arch_jump_label_transform_apply_entry(struct kretprobe_instance *p_ri, str
       }
    }
 
-   p_debug_kprobe_log(
-          "Leaving function <p_arch_jump_label_transform_apply_entry>\n");
-
    /* A dump_stack() here will give a stack backtrace */
    return 0x0;
 }
@@ -103,9 +98,6 @@ int p_arch_jump_label_transform_apply_ret(struct kretprobe_instance *ri, struct 
    unsigned int p_text = 0x0;
    unsigned int p_mod = 0x0;
 //   DECLARE_BITMAP(p_mod_mask, p_db.p_module_list_nr);
-
-   p_debug_kprobe_log(
-          "Entering function <p_arch_jump_label_transform_apply_ret>\n");
 
    bitmap_zero(p_db.p_jump_label.p_mod_mask, p_db.p_module_list_nr);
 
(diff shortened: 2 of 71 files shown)


src/modules/database/JUMP_LABEL/p_arch_jump_label_transform/p_arch_jump_label_transform.c
```diff
@@ -194,39 +183,24 @@ int p_install_arch_jump_label_transform_hook(void) {
...
-
-p_install_arch_jump_label_transform_hook_out:
-
-// STRONG_DEBUG
-   p_debug_log(P_LKRG_STRONG_DBG,
-          "Leaving function <p_install_arch_jump_label_transform_hook> (p_ret => %d)\n",p_ret);
-
+   BUG_ON(p_ret != 0x00);
    return p_ret;
 }
```
solardiz, line 199: Per my reading, this `BUG_ON` would be triggered if `register_kretprobe()` returns a value greater than 0. Perhaps that's impossible, but I don't like the `BUG_ON`. We can instead change the check of `register_kretprobe()` return value from:

```c
   if ( (p_ret = register_kretprobe(&p_arch_jump_label_transform_kretprobe)) < 0) {
```

to:

```c
   if ((p_ret = register_kretprobe(&p_arch_jump_label_transform_kretprobe))) {
```

so that we'll treat any non-zero return as an error right there. I've just checked that all (two) calls of `p_install_arch_jump_label_transform_hook()` merely care if the return value is 0 or not, but to be on the safe side (not potentially returning a positive value on error) and to simplify the code I think we should also drop `p_ret` (including the `int p_ret;` at the start of the function) and simply `return -1;` in there.

And then we should simply `return 0;` at the end of the function.

src/modules/database/JUMP_LABEL/p_arch_jump_label_transform_apply/p_arch_jump_label_transform_apply.c
```diff
@@ -241,39 +226,28 @@ int p_install_arch_jump_label_transform_apply_hook(void) {
...
-
-p_install_arch_jump_label_transform_apply_hook_out:
-
-// STRONG_DEBUG
-   p_debug_log(P_LKRG_STRONG_DBG,
-          "Leaving function <p_install_arch_jump_label_transform_apply_hook> (p_ret => %d)\n",p_ret);
+   BUG_ON(p_ret != 0x00);
 
    return p_ret;
 }
```
solardiz, line 246: Ditto, let's drop `p_ret`. There's only one call to `p_install_arch_jump_label_transform_apply_hook()`, which only cares for 0 or not.

src/modules/database/arch/x86/p_switch_idt/p_switch_idt.c
```diff
@@ -77,39 +65,24 @@ int p_install_switch_idt_hook(void) {
...
-
-p_install_switch_idt_hook_out:
-
-// STRONG_DEBUG
-   p_debug_log(P_LKRG_STRONG_DBG,
-          "Leaving function <p_install_switch_idt_hook> (p_ret => %d)\n",p_ret);
+   BUG_ON(p_ret != 0x00);
 
    return p_ret;
 }
```
solardiz, line 81: Ditto, let's drop `p_ret`. There's only one call to `p_install_switch_idt_hook()`, which only cares for 0 or not.

src/modules/database/p_database.c
```diff
@@ -432,21 +364,11 @@ int p_create_database(void) {
          "CREATING DATABASE ERROR: HASH FROM _STEXT!\n");
       p_ret = P_LKRG_GENERAL_ERROR;
       p_text_section_unlock();
-      goto p_create_database_out;
    } else {
       p_ret = P_LKRG_SUCCESS;
    }
    p_text_section_unlock();
```
solardiz, line 370: Looks like you've introduced a double-unlock bug there in one of the code paths. Fixing it will probably also let you simplify the code and get rid of `p_ret` in this function.

src/modules/exploit_detection/p_exploit_detection.c
```diff
@@ -1914,72 +1766,67 @@ int p_ed_pcfi_validate_sp(struct task_struct *p_task, struct p_ed_process *p_ori
...
 
-p_ed_pcfi_validate_sp_out:
-
-// STRONG_DEBUG - can be way too noisy
-//   p_debug_log(P_LKRG_STRONG_DBG,
-//          "Leaving function <p_ed_pcfi_validate_sp> (return => %d)\n",p_not_valid);
-
-   return p_not_valid;
+   return 0x0;
 }
```
solardiz, line 1830: The changes to `p_ed_pcfi_validate_sp()` took me a while to review with the indentation change. I find them reasonable, but we should note somewhere that this isn't just a cosmetic change, but a user-visible one: when the stack frame is wrong in more than one way, the function will now complain in the logs about one of those only. It used to complain about all. Adam might want to weigh in on accepting this user-visible change or not. I am fine with it.

src/modules/exploit_detection/syscalls/p_install.c
```diff
@@ -66,23 +62,13 @@ int p_install_hook(struct kretprobe *kretprobe, char *state, int p_isra) {
...
    (*state)++;
 
-p_install_hook_out:
-
-   p_debug_log(P_LKRG_STRONG_DBG,
-          "Leaving function <p_install_hook> for %s (p_ret => %d)\n",
-          p_name, p_ret);
-
    return p_ret;
 }
```
solardiz, line 66: Unrelated to changes of this PR, but we have the same problem of potentially returning a value greater than zero if `register_kretprobe()` returns it. So the new `BUG_ON` in other similar functions were also inconsistent with how we ignore that problem in this one. And maybe we should change this one to only return -1 or 0 for consistency with those (after my proposed changes to them). We can't fully drop `p_ret` from here as we check it for -22, but we can choose to never propagate it further like we wouldn't in those functions.

src/modules/ksyms/p_resolve_ksym.c
```diff
@@ -81,14 +81,10 @@ long get_kallsyms_address(void) {
 
 #if (LINUX_VERSION_CODE >= KERNEL_VERSION(5,7,0))
    struct kprobe p_kprobe;
+   int p_ret = P_LKRG_SUCCESS;
```
solardiz, line 84: Per my reading, with further changes we never make use of the initial value of `p_ret` in this function. We might nevertheless keep the initialization as defensive programming against future changes. I have no strong preference on this.

Alternatively, we can simplify the conditional compilation logic so that we check for 5.7.0+ just once and then both define the local variables and use them in the same piece. In fact, we probably don't need separate `p_ret` and `p_tmp`, as with these changes they serve the same purpose that Adam normally uses the name `p_tmp` for.

solardiz (changes requested):
I've skimmed through these changes. They mostly look good to me. (We had discussed them conceptually via e-mail a few months ago, so I don't repeat that here.)

Please fix at least the double-unlock bug. While at it, please also fix grammar in the commit message.

solardiz:
Suggested commit message (more than just the grammar fixes I was going to suggest at first):

```
Rework debug logging of entry and return from functions

Use instrumentation.  This reduces source code size by over 10%.

```

Adam-pi3:
Sorry for delay, I will review this PR soon. In the mean time I want to mention that we should not use BUG_ON at all. Various kernel configuration might be causing kprobes to fail. That's why LKRG prints appropriate message during each step of initialization. Some of the kprobes failures goes into ISRA  / CONSTPROP path some of them don't (maybe we should always go to this logic). 
Nevertheless, please avoid using BUG_ON since I don't think in LKRG case it's useful.

solardiz:
@oshogbo Unless you're happy rebasing these changes over and over, I think you should update them per my feedback (which is consistent with Adam's - both of us were unhappy with `BUG_ON`), make them apply again, and we should get them merged sooner rather than later. Thanks.

src/modules/database/p_database.c
oshogbo: Nice catch!

src/modules/exploit_detection/syscalls/p_install.c
oshogbo: This is something that @Adam-pi3 should look into.
I'm fine with changing this.

src/modules/ksyms/p_resolve_ksym.c
oshogbo: Yea I changed p_tmp to p_ret.

src/modules/database/JUMP_LABEL/p_arch_jump_label_transform/p_arch_jump_label_transform.c
solardiz: I see you're simply dropping the `BUG_ON` in the force-push now, but my suggestion was more elaborate, as above? Are you going to implement that or do you suggest we first merge your changes and then clean them up separately?

oshogbo: Yea, I would like to change this separately with more detailed commit msg.

src/modules/exploit_detection/p_exploit_detection.c
```diff
@@ -1922,72 +1774,67 @@ int p_ed_pcfi_validate_sp(struct task_struct *p_task, struct p_ed_process *p_ori
...
+      p_print_log(P_LKRG_INFO,
+             "process[0x%lx | %d | %s] vs orig[0x%lx | %d | %s]\n",
+             (unsigned long)p_task,
+             task_pid_nr(p_task),
+             p_task->comm,
+             (unsigned long)p_orig->p_ed_task.p_task,
+             p_orig->p_ed_task.p_pid,
+             p_orig->p_ed_task.p_comm);
+      return 0x1;
+   }
```
Adam-pi3, line 1799: Up to this I think is fine.

src/modules/exploit_detection/p_exploit_detection.c
Adam-pi3: Yes, current version only complain on first error. We would want to have in the logs all violations not just first one. I would prefer to keep it how it was since it's more valuable.

src/modules/database/p_database.c
```diff
@@ -401,15 +335,13 @@ int p_create_database(void) {
    if (p_install_arch_jump_label_transform_hook()) {
       p_print_log(P_LKRG_ERR,
              "ERROR: Can't hook arch_jump_label_transform function :(\n");
-      p_ret = P_LKRG_GENERAL_ERROR;
-      goto p_create_database_out;
+      return P_LKRG_GENERAL_ERROR;
```
Adam-pi3, line 338: By mistake, you've removed this return. Current version doesn't exit p_create_database() function in case of this error. That's wrong.

src/modules/database/p_database.c
```diff
@@ -401,15 +335,13 @@ int p_create_database(void) {
...
    }
 
    if (p_install_arch_jump_label_transform_static_hook()) {
       p_print_log(P_LKRG_ERR,
              "ERROR: Can't hook arch_jump_label_transform_static function :(\n");
-      p_ret = P_LKRG_GENERAL_ERROR;
-      goto p_create_database_out;
+      return P_LKRG_GENERAL_ERROR;
```
Adam-pi3, line 344: Same here

src/modules/database/p_database.c
Adam-pi3: I think it is still not fixed, is it?

src/modules/exploit_detection/syscalls/p_install.c
Adam-pi3: I think for now it's OK.

src/modules/kmod/p_kmod_notifier.c
```diff
@@ -2,7 +2,7 @@
  * pi3's Linux kernel Runtime Guard
  *
  * Component:
- *  - Kernel's modules module notifier 
+ *  - Kernel's modules module notifier
```
Adam-pi3, line 5: I think it's deleted comment as well - at least that's how it looks in my browser

src/modules/print_log/p_lkrg_debug_log.c
```diff
@@ -0,0 +1,198 @@
...
+#if LINUX_VERSION_CODE < KERNEL_VERSION(4,14,0)
+   P_LKRG_DEBUG_RULE_KPROBE(p_switch_idt),
+#endif
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,3,0)
+   P_LKRG_DEBUG_RULE_KPROBE(p_arch_jump_label_transform),
+   P_LKRG_DEBUG_RULE_KPROBE(p_arch_jump_label_transform_apply),
+#endif
+#endif
+
+   // Disable to noisy.
```
Adam-pi3, line 163: typo s/to/too/g

src/modules/print_log/p_lkrg_debug_log.c
```diff
@@ -0,0 +1,198 @@
...
+void __cyg_profile_func_enter(void *this_fn, void *call_site)
+__attribute__((no_instrument_function));
+void __cyg_profile_func_exit(void *this_fn, void *call_site)
+__attribute__((no_instrument_function));
+
+#ifdef P_LKRG_DEBUG_BUILD
+static struct p_addr_name {
+    uintptr_t	addr;
+    const char *name;
+} p_addr_name_array[] = {
```
Adam-pi3, line 35: I think not all the functions are on the list here. E.g. p_init_page_attr is missing.

Adam-pi3 (changes requested):

Issue #4, opened 2020-08-11
Make UMH blocking support pathnames on CPA-protected pages


solardiz:
Currently LKRG blocks disallowed usermodehelper pathnames by temporarily setting the pages containing those to read-write and overwriting the pathnames with all-slashes. This is incompatible with recent kernels' CPA. We need to come up with a different way.

Reference: https://www.openwall.com/lists/lkrg-users/2020/07/19/1

PR #18, opened 2020-10-17
Fix building on Linux 5.9

0xC0ncord:
1. `file` is no longer a field in the `subprocess_info` struct, so we just don't use it.
2. `__module_address` and `__module_text_address` are no longer exported, so we need to resolve these symbols at runtime.
3. `put_seccomp_filter` has been removed, so we need to use `__put_seccomp_filter` instead and modify calls to this function.

My understanding on LKRG's code is very limited, so please check this for errors. With these fixes I managed to build LKRG against 5.9.0 and it reports successful initialization when it is loaded.

### DIFF
--- a/src/modules/database/JUMP_LABEL/p_arch_jump_label_transform/p_arch_jump_label_transform.c
+++ b/src/modules/database/JUMP_LABEL/p_arch_jump_label_transform/p_arch_jump_label_transform.c
@@ -65,7 +65,11 @@ int p_arch_jump_label_transform_entry(struct kretprobe_instance *p_ri, struct pt
        * OK, *_JUMP_LABEL tries to modify kernel core .text section
        */
       p_db.p_jump_label.p_state = P_JUMP_LABEL_CORE_TEXT;
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,9,0)
+   } else if ( (p_module = P_SYM(p_module_text_address)(p_addr)) != NULL) {
+#else
    } else if ( (p_module = __module_text_address(p_addr)) != NULL) {
+#endif
       /*
        * OK, *_JUMP_LABEL tries to modify some module's .text section
        */
--- a/src/modules/database/JUMP_LABEL/p_arch_jump_label_transform_apply/p_arch_jump_label_transform_apply.c
+++ b/src/modules/database/JUMP_LABEL/p_arch_jump_label_transform_apply/p_arch_jump_label_transform_apply.c
@@ -115,7 +115,11 @@ int p_arch_jump_label_transform_apply_ret(struct kretprobe_instance *ri, struct 
 
          p_text++;
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,9,0)
+      } else if ( (p_module = P_SYM(p_module_text_address)(p_jl_batch_addr[p_cnt])) != NULL) {
+#else
       } else if ( (p_module = __module_text_address(p_jl_batch_addr[p_cnt])) != NULL) {
+#endif
 
          for (p_tmp = 0x0; p_tmp < p_db.p_module_list_nr; p_tmp++) {
             if (p_db.p_module_list_array[p_tmp].p_mod == p_module) {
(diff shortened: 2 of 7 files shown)


solardiz:
Thanks. Ideally, @Adam-pi3 should review this. Meanwhile, I skimmed it and have some quick comments for the 3 fixes/commits:

1. Simply skipping that code will likely lead to false positives: where we previously allowed "executing file from memory" we wouldn't do so anymore. Maybe we should instead drop the check of the `file` field only, or add "none" to `UMH_LIST`. I don't know when "executing file from memory" is needed, so I don't know what exactly will break. (We should also rework the whole UMH handling to address #4. When we do, this code we're patching here might not even exist in LKRG anymore.)

2. Your changes look correct to me at first glance. In fact, longer-term we might want to switch to always looking up those symbols dynamically in order to reduce source code size and increase code coverage of testing on one system, but for now it looks appropriate to add the special case for 5.9+ only so that we're confident we're not invalidating the testing of LKRG previously done on older kernels.

3. Your changes look correct to me assuming that your understanding of changes in the kernel itself is correct (I didn't cross-check against those).

0xC0ncord:
Thank you for your feedback.

1. I wasn't sure about the implementation of this "fix" either, so after secondary review let's see what we can do about that. The exact commit that changed this is [here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/include/linux/umh.h?id=55e6074e3fa67e1fb9ec140904db7e6cae6eda4b).
2. These changes were made in [here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/kernel/module.c?id=3fe1e56d0e68b623dd62d8d38265d2a052e7e185) and [here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/kernel/module.c?id=34e64705ad415ed7a816e60ef62b42fe6d1729d9).
3. These changes were made in [this commit](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/kernel/seccomp.c?id=3a15fb6ed92cb32b0a83f406aa4a96f28c9adbc3).

solardiz:
There are several recent changes to UMH: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/log/kernel/umh.c

Relevant one is: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/kernel/umh.c?id=55e6074e3fa67e1fb9ec140904db7e6cae6eda4b ("umh: Stop calling do_execve_file"). With this, maybe your change is OK, after all, and wouldn't result in false positives.

> Maybe we should instead drop the check of the `file` field only, or add "none" to `UMH_LIST`.

This would weaken LKRG's UMH hardening, letting exploits UMH-execute "none" in the current directory if they bypass pCFI.

0xC0ncord:
> Relevant one is: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/kernel/umh.c?id=55e6074e3fa67e1fb9ec140904db7e6cae6eda4b ("umh: Stop calling do_execve_file"). With this, maybe your change is OK, after all, and wouldn't result in false positives.

Ah ha. That makes sense.

Adam-pi3:
Hi,

Thanks for the PR addressing these issues. However, I think that UMH fix is not fully complete. I've reviewed last kernel changes and it looks like they removed the call_usermodehelper_setup_file() function. To be able to support that case, blob_to_mnt() function is created:
https://patchwork.kernel.org/project/linux-security-module/patch/87d05mf0j9.fsf_-_@x220.int.ebiederm.org/

I think we can merge your commits as a temporary fix, however, if bpfilter dynamically loads program, it can generate FP:
https://elixir.bootlin.com/linux/latest/source/net/ipv4/bpfilter/sockopt.c#L72
https://elixir.bootlin.com/linux/latest/source/kernel/usermode_driver.c#L12

Other 2 commits looks fine, thanks!

0xC0ncord:
> Other 2 commits looks fine, thanks!

Sounds good. Would you rather I revert the offending 1st commit or just drop it from my branch?

PR #20, opened 2020-10-21
Minor fixes.

oshogbo:


src/modules/database/JUMP_LABEL/p_arch_jump_label_transform_apply/p_arch_jump_label_transform_apply.c
```diff
@@ -244,7 +244,7 @@ int p_install_arch_jump_label_transform_apply_hook(void) {
                (unsigned long)p_arch_jump_label_transform_apply_kretprobe.kp.addr);
    p_arch_jump_label_transform_apply_kretprobe_state = 0x1;
 
-   return p_ret;
+   return 0x0;
```
Adam-pi3, line 247: I wonder if we should return P_LKRG_SUCCESS instead of hardcoded 0x0. But that's just minor think.

src/modules/exploit_detection/p_exploit_detection.c
```diff
@@ -1774,11 +1775,11 @@ int p_ed_pcfi_validate_sp(struct task_struct *p_task, struct p_ed_process *p_ori
       p_print_log(P_LKRG_INFO, "[base:0x%lx c:0x%lx]\n",
              p_stack,
              p_sp);
-      return 0x1;
+      p_not_valid = 0x01;
```
Adam-pi3, line 1778: 0x1 instead of 0x01 ? However, @solardiz doesn't like that as well and I think he recommends to use just 1

src/modules/exploit_detection/p_exploit_detection.c
```diff
@@ -1758,6 +1758,7 @@ int p_ed_pcfi_validate_sp(struct task_struct *p_task, struct p_ed_process *p_ori
 
    unsigned long p_stack = (p_orig) ? (unsigned long) p_orig->p_ed_task.p_stack : 0x0;
    register unsigned long p_stack_offset;
+   char p_not_valid = 0x0;
```
Adam-pi3, line 1761: P_LKRG_SUCCESS?

src/modules/database/JUMP_LABEL/p_arch_jump_label_transform/p_arch_jump_label_transform.c
```diff
@@ -187,7 +187,7 @@ int p_install_arch_jump_label_transform_hook(void) {
 
    int p_ret;
 
-   if ( (p_ret = register_kretprobe(&p_arch_jump_label_transform_kretprobe)) < 0) {
+   if ( (p_ret = register_kretprobe(&p_arch_jump_label_transform_kretprobe)) != 0) {
       p_print_log(P_LKRG_ERR, "[kretprobe] register_kretprobe() for <%s> failed! [err=%d]\n",
                   p_arch_jump_label_transform_kretprobe.kp.symbol_name,
                   p_ret);
```
solardiz, line 193: As I wrote in my review of #16:

```
I've just checked that all (two) calls of p_install_arch_jump_label_transform_hook() merely care if the return value is 0 or not, but to be on the safe side (not potentially returning a positive value on error) and to simplify the code I think we should also drop p_ret (including the int p_ret; at the start of the function) and simply return -1; in there.

And then we should simply return 0; at the end of the function.
```

I now see that we need `p_ret` for the `p_print_log`, so OK to keep the variable (maybe renaming it to `p_tmp`?) but maybe let's return explicit -1 and 0 from this function?

src/modules/database/JUMP_LABEL/p_arch_jump_label_transform_apply/p_arch_jump_label_transform_apply.c
solardiz: More importantly, we should consistently use explicit -1 and 0 returns, as I also suggest for `p_install_arch_jump_label_transform_hook`.

src/modules/database/arch/x86/p_switch_idt/p_switch_idt.c
```diff
@@ -76,7 +76,7 @@ int p_install_switch_idt_hook(void) {
                (unsigned long)p_switch_idt_kretprobe.kp.addr);
    p_switch_idt_kretprobe_state = 0x1;
 
-   return p_ret;
+   return 0x0;
```
solardiz, line 79: Ditto. There's still a `return p_ret;` inside the `if` block, which I think should be `return -1;`.

src/modules/exploit_detection/p_exploit_detection.c
solardiz: We should eventually drop all of those 0x things (except where the numbers are genuinely in a power of 2 based radix), and just use 0 and 1.

src/modules/exploit_detection/p_exploit_detection.c
solardiz: Probably if we call it `p_ret` or such. Also, as discussed via e-mail before, let's consistently use `int` (and not `char`) for Booleans and status codes. In this case, the function return type is (correctly) an `int`, so it's weird to use `char` internally.

However, we currently also have a `char p_not_valid = 0x0;` in `p_ed_enforce_pcfi`, which isn't yet modified in this PR, so let's keep things consistent with that one for now.

solardiz (approved):
I don't mind merging this as-is, but maybe revise the install hook functions as I suggested?

solardiz:
I think that our commit messages should ideally state _what_ is being modified - subsystem, function, or/and kind of logic, etc. - not only how. Sure the "_what_" is usually seen from looking at the diff, but I think more self-contained commit messages would be a plus. For example:

```
p_ed_pcfi_validate_sp: Validate all conditions before returning

We would want to have in the logs all violations, not just the first one.

```

```
Require a zero return from register_kretprobe(), not just non-negative

```

@oshogbo Please let us know if you're going to make changes to this PR per our suggestions. Thanks!

oshogbo:
Yes, I will fix those.

src/modules/database/JUMP_LABEL/p_arch_jump_label_transform/p_arch_jump_label_transform.c
oshogbo: I'm not sure if you say explicite -1 and 0 do you mean P_LKRG_GENERAL_ERROR and P_LKRG_SUCESS?

solardiz: I meant literally -1 and 0. I think LKRG's use of those named error codes is currently unreasonable. I just ran some greps, and the only one LKRG ever checks for (in just one place) is `P_LKRG_KMOD_DUMP_RACE`. I don't know why it cares to distinguish the rest at all, and why it bothers with doing so given that it does that in very few places. I think it'd be cleaner with explicit -1 and 0, and us dropping those names (except for `P_LKRG_KMOD_DUMP_RACE`, if that one is needed in that one place).

solardiz: I don't mind using P_LKRG_GENERAL_ERROR and P_LKRG_SUCESS in this code for now if that keeps it more consistent with what it used to be and with nearby code. We can clean that up separately.

solardiz: You still have `return p_ret;` in this function, unlike in two others now modified to use explicit return codes.

solardiz:
FWIW, I've confirmed that (on a certain build) `p_lkrg.o` remains unchanged with the "lkrg: drop the 0x prefix." commit. (Of course, with our various `#ifdef`s one build doesn't achieve 100% code coverage.)

`p_lkrg.ko` changes slightly - per `modinfo`, `srcversion` changes. I think we should simply be using `p_lkrg.o` for checks like this (when we make more coding style only changes in the future).

oshogbo:
Oh, that's an excellent idea. I didn't think about that. Thank you for the hint!

PR #23, opened 2020-11-04
umh: Kill process using the proper SIGKILL signal.

oshogbo:
When we detect an invalid binary, we forbid it execution by changing the
subprocess execution path with slashes.
On some kernels, this may not work because we have to map the memory page as
writable. At this moment of execution, we can stop executing the process using
the froce_sig to send signals. The send_sig_info will not arrive at the right
time.

This should fix [Issue 4](https://github.com/openwall/lkrg/issues/4)

### DIFF
--- a/src/modules/exploit_detection/p_exploit_detection.h
+++ b/src/modules/exploit_detection/p_exploit_detection.h
@@ -357,6 +357,12 @@ static inline unsigned long get_random_long(void) {
 }
 #endif
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,3,0)
+#define p_force_sig(sig) force_sig((sig))
+#else
+#define p_force_sig(sig) force_sig((sig), current)
+#endif
+
 /* Task integrity lock API */
 static inline void p_tasks_read_lock_raw(rwlock_t *p_arg) {
 
--- a/src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
+++ b/src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
@@ -52,7 +52,6 @@
    }
 
 char p_call_usermodehelper_kretprobe_state = 0;
-static DEFINE_SPINLOCK(p_umh_attr_lock);
 
 static struct kretprobe p_call_usermodehelper_kretprobe = {
 #if LINUX_VERSION_CODE < KERNEL_VERSION(4,3,0)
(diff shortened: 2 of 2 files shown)


solardiz:
This failed 2 out of 3 CI builds:

```
/home/travis/build/openwall/lkrg/src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c: In function ‘p_call_usermodehelper_entry’:
/home/travis/build/openwall/lkrg/src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c:175:18: error: too many arguments to function ‘force_sig’
                  force_sig(SIGKILL, current);
                  ^~~~~~~~~
In file included from ./include/net/scm.h:11:0,
                 from ./include/linux/netlink.h:9,
                 from ./include/uapi/linux/neighbour.h:6,
                 from ./include/linux/netdevice.h:46,
                 from /home/travis/build/openwall/lkrg/src/modules/exploit_detection/syscalls/p_call_usermodehelper/../../../../p_lkrg_main.h:39,
                 from /home/travis/build/openwall/lkrg/src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c:22:
./include/linux/sched/signal.h:341:13: note: declared here
 extern void force_sig(int);
             ^~~~~~~~~
```

More importantly, I doubt this code always run in (relevant, or any) process context, or does it?

oshogbo:
Oh I didn't test it on the old kernels.
For the new kernels it works fine and the correct process context killed. Do you have any particular context in mind which may not be the correct one?

solardiz:
> Do you have any particular context in mind which may not be the correct one?

No, but there are lots of places in the kernel where UMH can be invoked from, and I guess these vary in terms of (not) having a process context.

Personally, I think we should simply drop UMH hardening from LKRG, but Adam disagrees. So I'd rather leave this to Adam.

Adam-pi3:
That specific hook for UMH should always run with proper process context (to be more specific, it should run under UMH WQ pool threads).
I've verified that all the kernels (since 3.x, didn't look at 2.6.xx) up to the latest one, will fail do_execve() during copy of argv* if there are pending forcible signals which are unmaskable. It looks OK and we need to test it. Also, the latest kernel (since 5.3+) modified force_sig() API and do not allow to provide task_struct pointer to whom signal is going to be sent. However, we are lucky since we always send it to the 'current' process. Modified API is setting 'current' under-the-hood so we should be fine using the same API.
I've synced with @oshogbo offline and he is going to validate that.

oshogbo:
Ok I verify on both kernels and it works perfect.

src/modules/exploit_detection/p_exploit_detection.h
```diff
@@ -357,6 +357,12 @@ static inline unsigned long get_random_long(void) {
 }
 #endif
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,1,0)
+#define p_force_sig(sig, cur) force_sig((sig))
+#else
+#define p_force_sig(sig, cur) force_sig((sig), (cur))
+#endif
```
solardiz, line 364: Since we're not always able to pass `cur` to the underlying API, we shouldn't accept it in ours either. We should have:

```c
#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,1,0)
#define p_force_sig(sig) force_sig((sig))
#else
#define p_force_sig(sig) force_sig((sig), current)
#endif
```

and use it accordingly.

oshogbo: Ok.

solardiz (approved):
Looks good to me now. Adam, is this ready to merge or do you want to test it more?

Adam-pi3:
I want to run a few more tests before merging - sorry for delay. I will try to do it over the weekend.

src/modules/exploit_detection/p_exploit_detection.h
```diff
@@ -357,6 +357,12 @@ static inline unsigned long get_random_long(void) {
 }
 #endif
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,1,0)
```
Adam-pi3, line 360: I think it is wrong. It should be >= 5.3 or < 5.2. Linux kernel 5.1 is still using old interface

oshogbo: Ugh, I was pretty sure I have had verified that.
NVM. I pushed fix.

Adam-pi3:
Verified.

solardiz:
Adam, I think we should be consistently using "rebase and merge" for PRs, unless there are reasons not to in a given case. To avoid inadvertent creation of merge commits going forward, I've just disabled that functionality for this repo.

PR #28, opened 2020-11-19
UMH: Allow all the pathnames listed in p_umh_global

oshogbo:
UMH contains a list of files that are allowed to execute.
LKRG during the load checks if the file exists; if so, the file is added to
the whitelist. The umh_preexists_files allows controlling this check.
By disabling lkrg.umh_prexists_files sysctl, all the files from the
lists will be allowed no matter if they were created after LKRG was loaded.

### DIFF
--- a/src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
+++ b/src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
@@ -21,37 +21,6 @@
 
 #include "../../../../p_lkrg_main.h"
 
-#define UMH_LIST                                                \
-   {                                                            \
-       UMH_UNIFIFW                                              \
-       UMH_LNET_UPCALL                                          \
-       UMH_OSD_LOGIN                                            \
-       { "/bin/false", false },                                 \
-       { "/bin/true", false },                                  \
-       { "/etc/acpi/events/RadioPower.sh", false },             \
-       { "/etc/acpi/wireless-rtl-ac-dc-power.sh", false },      \
-       { "/lib/systemd/systemd-cgroups-agent", false },         \
-       { "/sbin/bridge-stp", false },                           \
-       { "/sbin/critical_overtemp", false },                    \
-       { "/sbin/drbdadm", false },                              \
-       { "/sbin/hotplug", false },                              \
-       { "/sbin/modprobe", false },                             \
-       { "/sbin/nfs_cache_getent", false },                     \
-       { "/sbin/nfsd-recall-failed", false },                   \
-       { "/sbin/nfsdcltrack", false },                          \
-       { "/sbin/ocfs2_hb_ctl", false },                         \
-       { "/sbin/pnpbios", false },                              \
-       { "/sbin/poweroff", false },                             \
-       { "/sbin/request-key", false },                          \
-       { "/sbin/tomoyo-init", false },                          \
-       { "/sbin/v86d", false },                                 \
-       { "/system/bin/start", false },                          \
-       { "/usr/lib/systemd/systemd-cgroups-agent", false },     \
-       { "/usr/lib/systemd/systemd-coredump", false },          \
-       { "/usr/sbin/eppfpga", false },                          \
-       { "/usr/share/apport/apport", false }                    \
-   }
-
 char p_call_usermodehelper_kretprobe_state = 0;
 
 static struct kretprobe p_call_usermodehelper_kretprobe = {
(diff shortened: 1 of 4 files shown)


solardiz (changes requested):
Thank you for working on this, but:

The sysctl addition is unacceptable. We should make this change unconditionally. Part of the reason for it is to simplify LKRG. This PR as it is complicates LKRG, so is counterproductive.

solardiz:
Re-implementing this without the sysctl will enable simplifications not seen here: the list of allowed pathnames should become `const` and be used directly, without copying it onto stack. The Booleans should be dropped from it.

oshogbo:
I see I was pretty sure that @Adam-pi3 wanted to leave the UMH as-it-is? Did something changed?
(we didn't discuss this particular change before).

solardiz:
We have to decide on it one way or the other. No point in having it both ways. Adam didn't fully agree with me, but in the end replied with "OK." to my "Let's not overcomplicate things." in a message CC'ed to you. So my reading is he's OK with us simplifying this unconditionally.

@Adam-pi3 Please confirm this or correct me. Thanks!

Adam-pi3:
Hi. Sorry for the inaccuracy. I'm fine with dropping our current logic (dynamic creation of the list on the stack - even it has a lot of security advantages) and switch to the simplified version. The only hard-requirement is to be sure that list is being kept in the RO-page.
As we discussed with @solardiz it's better to NOT keep it in our internal RO-page and I like the @solardiz idea of being it as a part of the .text section. I think this should be the ideal situation.
@oshogbo would you mind going to that direction?

Thanks,
Adam

oshogbo:
I addressed your suggestions.

src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
```diff
@@ -67,54 +36,42 @@ static struct kretprobe p_call_usermodehelper_kretprobe = {
...
-                "Allowing: [%s]\n", p_umh_global[i].p_path);
-         p_allowed++;
-      }
-   }
-
-   p_print_log(P_LKRG_WARN,
-          "%zu/%zu UMH paths are allowed...\n",
-          p_allowed, nitems(p_umh_global));
-}
+static const char *p_umh_global[] = {
```
solardiz, line 39: Should be `static const char * const` so that both the strings and the pointers to them are constant. See how we have it e.g. for `p_log_level_string`.

solardiz:
Please fix the commit message grammar, e.g. changing it to:

```
UMH: Allow all the pathnames listed in p_umh_global.

Previously, we'd only allow pathnames present in the filesystem at LKRG startup.
Now we don't require them to be present in the filesystem.
```

Please also edit this PR's title.

src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
oshogbo: Fixed.

src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
```diff
@@ -132,18 +89,12 @@ int p_call_usermodehelper_entry(struct kretprobe_instance *p_ri, struct pt_regs
...
-            break;
-         }
-      }
-   }
+    for (i = 0; i < nitems(p_umh_global); i++) {
+        if (!strcmp(p_umh_global[i], p_subproc->path)) {
+           p_umh_allowed = 1;
+           break;
+        }
+    }
```
solardiz, line 97: What's happened to the indentation here?

src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
```diff
@@ -67,54 +36,42 @@ static struct kretprobe p_call_usermodehelper_kretprobe = {
...
+       "/sbin/poweroff",
+       "/sbin/request-key",
+       "/sbin/tomoyo-init",
+       "/sbin/v86d",
+       "/system/bin/start",
+       "/usr/lib/systemd/systemd-cgroups-agent",
+       "/usr/lib/systemd/systemd-coredump",
+       "/usr/sbin/eppfpga",
+       "/usr/share/apport/apport",
+};
```
solardiz, line 67: The indentation of this initializer is greater than we currently use in LKRG, but if these are tabs then it'll eventually be consistent with the Linux kernel coding style that we should move to.

We didn't previously have a comma after the last entry in this list. I don't mind having one, though.

solardiz (approved):
Overall, this looks good to me. It's obvious enough that it got to work, but I think it's good practice to always mention what testing was (not) done on each proposed change, so I'd appreciate having that info in a comment in this PR. Thanks!

solardiz:
@oshogbo If you're going to amend the commit in any way (such as for the indentation), please also adjust the commit message further to:

```
UMH: Allow all the pathnames listed in p_umh_global

Previously, we'd only allow pathnames present in the filesystem at LKRG
startup.  Now we don't require them to be present in the filesystem.
```

(Wrap the long line. Let's keep them to at most 75 chars since `git log` shows them indented by 4.)

src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
oshogbo: Those aren't tabs. I will fix them according to the rest of LKRG (3 spaces).

src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
oshogbo: Sorry I used wrong vim config ;/

PR #31, opened 2020-12-15
Build LKRG in-tree using copy-builtin.sh

sempervictus:
DKMS is not always desirable on a system which one intends to
harden from local attackers. The build-chain creates a wide local
attack surface and can be used to build kmode rootkits and other
fun toys for red team. There are also space, performance,
and security considerations such as leaving RANDSTRUCT seeds
around in headers which can be addressed by bringing the module
in with the kernel package.

In order to avoid this, and take advantage of pending advances
in LTO, KCFI, and other methods depending on or benefitting from
having all of the code available at initial compile/link-time, a
script has been added to update the desired kernel tree with the
LKRG source code and relevant Makefile and Kconfig directives.

Since LKRG is still in active development, this script pulls all
relevant content and the current commit hash to let the kernel
maintainers using it keep aware of how far they've fallen behind
and see if subsequent commits merit a rebuild of the in-tree code.

Testing:
 Build-tested against linux-hardened 5.4.83.a as module

TODO:
 Determine if CONFIG_KALLSYMS_ALL is required when building the
module with the rest of the kernel or linking it directly into the
kernel binary.
 Test linking the module directly into the kernel, test resulting
kernel stability, defensive functions, etc.
 Create and apply documentation diffs in the script to inform
users of relevant boot and runtime parameters, as well as where to
go for help and issue reports.

### DIFF
--- a/scripts/copy-builtin.sh
+++ b/scripts/copy-builtin.sh
@@ -1,0 +1,64 @@
+#!/bin/sh
+
+set -e
+# Build simple variables
+KDIR="${KDIR:="/usr/src/linux"}"
+LDIR="$KDIR/security/lkrg"
+BASEDIR="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+SDIR="$BASEDIR/../src"
+COMMIT="LKRG in-tree @ $(git log|head -1|cut -d' ' -f2|cut -c 1-24)"
+# Build heredoc variables
+KCONFIG=$( cat <<EOC
+# SPDX-License-Identifier: GPL-2.0-only
+config SECURITY_LKRG
+	tristate "LKRG support"
+	depends on SECURITY
+	default m
+	help
+	  This selects LKRG - Linux Kernel Runtime Guard, which provides
+          integrity validation and anti-exploitation functions.
+
+	  If you are unsure how to answer this question, answer M.
+EOC
+)
+
+MAKEFILE=$(cat <<EOC
+# SPDX-License-Identifier: GPL-2.0-only
+
+obj-\$(CONFIG_SECURITY_LKRG) := p_lkrg.o
+$(awk '/^p_lkrg-objs/,/^$/' "$BASEDIR/../Makefile"|sed -e 's|src/||')
+ 
+EOC
+)
+
+MAKEADD=$(cat <<EOC
+
+# LKRG file list
+subdir-\$(CONFIG_SECURITY_LKRG)         += lkrg
+obj-\$(CONFIG_SECURITY_LKRG)            += lkrg/
+EOC
+)
+# Tell user what we're about to do
+echo "Copying $SDIR/* to $LDIR along with Kconfig:"
+echo "$KCONFIG"
+echo
+echo "and Makefile"
+echo "$MAKEFILE"
+echo "Commit msg: $COMMIT"
+echo "Ctrl+c to quit, any other key to continue"
+read CANCEL
+# Execute copy
+mkdir -p "$LDIR"
+echo "$KCONFIG" > "$LDIR/Kconfig"
+echo "$MAKEFILE" > "$LDIR/Makefile"
+pushd .
+cd "$SDIR"
+cp -a . "$LDIR/"
+cd "$KDIR"
+# Update sources for built-in usage
+sed -i '/source "security\/integrity\/Kconfig"/asource "security/lkrg/Kconfig"' security/Kconfig
+echo "$MAKEADD" >> security/Makefile
+# Commit the changes
+git add "security/lkrg"
+git commit -am "$COMMIT"
+popd


solardiz:
Thanks @sempervictus. I summarized another similar effort in https://www.openwall.com/lists/lkrg-users/2020/05/02/8, including:

> Then the kernel
> would lock up on bootup.  Nikolay further solved that by changing our
> module_init to late_initcall_sync (he also says simple late_initcall
> didn't help), so that LKRG initialization is postponed.

Did you somehow not run into this issue? Should we possibly make that change anyway (perhaps for compatibility with other kernels, like whatever Nikolay was using)? Would you try it?

scripts/copy-builtin.sh
```diff
@@ -0,0 +1,59 @@
...
+echo "$MAKEFILE"
+echo "Commit msg: $COMMIT"
+echo "Ctrl+c to quit, any other key to continue"
+read CANCEL
+
+mkdir -p "$LDIR"
+echo "$KCONFIG" > "$LDIR/Kconfig"
+echo "$MAKEFILE" > "$LDIR/Makefile"
+pushd .
+cd "$SDIR"
```
solardiz, line 52: I think the script should start with `set -e` (right after `#!/bin/sh`), so that if a command like `cd` fails, it terminates right away and doesn't do anything stupid.

scripts/copy-builtin.sh
```diff
@@ -0,0 +1,59 @@
...
+cp -r * "$LDIR/"
```
solardiz, line 53: I think this is better written as `cp -a . "$LDIR/"`. No need for wildcard expansion and for excluding dot-files.

scripts/copy-builtin.sh
```diff
@@ -0,0 +1,59 @@
...
+
+KDIR="${KDIR:="/usr/src/linux"}"
+LDIR="$KDIR/security/lkrg"
+BASEDIR="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+SDIR="$BASEDIR/../src"
+COMMIT="LKRG in-tree @ $(git log|head -1|cut -d' ' -f2|cut -c 1-24)"
+KCONFIG=$( cat <<EOC
+# SPDX-License-Identifier: GPL-2.0-only
+config SECURITY_LKRG
+	tristate "LKRG support"
```
solardiz, line 11: Until building in-tree and booting the resulting kernel is actually tested and works, can we perhaps allow only `n` and `m` here?

solardiz (commented):
Although I haven't fully reviewed this (would need to check a few external things for that), here are some suggested changes.

sempervictus:
Hey Alex, thanks for looking into this. Responses in order:
1. Have not run into the boot-time lockup, likely because i'm not loading the module at boot but well into the userspace startup
2. Will add `set -e`
3. I can make the change to use this, though there shouldn't be any dotfiles given that we're copying the `.c` and `.h` files
4. I've never done that, nor had to really, but will look it up and see if thats an option (m^n vs tristate)

solardiz:
>     1. Have not run into the boot-time lockup, likely because i'm not loading the module at boot

That's a misunderstanding. It's OK to load the module early at boot, and shouldn't cause any issues. The issue I mentioned is for linking LKRG into the kernel, which you're not doing yet - so that's why you shouldn't have run into the issue yet.

>     3. I can make the change to use this, though there shouldn't be any dotfiles given that we're copying the `.c` and `.h` files

I think avoiding unnecessary shell wildcards is good practice for robustness in general.

>     4. I've never done that, nor had to really, but will look it up and see if thats an option (m^n vs tristate)

Alternatively, maybe your time is better spent on supporting `y` sooner rather than later.

sempervictus:
I've kicked off a `CONFIG_SECURITY_LKRG=y` build, lets see if it eats a VM or two.
Also pushed the `cp -a .` change as requested.

The "in-tree as module" VM just completed 10h of ZFS zloop along with a looping CPU syscall/csw/etc benchmark to load up the system in fun and potentially conflicting ways. No damage so far.
While the build and subsequent tests run, any thoughts on whether `CONFIG_KALLSYMS_ALL=y` is needed when building this way?

sempervictus:
@solardiz: you mean **[this](https://pasteboard.co/JF4JClQ.png)** boot-time crash?

solardiz:
> any thoughts on whether `CONFIG_KALLSYMS_ALL=y` is needed when building this way?

My guess is it's unrelated to how you build the module. LKRG generally needs it when built as a module, but might sometimes succeed without it.

When you link LKRG into the kernel, this option will probably not be needed.

@Adam-pi3 probably has a better idea on this.

>  you mean **[this](https://pasteboard.co/JF4JClQ.png)** boot-time crash?

Nikolay mentioned lock-up, not crash, so this is probably something different, but I don't really know.

solardiz:
@sempervictus "Drop shell wildcard per Solar Designer" shouldn't be a separate commit when we merge this - it's part of this PR's history, but is too minor to be part of LKRG project's history. Please just amend your one commit and force-push. For `set -e`, too. And retest the script with these changes. Thanks!

sempervictus:
Squashed back to one commit - normally keep the individual ones during review and rebase for a final merge (at least in-house).
Far as the crash, does lkrg have logic to handle `ro_after_init`? Would be nice if the site at which the detection occurred were a bit more verbosely displayed, but whatever the case, it seems reasonable to assume that its sampling something that changes during init since i've beat the bag out of systems running with it on as a module and seen no such ill effects despite them loading modules at runtime to support OpenVSwitch, XTables, OverlayFS/AUFS, etc.
One possible solution may be for it to "establish a baseline" after its own init - run multiple comparisons to ensure stability in the RO sections before starting to respond to their changes (a deterministic delay to enforcing protections and possibly a regional one where things in flux stay in a quarantine until they're stable for 3 checks in a row).

sempervictus:
@solardiz: how do we want to handle the `TRISTATE` issue? Is there any momentum for changes in the actual module to deal with initializing at the same time as other components which may alter what should be romem at init after the initial sample is taken, or should we try to find some hackery to force module-only builds in the script? Kind of a shame for stuff like GrapheneOS which doesn't use modules though, and other likely compat issues aside, could benefit from having a live integrity validation tool for kernel runtime memory.

solardiz:
> Squashed back to one commit

If you don't mind, please replace the commit message with:

```
Add copy-builtin.sh, which allows to build LKRG in tree

Only building as a module is currently fully supported.
```

I think most of what you wrote in there belongs in this PR (thank you!), but not in the commit message.

>  how do we want to handle the `TRISTATE` issue?

Maybe document (in the option description) that "building LKRG into the kernel is currently known to result in a non-working kernel, and is thus at this time only for use by LKRG developers."

> Is there any momentum for changes in the actual module to deal with initializing at the same time as other components

Currently, no. Adam is busy with certain other aspects of LKRG development first.

Can you try Nikolay's "changing our module_init to late_initcall_sync", just in case this is sufficient against the issue you're seeing as well? Somehow things did work for Nikolay.

sempervictus:
@solardiz: could you please link me to the relevant commit? Kinda doing this touch-and-go in spare minutes between actual work (otherwise i'd dig into the init issue myself).
Will update the commit msg once we've tracked down this lead if we cant make it work outright.

solardiz:
> could you please link me to the relevant commit?

There was no commit, only a private e-mail from Nikolay stating those things, without further detail.

Adam-pi3:
@sempervictus The problem which you are seeing is related to the modification of the RO section during boot. I think that the best solution for this problem is to load/run LKRG after early boot initialization. @solardiz suggestion to use Nikolay's changes (use late_initcall_sync) might be the solution for the problem. Another solution might be to synchronize LKRG with some global kernel knob. In such case, LKRG would do busy wait until such 'knob' is initialized. However, "late_initcall_sync" might be easier and simpler path to go.

Adam-pi3:
Can you try this diff ?

```
$ git diff
diff --git a/src/p_lkrg_main.c b/src/p_lkrg_main.c
index 0d7286f..343ca81 100644
--- a/src/p_lkrg_main.c
+++ b/src/p_lkrg_main.c
@@ -678,7 +678,7 @@ static void __exit p_lkrg_deregister(void) {
 }


-module_init(p_lkrg_register);
+late_initcall_sync(p_lkrg_register);
 module_exit(p_lkrg_deregister);

 module_param(log_level, uint, 0000);
```

sempervictus:
@Adam-pi3: thanks, i implemented it as a `sed` call in the script for now, but it **does** work:
```
[    3.093964] ### dt-test ### FAIL of_unittest_overlay_high_level():2392 overlay_base_root not initialized
[    3.094617] ### dt-test ### end of unittest - 218 passed, 1 failed
[    3.094620] [p_lkrg] Loading LKRG...
[    3.096424] Freezing user space processes ... (elapsed 0.000 seconds) done.
[    3.096444] OOM killer disabled.
[    3.208469] [p_lkrg] [kretprobe] register_kretprobe() for <ovl_create_or_link> failed! [err=-22]
[    3.208470] [p_lkrg] Trying to find ISRA / CONSTPROP name for <ovl_create_or_link>
[    3.228757] [p_lkrg] [kretprobe] register_kretprobe() for ovl_create_or_link failed and ISRA / CONSTPROP version not found!
[    3.229434] [p_lkrg] Can't hook 'ovl_create_or_link' function. This is expected if you are not using OverlayFS.
[    3.257337] [p_lkrg] [kretprobe] register_kretprobe() for <ttwu_do_wakeup> failed! [err=-22]
[    3.257337] [p_lkrg] Trying to find ISRA / CONSTPROP name for <ttwu_do_wakeup>
[    3.275456] [p_lkrg] [kretprobe] register_kretprobe() for ttwu_do_wakeup failed and ISRA / CONSTPROP version not found!
[    3.276105] [p_lkrg] LKRG won't enforce validation when task is changing the state to RUNNING
[    3.381873] [p_lkrg] LKRG initialized successfully!
[    3.382253] OOM killer enabled.
[    3.382254] Restarting tasks ... done.
```

sempervictus:
@solardiz: i'll merge down the commits and reword once we're all agreed its ready to roll

solardiz:
@sempervictus I think this is fine to merge once you squash the commits and remove the excessive detail from commit messages. Such detail is very welcome here in PR comments, but not in the project's long-term history.

Longer-term, we should consider making the code change unconditionally, including for module builds. Maybe you can help test that.

Thank you!

As to this part:

```
[    3.257337] [p_lkrg] [kretprobe] register_kretprobe() for <ttwu_do_wakeup> failed! [err=-22]
[    3.257337] [p_lkrg] Trying to find ISRA / CONSTPROP name for <ttwu_do_wakeup>
[    3.275456] [p_lkrg] [kretprobe] register_kretprobe() for ttwu_do_wakeup failed and ISRA / CONSTPROP version not found!
[    3.276105] [p_lkrg] LKRG won't enforce validation when task is changing the state to RUNNING
```

Luckily, this is functionality that we intend to drop anyway, but I am mildly curious what prevented that symbol from being found.

sempervictus:
So it looks like there might be some "weirdness" afoot - it's not a fan of `dockerd`:
```
[ 7346.968292] [p_lkrg] <Exploit Detection> process[1584 | dockerd] has corrupted 'off' flag!
[ 7346.968378] [p_lkrg] <Exploit Detection> Trying to kill process[dockerd | 1584]!
[ 7346.968454] [p_lkrg] <Exploit Detection> process[1584 | dockerd] has corrupted 'off' flag!
[ 7346.968506] [p_lkrg] <Exploit Detection> Trying to kill process[dockerd | 1584]!
[ 7346.968574] [p_lkrg] <Exploit Detection> process[1584 | dockerd] has corrupted 'off' flag!
[ 7346.968659] [p_lkrg] <Exploit Detection> Trying to kill process[dockerd | 1584]!
[ 7346.968718] [p_lkrg] <Exploit Detection> process[1584 | dockerd] has corrupted 'off' flag!
[ 7346.968774] [p_lkrg] <Exploit Detection> Trying to kill process[dockerd | 1584]!
[ 7346.968821] [p_lkrg] <Exploit Detection> process[1584 | dockerd] has corrupted 'off' flag!
[ 7346.968881] [p_lkrg] <Exploit Detection> Trying to kill process[dockerd | 1584]!
[ 7346.968937] [p_lkrg] <Exploit Detection> process[1584 | dockerd] has corrupted 'off' flag!
[ 7346.968990] [p_lkrg] <Exploit Detection> Trying to kill process[dockerd | 1584]!
```
 Running another build with LKRG as `=m` to see if this is related to whats been done here, will merge-squash-reword the commits meanwhile.

solardiz:
@sempervictus Regarding the Docker false positive: https://www.openwall.com/lists/lkrg-users/2020/02/01/1

We need to load overlayfs support before LKRG is loaded. I guess with both being compiled into the kernel image, the order might not be right. I guess we'll need to come up with some way of making it right. @Adam-pi3?

solardiz:
> We need to load overlayfs support before LKRG is loaded. I guess with both being compiled into the kernel image, the order might not be right. I guess we'll need to come up with some way of making it right.

... or better yet, if LKRG fails to hook overlayfs functions right away, it should monitor for them appearing later and hook them when available. This should help when overlayfs is built as a module.

And, I don't understand the issue when everything is built into the kernel, then. @sempervictus In your testing above, was module support possibly left enabled, and `overlay` got loaded as a module when Docker requested it? Just a guess.

Adam-pi3:
> it should monitor for them appearing later and hook them when available. This should help when overlayfs is built as a module.

Hm... That's an interesting concept. However, from the technical point of view I could only imagine that we can do strcmp() on the newly loaded module name. If such dumb check is positive then we can try to re-apply such hook.

Nevertheless, I'm not sure if we should go to this path. Maybe we should (don't know yet). It is somehow related to our offline discussion how to make it more clear and *visible* that if someone want to protect (and be compatible) namespaces and docker via LKRG, it needs to load related modules (like overlayfs) before loading LKRG.

solardiz:
> from the technical point of view I could only imagine that we can do strcmp() on the newly loaded module name. If such dumb check is positive then we can try to re-apply such hook.

This would be sufficient to prevent most of the problem. Few people rename the module or such. Also, from a technical perspective, we could do crazy things like try to setup the hooks every few seconds. I'm not saying we should. Regardless, an issue remains that we might not be quick enough, if module loading is triggered immediately before its use.

> if someone want to protect (and be compatible) namespaces and docker via LKRG, it needs to load related modules (like overlayfs) before loading LKRG.

Much of the value in LKRG, e.g. compared to kernel patches, is in its ease of use. When we start writing non-trivial installation instructions with (eventually) lots of caveat, we've lost that value. I think we should try hard to make LKRG just work right after the easiest installation procedure and little reading of documentation. Sometimes this requires being smart in the code. Sometimes it may require dropping features. Even dropping protections.

sempervictus:
Overlayfs was built as a module, while lkrg was not. I can confirm that loading overlay first does resolve it.
Far as hooking module load, can we piggyback on the signature verification call? Detour their function return to call lkrg for instrumentation or the like? How to LSMs hook module loading actions?

Separately, looks like it does work to catch evildoers running basic privileged docker escapes:


docker_event

sempervictus:
@solardiz: far as the ease of use thing, i totally agree. Thing is though, adoption into a distribution will be the fastest way to perform real world testing and shakedown. Distros want things that are "easy to maintain" and use, so i think that ironing out compatibility issues like this would help remove roadblocks to wider adoption. I'm already in the process of getting it into VyOS.

Separately, the late initcall thing _should_ probably go in the main codebase because even with modules built via DKMS users can stuff the thing into initrd and then we have the issue of the real FS containing modules loaded after LKRG inits.

Adam-pi3:
Couple of notes:
- CONFIG_KALLSYMS_ALL - even if LKRG is linked to the kernel, we are still using directly or indirectly (via kprobes) official kallsyms_* related API. I would expect this option to be mandatory.
- Thanks for verifying docker escapes. It is one of the feature which we wanted to have in LKRG. However, your escape is based on UMH, not kernel data corruption. Nevertheless, we do lock-down UMH as well.
- @sempervictus  "(...) can we piggyback on the signature verification call" - I'm not sure if LKRG should go to this path. You can build Linux kernel with CONFIG_MODULE_SIG option to achieve what you want. It should not be LKRG's job.
- @solardiz I think you have a point. I will take a look how we can add such feature in not a disaster way ;p However, I'm not sure I will agree with "Sometimes it may require dropping features. Even dropping protections".
- I've just pushed "late_initcall_sync" change to the official repo:
https://github.com/openwall/lkrg/commit/ddc14c6544cb1639132b765fe6df72c3feba14aa

Thanks,
Adam

sempervictus:
@Adam-pi3: i removed the `sed` call to do late init since ddc14c654 is in now. Thank you.

Adam-pi3:
@sempervictus I'm not sure if you had a chance to see the recent commits, but I wanted to let you know that we added support for the dynamic handling of 'overlay[2]' module.

Thanks,
Adam

sempervictus:
Thank you, will update the builtin :)

PR #32, opened 2020-12-16
p_integrity_timer.c: add compatibiity definition for kzfree() (fix bu…

wladmis:
…ild for kernel >= 5.10)

### Description
FTBFS against kernel 5.10

### How Has This Been Tested?
...

### DIFF
--- a/src/p_lkrg_main.h
+++ b/src/p_lkrg_main.h
@@ -76,6 +76,12 @@
 #include <linux/cryptohash.h>
 #endif
 
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5,10,0)
+#define p_kzfree kzfree
+#else
+#define p_kzfree kfree_sensitive
+#endif
+
 #include <linux/stacktrace.h>
 #include <asm/stacktrace.h>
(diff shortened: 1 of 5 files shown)


src/p_lkrg_main.h
```diff
@@ -76,6 +76,10 @@
 #include <linux/cryptohash.h>
 #endif
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,10,0)
+#define kzfree kfree_sensitive
```
Adam-pi3, line 80: I see in this commit:
https://github.com/torvalds/linux/commit/23224e45004ed84c8466fd1e8e5860f541187029

They have removed definition of kzfree and replaced it with kfree_sensitive as you define here. I wonder if in such case isn't better to replace our calls to kzfree() with p_kzfree() and define p_kzfree() as kzfree() or kfree_sensitive() depends on kernel version. Your commit is fine,  but scenario which I'm describing should give us better control over it.
@solardiz what do you think?

Adam-pi3 (changes requested):

Adam-pi3 (approved):

PR #33, opened 2020-12-30
In-tree: fix for 58b9ec06

sempervictus:
After 58b9ec06, the script which copies module files into the
kernel source tree does not properly capture the files to be built
into p_lkrg.o.
Fix the awk expression performing the capture to adjust for the
new `$(TARGET)-obj` syntax and sed in-place to restore the original
name for in-tree build.

### DIFF
--- a/scripts/copy-builtin.sh
+++ b/scripts/copy-builtin.sh
@@ -26,7 +26,7 @@ MAKEFILE=$(cat <<EOC
 # SPDX-License-Identifier: GPL-2.0-only
 
 obj-\$(CONFIG_SECURITY_LKRG) := p_lkrg.o
-$(awk '/^p_lkrg-objs/,/^$/' "$BASEDIR/../Makefile"|sed -e 's|src/||')
+$(awk '/^\$\(TARGET\)-objs/,/^$/' "$BASEDIR/../Makefile"|sed -e 's|src/||; s|$(TARGET)-objs|p_lkrg-objs|')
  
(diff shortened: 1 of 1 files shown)


sempervictus:
@Adam-pi3 @solardiz - if you guys want to rebase and move the original in-tree commit to HEAD, you can squash this right into it.

solardiz:
Thank you, @sempervictus!

solardiz:
@sempervictus BTW, GitHub does not recognize your commits in these PRs as belonging to your GitHub account, and thus does not link to it from there and does not count/list you among our project contributors. I guess you don't have the e-mail address you use on commits listed among your addresses on GitHub. If so, you might want to add it to there.

sempervictus:
@solardiz - thanks for the heads up. GH wont take the commitline `Author: RageLtMan <rageltman [at] sempervictus>` since its not an email per se. Years ago we had a problem in the metasploit community with arses of various calibers trawling our commits and module signatures to spam the contributors. Took up this format to avoid the skiddie scrapers, but there are some issues. If you guys could possibly manually add me to the contributor list, i'd appreciate it, plus gives some cache when explaining the project to people who may want to leverage it for their use.

solardiz:
> If you guys could possibly manually add me to the contributor list

I doubt there's a way to do that, as I think the list is based solely on commits.

We could add you as a Collaborator on the repository, but it's not the same thing.

PR #36, opened 2021-01-03
WIP: Replace 'search_binary_handler' hook

### LINKED ISSUE #35: register_kretprobe() for search_binary_handler failure on Linux 5.10
Linux 5.10 no longer exports the `search_binary_handler` symbol, and LKRG is unable to locate it. On a 5.10 kernel, the following messages are logged:

```
[   27.130301] [p_lkrg] Loading LKRG...
[   27.148158] Freezing user space processes ... (elapsed 0.000 seconds) done.
[   27.150367] OOM killer disabled.
[   27.177919] [p_lkrg] [kretprobe] register_kretprobe() for <search_binary_handler> failed! [err=-22]
[   27.179199] [p_lkrg] Trying to find ISRA / CONSTPROP name for <search_binary_handler>
[   27.203672] [p_lkrg] [kretprobe] register_kretprobe() for search_binary_handler failed and ISRA / CONSTPROP version not found!
[   27.205072] [p_lkrg] ERROR: Can't hook search_binary_handler :(
[   27.206020] [p_lkrg] Can't initialize exploit detection features! Exiting...
```

Looks like the relevant upstream commit where this happened is [here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=9d9488d4d7c85a76e9fd30d7a070ed018f767233)

0xC0ncord:
Linux 5.10 removes external usage of this symbol. Instead, we hook into
the entry of 'security_bprm_check' and the return from
'proc_exec_connector' for pre-exec actions and post-exec actions,
respectively. This approach allows us to use a single hook method across
all supported kernel versions.

Fixes #35 

### Description
The hook for `search_binary_handler` is now split into two separate kprobe hooks: one for the entry into `security_bprm_check` and one for the return from `proc_exec_handler`.

### How Has This Been Tested?
This was tested on a running Linux 5.10.4 machine with regular desktop activity, including the usage of SUID binaries (i.e. `sudo`). Hooking into the entry of `proc_exec_connector` was found to be too early to run post-exec actions, and as a side effect, would trigger LKRG's exploit detection prematurely on valid SUID binaries. Through testing, hooking into the return of this function was found not to trigger this behavior.


0xC0ncord:
Actually, double-checking things would indicate this might need more testing. We should verify that exploit detection is functioning properly at all after this change.

solardiz (approved):
This looks OK to me, but let's wait to hear from Adam before we merge it.

solardiz:
Here's an issue (which I didn't spot when suggesting this approach earlier today, my bad): we got a dependency on `CONFIG_PROC_EVENTS` now. If this is acceptable, we should require this setting at LKRG build time. If not, we should reconsider the approach at least for pre-5.10 kernels (where we can continue to hook `search_binary_handler`).

solardiz:
> Hooking into the entry of `proc_exec_connector` was found to be too early to run post-exec actions, and as a side effect, would trigger LKRG's exploit detection prematurely on valid SUID binaries. Through testing, hooking into the return of this function was found not to trigger this behavior.

That's weird. I don't see anything in `proc_exec_connector` which would make a relevant difference. Here it is:

```c
void proc_exec_connector(struct task_struct *task)
{
	struct cn_msg *msg;
	struct proc_event *ev;
	__u8 buffer[CN_PROC_MSG_SIZE] __aligned(8);

	if (atomic_read(&proc_event_num_listeners) < 1)
		return;

	msg = buffer_to_cn_msg(buffer);
	ev = (struct proc_event *)msg->data;
	memset(&ev->event_data, 0, sizeof(ev->event_data));
	ev->timestamp_ns = ktime_get_ns();
	ev->what = PROC_EVENT_EXEC;
	ev->event_data.exec.process_pid = task->pid;
	ev->event_data.exec.process_tgid = task->tgid;

	memcpy(&msg->id, &cn_proc_event_id, sizeof(msg->id));
	msg->ack = 0; /* not used */
	msg->len = sizeof(*ev);
	msg->flags = 0; /* not used */
	send_msg(msg);
}
```

In fact, it probably normally hits the early `return;`, so really does nothing most of the time.

@0xC0ncord I suspect an issue in your testing of the LKRG code revision with hooking entry into that function - e.g., incomplete rebuild or something. I think either should work, and hooking entry would be very slightly preferable.

solardiz:
I've just built and loaded LKRG with this PR's changes (and a workaround for #37) on a CentOS 7 system to see that this approach works on the oldest kernels we support. No issues seen so far.

0xC0ncord:
> Here's an issue (which I didn't spot when suggesting this approach earlier today, my bad): we got a dependency on `CONFIG_PROC_EVENTS` now. If this is acceptable, we should require this setting at LKRG build time. If not, we should reconsider the approach at least for pre-5.10 kernels (where we can continue to hook `search_binary_handler`).

One of the issues I ran into right away when building this patch is how the changes are to be structured. Previously I had tested keeping `search_binary_handler` in the code and only including it by checking the kernel version, which included a patch to the `Makefile` to only build it into LKRG if on a kernel version where it still worked. This is perfectly do-able but makes the code a little messy. Otherwise, we can continue to build it, but then we end up having unreachable code being built in when it is not needed.

>I suspect an issue in your testing of the LKRG code revision with hooking entry into that function - e.g., incomplete rebuild or something. I think either should work, and hooking entry would be very slightly preferable.

I'm looking into it right now. I thought the same thing when I investigated it.

solardiz:
> keeping `search_binary_handler` in the code and only including it by checking the kernel version

If we do that, I think the kernel version check should be in the C sources, not in the `Makefile`. It is OK for a source file to end up being effectively empty after the C preprocessor, and to produce a dummy object file.

Besides, I think at a later time we'll need to abandon Adam's keeping of every hooked function in its separate source file. I'd prefer source files per functionality they implement (e.g., hooking everything needed to support `execve` right can be in one source file), not per current implementation detail (which as we can see keeps changing, including in what functions we hook, and I find it a bit ridiculous that under the current convention we have to be adding/removing source files for that).

0xC0ncord:
> It is OK for a source file to end up being effectively empty after the C preprocessor, and to produce a dummy object file.

OK. I wasn't sure if this was appropriate, but now I understand. If we decide to go that route, I will rework this patch to do just that, keeping `search_binary_handler` for compatibility with older kernels.

> That's weird. I don't see anything in proc_exec_connector which would make a relevant difference.
https://github.com/0xC0ncord/lkrg/commit/c32e2fe66cc04cff4382e575ef07ccab1d504e90 (`next` branch) has a change on top of this PR to use entry to `proc_exec_connector`, and I'm able to reproduce the behavior where valid SUID executables trigger exploit detection on my system.

solardiz:
> [0xC0ncord@c32e2fe](https://github.com/0xC0ncord/lkrg/commit/c32e2fe66cc04cff4382e575ef07ccab1d504e90) (`next` branch) has a change on top of this PR to use entry to `proc_exec_connector`, and I'm able to reproduce the behavior where valid SUID executables trigger exploit detection on my system.

Thanks. I think this way of hooking might not actually work:

```c
    .handler = NULL,
    .entry_handler = p_proc_exec_connector_entry,
```

It's a kretprobe that you're trying not to use as such. I think elsewhere in LKRG all kretprobes have a return hook, even if dummy. BTW, we might want to optimize those to pure kprobes, @Adam-pi3?

solardiz:
Anyway, it's fine to continue with the return hook only for now, until @Adam-pi3 possibly approves cheaper hooking of entry only, which we also need e.g. for `security_capable()` (are currently wasting CPU time on a return hook there).

0xC0ncord:
> It's a kretprobe that you're trying not to use as such. I think elsewhere in LKRG all kretprobes have a return hook, even if dummy. 

Ah I see. I will test with dummy hooks to see if that's the problem.

0xC0ncord:
Seems that https://github.com/0xC0ncord/lkrg/commit/5c7b5953101b3049799662d79c6ba9c5144aede1 did not fix the issue.

Adam-pi3:
I've been looking into this issue and PR. Most of my kernel (including 5.10+) doesn't have that problem with 'search_binary_handler'. However, one of my very custom compilation of 5.9.7 does. I assume it is mostly related to the configuration and optimization which GCC uses.
@0xC0ncord Is it possible that you can share some more information what kernel(s) are you using when you are seeing this issue? Is it custom compilation or distro kernels? What is enabled that you might see that issue?

Nevertheless, I was successfully compiled and run these changes. I've also run some of my exploits and it seems everything correct. However, I'm not feeling comfortable (yet) to merge this PR now since it's a core of LKRG's tracking logic and we need more comprehensive tests.
On the other hand, we are planning to revisit that specific piece of LKRG's functionality in v.next version and maybe it is safer to leave 'search_binary_handler' as is for kernels < 5.10 and hook 'do_execveat_common' in 5.10+. @solardiz what do you think? It is also worth to mention that it is very likely that we end-up having more than 1 hook for exec* in the future revision of it.
Another option might be to hook each supported binary format handler independently like 'load_elf_binary', 'load_aout_binary', etc.

Btw. There is an independent discussion about that in open 'issue' for that problem. Do we prefer to continue discussion on it here or there?
Btw2. @0xC0ncord  what SUID binary are you executing that triggers FP with this PR ? I've run a few and didn't notice any GP

solardiz:
> Btw. There is an independent discussion about that in open 'issue' for that problem. Do we prefer to continue discussion on it here or there?

I think most of what you wrote belongs on the issue and not this PR.

> Btw2. @0xC0ncord what SUID binary are you executing that triggers FP with this PR ? I've run a few and didn't notice any GP

That's a misunderstanding: this PR isn't known to cause any FPs. @0xC0ncord had another code version that did.

solardiz:
Here's a likely problem with the approach I had suggested and @0xC0ncord implemented here: the call to `proc_exec_connector` isn't reached on (at least some) exec failures. This would be good in that we'd skip LKRG shadow credentials update on failures, but it's unacceptable in that we'd never reset `p_off`, which we set on `security_bprm_check`.

I should have given this more thought before suggesting that @0xC0ncord implements it. Sorry.

solardiz (changes requested):
Setting "Request changes" because of the likely issue I just described, and no way to simply recall my prior approval. I am not actually requesting changes just yet. I think it might be cleaner to close this PR, and open a new one based on whatever approach we come up with in discussion with Adam on the issue.

0xC0ncord:
@Adam-pi3 I will post answers to your questions on the relevant issue #36

solardiz:
Adam ended up dealing with the issue differently. I'll close this PR.

PR #42, opened 2021-01-17
p_cmp_creds considers sudo to be a privesc attack

sempervictus:
Check whether the current process is "sudo" before running all of
the UID checks.

The stack traces produced priorto this commit show failures in the
EUID, SUID, and FSUID checks.

This may not be the best way to handle this as attackers running
privesc binaries just need to name them "sudo," but it does fix
the runtime issue of trying to use the legitimate sudo bin. A path
check for "/sbin/sudo" might be better, but distributions place
binaries in different places... Ideally we'd have some better way
to determine if the escalating binary is legitimate along with
coverage for other potentially impacted valid privesc methods.

### Description
Added a dirty hack to let me use `sudo` by checking the process
name for a `strcmp` against "sudo" - probably not the best way to
do this.

### How Has This Been Tested?
Built module against linux-hardened 5.4.89, loaded, ran sudo.

### DIFF
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -1226,55 +1226,55 @@ static unsigned int p_iterate_lkrg_tasks_paranoid(void) {
 int p_cmp_creds(struct p_cred *p_orig, const struct cred *p_current_cred, struct task_struct *p_current, char p_opt) {
 
    int p_ret = 0;
-
-   /* *UID */
-   if (!uid_eq(p_orig->uid, p_current_cred->uid)) {
-      if (p_opt) {
-         p_print_log(P_LKRG_CRIT,
-             "<Exploit Detection> process[%d | %s] has different UID! %d vs %d\n",
-             task_pid_nr(p_current), p_current->comm,
-             p_get_uid(&p_orig->uid), p_get_uid(&p_current_cred->uid));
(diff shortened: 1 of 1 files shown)


sempervictus:
There's got to be a better way to do this, but in the meantime, its a bypass for #41

sempervictus:
The CI failure seems to be CI-related:
```
$ set -x; sudo m-a prepare
++sudo m-a prepare
Getting source for kernel version: 5.4.0-56-generic
apt-get install linux-headers-5.4.0-56-generic 
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package linux-headers-5.4.0-56-generic
E: Couldn't find any package by glob 'linux-headers-5.4.0-56-generic'
E: Couldn't find any package by regex 'linux-headers-5.4.0-56-generic'
Creating symlink...
apt-get install build-essential 
Reading package lists...
Building dependency tree...
Reading state information...
build-essential is already the newest version (12.4ubuntu1).
0 upgraded, 0 newly installed, 0 to remove and 172 not upgraded.
Done!
```
... other builds passed.

solardiz:
Like you say, this is a dirty hack. I think it makes sense for you to use this in a VM to temporarily bypass this one issue and perhaps proceed to identify other incompatibilities of LKRG with your kernel, but there's no reason for us to merge this. Thanks for sharing anyway.

Issue #34, opened 2021-01-01
Add more dependencies to the copy-builtin.sh script to avoid kernel build failures 


morfikov:
I wanted to try the in-kernel build script, but I got the following error (I just changed the KDIR var):

```
$ ./scripts/copy-builtin.sh
Copying /media/debuilder/git-lkrg/lkrg/scripts/../src/* to /media/debuilder/git-kernel/linux-5.10/security/lkrg along with Kconfig:
# SPDX-License-Identifier: GPL-2.0-only
config SECURITY_LKRG
        tristate "LKRG support"
        depends on SECURITY
        default m
        help
          This selects LKRG - Linux Kernel Runtime Guard, which provides
          integrity validation and anti-exploitation functions.

          If you are unsure how to answer this question, answer M.

and Makefile
# SPDX-License-Identifier: GPL-2.0-only

obj-$(CONFIG_SECURITY_LKRG) := p_lkrg.o
p_lkrg-objs += modules/ksyms/p_resolve_ksym.o \
                  modules/hashing/p_lkrg_fast_hash.o \
                  modules/comm_channel/p_comm_channel.o \
                  modules/integrity_timer/p_integrity_timer.o \
                  modules/kmod/p_kmod.o \
                  modules/database/CPU.o \
                  modules/database/arch/x86/p_x86_metadata.o \
                  modules/database/arch/x86/p_switch_idt/p_switch_idt.o \
                  modules/database/arch/arm64/p_arm64_metadata.o \
                  modules/database/arch/arm/p_arm_metadata.o \
                  modules/database/arch/p_arch_metadata.o \
                  modules/database/JUMP_LABEL/p_arch_jump_label_transform/p_arch_jump_label_transform.o \
                  modules/database/JUMP_LABEL/p_arch_jump_label_transform_apply/p_arch_jump_label_transform_apply.o \
                  modules/database/FTRACE/p_ftrace_modify_all_code/p_ftrace_modify_all_code.o \
                  modules/database/FTRACE/p_ftrace_enable_sysctl/p_ftrace_enable_sysctl.o \
                  modules/database/p_database.o \
                  modules/notifiers/p_notifiers.o \
                  modules/self-defense/hiding/p_hiding.o \
                  modules/exploit_detection/p_rb_ed_trees/p_rb_ed_pids/p_rb_ed_pids_tree.o \
                  modules/exploit_detection/syscalls/p_install.o \
[...]

PR #44, opened 2021-01-17
Copy-builtin: use bash and correct depends

sempervictus:
### Description
Address #34

### DIFF
--- a/scripts/copy-builtin.sh
+++ b/scripts/copy-builtin.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 set -e
 # Build simple variables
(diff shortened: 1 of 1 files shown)


scripts/copy-builtin.sh
```diff
@@ -12,7 +12,7 @@ KCONFIG=$( cat <<EOC
 # SPDX-License-Identifier: GPL-2.0-only
 config SECURITY_LKRG
 	tristate "LKRG support"
-	depends on SECURITY
+	depends on SECURITY && CONFIG_KPROBES && CONFIG_MODULE_UNLOAD && CONFIG_KALLSYMS_ALL
```
solardiz, line 15: These shouldn't have the `CONFIG_` prefix, but I guess having it somehow didn't hurt? Anyway, I suggest it be dropped from here.

sempervictus: Thanks - done.

PR #50, opened 2021-02-18
copy-builtin: avoid errors when committing changes

0xC0ncord:
If the source tree is not a git repository or git is not installed,
avoid throwing errors.

### Description
If the source tree to copy LKRG to as a built-in is not a git repository or git is not installed on the machine, errors are thrown although the install is a success. This commit checks for these prerequisites before trying to commit changes to a git repository.

### How Has This Been Tested?
The script was modified and run on both a machine with git installed but the source tree was not a git repository, as well as a machine where git was installed and the source tree was also a repository. In both cases no errors were thrown from the script and the copying of LKRG as a built-in was a success.

### DIFF
--- a/scripts/copy-builtin.sh
+++ b/scripts/copy-builtin.sh
@@ -6,7 +6,9 @@ KDIR="${KDIR:="/usr/src/linux"}"
 LDIR="$KDIR/security/lkrg"
 BASEDIR="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
 SDIR="$BASEDIR/../src"
-COMMIT="LKRG in-tree @ $(git log|head -1|cut -d' ' -f2|cut -c 1-24)"
+if [ git rev-parse --git-dir &>/dev/null ]; then
+    COMMIT="LKRG in-tree @ $(git log|head -1|cut -d' ' -f2|cut -c 1-24)"
+fi
 # Build heredoc variables
 KCONFIG=$( cat <<EOC
 # SPDX-License-Identifier: GPL-2.0-only
@@ -44,7 +46,9 @@ echo "$KCONFIG"
 echo
 echo "and Makefile"
 echo "$MAKEFILE"
-echo "Commit msg: $COMMIT"
+if [ git rev-parse --git-dir &>/dev/null ]; then
+    echo "Commit msg: $COMMIT"
+fi
 echo "Ctrl+c to quit, any other key to continue"
 read CANCEL
 # Execute copy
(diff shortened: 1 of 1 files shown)


Adam-pi3:
Hm... but should we thrown an error and break the entire process if git is not installed? Otherwise you won't know that you didn't execute commands which you wanted to. What am I missing?

0xC0ncord:
> Hm... but should we thrown an error and break the entire process if git is not installed? Otherwise you won't know that you didn't execute commands which you wanted to. What am I missing?

That might be the better approach. The use case for this is when you are adding LKRG to a source tree that isn't a git repository. In my particular case, I would like to be able to add a Portage hook on Gentoo to do this with the kernel releases. The script still works, but it generates noise.

scripts/copy-builtin.sh
```diff
@@ -6,7 +6,9 @@ KDIR="${KDIR:="/usr/src/linux"}"
 LDIR="$KDIR/security/lkrg"
 BASEDIR="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
 SDIR="$BASEDIR/../src"
-COMMIT="LKRG in-tree @ $(git log|head -1|cut -d' ' -f2|cut -c 1-24)"
+if [ git rev-parse --git-dir 2>/dev/null ]; then
```
solardiz, line 9: Wouldn't the three lines like this each print `.git` to stdout? Wouldn't that be weird?

solardiz:
Contrary to what Adam wrote, shouldn't we instead make this script git-unaware? Just why would it need to make a git commit? If the tree is managed by git, then git will also detect that there are uncommitted changes, and it'd be up to the user to commit those or not.

scripts/copy-builtin.sh
0xC0ncord: You're right. Let's fix that.

0xC0ncord:
> Contrary to what Adam wrote, shouldn't we instead make this script git-unaware? Just why would it need to make a git commit? If the tree is managed by git, then git (edit: if and when manually run by the user later) will also detect that there are uncommitted changes, and it'd be up to the user to commit those or not.

I think I agree with this take.

solardiz:
> I think I agree with this take.

Should we first merge this PR as-is, and then you send us another removing the git-awareness?

solardiz:
> send us another removing the git-awareness?

Actually, we should probably expect that (or maybe an objection) from @sempervictus since this was his contribution.

solardiz:
I'm going to merge this now, but I would like to hear from @sempervictus on making the script git-unaware with a separate PR.

sempervictus:
Thank you, and i'm no contrib snob - please feel free to merge any changes which you see fit atop any work i submit. Worst case i'll fix it.

solardiz:
@sempervictus So what do you think of dropping git support from this script?

sempervictus:
My take is to safety the calls - `if [ -z "$(which git)" ]; then ...`
Leaves the original functionality intact, handles folks dealing with archive-sourced builds (or `rm -rf .git` types)

sempervictus:
PR'd

PR #51, opened 2021-02-23
Copy-builtin: address lack of git or .git

sempervictus:
Some systems do not have the git binary or in the current PATH.
Concurrently, some users do not have git source control for LKRG
or their kernel tree, making the assumptive git commands useless.

Address by way of conditional checking for presence of the git bin
in the current PATH as well as the existence of a git revision in
the CWD.

### DIFF
--- a/scripts/copy-builtin.sh
+++ b/scripts/copy-builtin.sh
@@ -6,8 +6,10 @@ KDIR="${KDIR:="/usr/src/linux"}"
 LDIR="$KDIR/security/lkrg"
 BASEDIR="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
 SDIR="$BASEDIR/../src"
-if [ git rev-parse --git-dir &>/dev/null ]; then
+if [ -n "$LKRG_USE_GIT" && git rev-parse --git-dir &>/dev/null ]; then
     COMMIT="LKRG in-tree @ $(git log|head -1|cut -d' ' -f2|cut -c 1-24)"
+else
+    COMMIT="LKRG in-tree @ $(date +%Y%m%d%H%M)"
 fi
 # Build heredoc variables
 KCONFIG=$( cat <<EOC
@@ -46,9 +48,7 @@ echo "$KCONFIG"
 echo
 echo "and Makefile"
 echo "$MAKEFILE"
-if [ git rev-parse --git-dir &>/dev/null ]; then
-    echo "Commit msg: $COMMIT"
-fi
+echo "Commit msg: $COMMIT"
 echo "Ctrl+c to quit, any other key to continue"
 read CANCEL
 # Execute copy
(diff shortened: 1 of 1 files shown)


Adam-pi3:
LGTM
@solardiz @0xC0ncord any objections from your side?

0xC0ncord:
LGTM

solardiz:
> any objections from your side?

Sort of:

1. I see absolutely no reason for this script to include usage of `git` in it, except that @sempervictus mysteriously wants it in there. I think this is a misfeature, which we should ideally drop. I don't strictly object, though.
2. The `which` commands would be printing stuff to the script's stderr, so `2>/dev/null` is needed there.
3. `! -z` could be simplified to `-n`.
4. No reason to check the program's string output when it also returns an exit code directly. In other words, there's actually no need for either `!  -z` nor `-n`.
5. `which` is an external program. Since the script relies on `bash` anyway, we can instead use `type`, which is a built-in.

Anyway, this is mostly a bikeshed, so I think we should either drop it as a misfeature or not spend too much time polishing it.

0xC0ncord:
>5. which is an external program. Since the script relies on bash anyway, we can instead use type, which is a built-in.

Just want to point out that we could also use `command -v` which is also built-in and functionally equivalent to `which` for this kind of usage.

sempervictus:
@solardiz - the reason for using git this way is to keep track of commit hashes in downstream projects. Unless LKRG starts doing detailed versioning and releasing point versions at short and regular intervals, downstream needs to know which commit they merged into their code. Far as using built-ins in shell scripts, you're totally right, i tend to just muscle-memory these things because they're utility scripts, but formally they really should be tighter on those calls.
I'll clean it up.

solardiz:
> the reason for using git this way is to keep track of commit hashes in downstream projects.

This is a reason to record the commit hash of the LKRG tree somewhere (e.g., in a file in the copied tree or simply not to exclude LKRG's `.git`, if present, from the copy), but not to automatically make a new git commit.

sempervictus:
> > the reason for using git this way is to keep track of commit hashes in downstream projects.
> 
> This is a reason to record the commit hash of the LKRG tree somewhere (e.g., in a file in the copied tree or simply not to exclude LKRG's `.git`, if present, from the copy), but not to automatically make a new git commit.

The new commit is being made in the target repo to complete the in-tree transaction, not in LKRG's repo.
I have scripts like this for a ton of out-of-tree stuff (PF, intel, accel-ppp, xtables, etc), each one gets gated in as a block of effort in a single commit this way. You can see how this plays out in https://github.com/vyos/vyos-build/pull/132
If this doesn't line up with how upstream wants to do it, i can definitely rewrite it to just pull the commit has to a file (although having plaintext files in the tree seems a bit odd to me when metadata about the tree is kept in version control). Dealer's choice sir - tell me how your preferred implementation would look, and i'll update it for the users.

0xC0ncord:
As a user I don't see the harm in having this functionality, but maybe in order to avoid potentially unexpected behavior, the script should instead ask the user if they would like to commit the inclusion of LKRG in-tree instead of doing it automatically?

solardiz:
Looks like functionality to support @sempervictus' convention for merging out-of-tree stuff. Perhaps this should be enabled with a command-line option to the script? Perhaps even undocumented, since there's exactly one user for it. ;-)

> having plaintext files in the tree seems a bit odd to me when metadata about the tree is kept in version control

Right. So maybe by default it's sufficient that `lkrg/.git` will get copied along with the rest of the sub-tree, with no git-awareness needed from the script for this to happen?

solardiz:
@sempervictus Please rename `USEGIT` to `LKRG_USE_GIT` and squash the commits. Thank you!

sempervictus:
Thanks boss

PR #52, opened 2021-03-02
Fix compilation fo kernel 5.11 on arm64

wladmis:
Since kernel commit 3d2403fd10a1dbb359b154af41ffed9f2a7520e8 there is no
more addr_limit on arm64.


Adam-pi3:
I've reviewed (https://github.com/torvalds/linux/commit/3d2403fd10a1dbb359b154af41ffed9f2a7520e8#diff-cbc0f5360075634c9a212308daa3baab1eed2294d5fc020e45b940b9d8f24b3f) and LGTM. I've verified that ARM still has it.

PR #53, opened 2021-03-02
copy-builtin: fix syntax error

0xC0ncord:
This fixes the following error when using bash:
bash: [: missing `]'


solardiz (approved):
Oh wow. Sometimes review alone can miss obvious things.

PR #54, opened 2021-03-03
Improve shell script

wladmis:
$subj


Adam-pi3:
LGTM, @solardiz what do you think?

solardiz:
Looks good to me, too. However, I think there are more improvements to make (separately):

1. We could `exec` the other script, so that we don't unnecessarily keep an extra shell process around while it's running. We can do this because we don't do anything upon return from the other script anyway.
2. All of the `P_PWD` stuff looks unnecessary to me, in both scripts (it would be needed if the current directory were changed after obtaining `pwd`, but I don't see anything like that in there). Instead of `$P_PWD/$P_LKRG_SYSTEMD` we could use simply `$P_LKRG_SYSTEMD` (would be `exec "$P_LKRG_SYSTEMD"`). And similar in the other script. The `P_PWD` variable should be gone from both.

@wladmis You might want to implement and test the above changes and send us another PR with those. Thank you!

PR #57, opened 2021-03-17
Rework selinux architecture

oshogbo:
### Description
* Rework the SELinux architecture.
Separate SELlinux to a new file
Introduce API for SELinux related orations.
Deduplicat code.

* Add support for kernels with rand structs (CONFIG_GCC_PLUGIN_RANDSTRUCT option).
In this case we disable SELinux monitoring.

* Add support for kernels newer then 5.6 the (CONFIG_SECURITY_SELINUX_DISABLE option)
Reflect changes in the selinux_status struct.

### How Has This Been Tested?
On ubuntu machine with custom 5.11.6 kernel build.
I used p_vuln to implement overwriting the selinux_status struct fields, and checked if the kernel will panic and if the original value will be restored.

### DIFF
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -1684,12 +1684,11 @@ static void p_validate_selinux(void) {
 #endif
 
 #ifdef CONFIG_SECURITY_SELINUX_DEVELOP
- #if LINUX_VERSION_CODE >= KERNEL_VERSION(4,17,0)
-   if (p_ed_guard_globals.p_selinux.p_selinux_state.enforcing != P_SYM(p_selinux_state)->enforcing) {
+   if (p_selinux_state_changed()) {
       p_print_log(P_LKRG_CRIT,
              "<Exploit Detection> Detected data corruption against SELINUX! 'selinux_state->enforcing' has "
                                   "different value [%d vs %d] than expected!\n",
-                                  P_SYM(p_selinux_state)->enforcing,p_ed_guard_globals.p_selinux.p_selinux_state.enforcing);
+                                  p_selinux_state_enforcing(), p_ed_guard_globals.p_selinux.p_selinux_enforcing);
 
       switch (P_CTRL(p_kint_enforce)) {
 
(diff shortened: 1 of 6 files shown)


src/modules/exploit_detection/p_exploit_detection.c
```diff
@@ -1701,46 +1700,18 @@ static void p_validate_selinux(void) {
 
          case 1:
            p_print_log(P_LKRG_CRIT, "Original value will be restored!\n");
-           P_SYM(p_selinux_state)->enforcing = p_ed_guard_globals.p_selinux.p_selinux_state.enforcing;
-           break;
+           p_lkrg_counter_lock_unlock(&p_ed_guard_globals.p_selinux_lock, &p_flags);
```
Adam-pi3, line 1703: Why do you unlock the lock here? I think it's wrong.

Adam-pi3 (changes requested):

src/modules/exploit_detection/p_exploit_detection.c
oshogbo: Yes, this was from some other experiment sorry.

solardiz:
Suggested replacement commit messages, mostly just fixing typos:

```
SELinux: Refactor state monitoring

Introduce a new API for SELinux state monitoring.
Move the API to a separate file.
```

```
SELinux: Recognize that the disable parameter is optional

Starting with Linux 5.6 a new config option was introduced:
CONFIG_SECURITY_SELINUX_DISABLE. Reflect in LKRG selinux_struct
that the "disabled" flag is optional for these newer kernels.
```

```
SELinux: Disable state monitoring for kernels with rand structs

New Linux kernels may be built with the CONFIG_GCC_PLUGIN_RANDSTRUCT
option. This is option randomly changes the order of fields in structures.
The selinux_state structure has this attribute. Currently, LKRG isn't capable to
recreate the structure layout.

To make LKRG available on kernels with this option, we disable
SELinux monitoring.

CONFIG_GCC_PLUGIN_RANDSTRUCT was introduced to make it harder
for attackers to overwrite particular fields of the structure.
LKRG's goal was the same. So even disabling LKRG's monitoring,
we still have some mitigations for the selinux_state structure.

We expect future work to make LKRG capable to recreate randomized
structures.
```

src/p_lkrg_main.h
```diff
@@ -112,6 +112,10 @@
  #define P_LKRG_MARKER2 0xdeadbabe
 #endif
 
+#ifdef CONFIG_GCC_PLUGIN_RANDSTRUCT
+#undef CONFIG_SECURITY_SELINUX_DEVELOP
+#endif
```
solardiz, line 117: I don't really mind this hack, but I wonder if the `#undef CONFIG_SECURITY_SELINUX_DEVELOP` can possibly cause inconsistencies in kernel headers if we include any further ones after this point. Ideally, we'd introduce and use our own macro for whether we're monitoring SELinux state. I think this would also let us avoid nested `#if` in a few places.

oshogbo:
@solardiz I have changed the commit messages.

src/p_lkrg_main.h
oshogbo: Right now most of our includes are in single file and p_lkrg_main.h is the last one that we include - so i'm not sure if this is an issue for now.
But in general terms I totally agree that this would be a cleaner solution.
TBH I'm not sure if we would avoid nested `#if` . Do you have any particular one in mind?

Adam-pi3: We already do similar checks e.g., here:
https://github.com/openwall/lkrg/blob/main/src/modules/exploit_detection/p_exploit_detection.h#L64

Similar work can be done in SELinux file where you define e.g., P_SELINUX_VERIFY variable which depends on CONFIG_SECURITY_SELINUX_DEVELOP and CONFIG_GCC_PLUGIN_RANDSTRUCT. Later you rely on P_SELINUX_VERIFY instead of CONFIG_SECURITY_SELINUX_DEVELOP

oshogbo: Yes I'm understand the idea - I just not sure which part of code it will remove nested `#if`s ?

solardiz: I took a closer look, and I now think I was wrong - this won't let us remove nested `#if`s, at least not on its own.

solardiz (approved):
I think this is OK to merge. Adam?

src/p_lkrg_main.h
Adam-pi3: I still think it's a good idea to introduce P_SELINUX_VERIFY. @oshogbo can you do it?

Adam-pi3:
> I think this is OK to merge. Adam?

I think current version is OK. As soon as @oshogbo introduces P_SELINUX_VERIFY we can merge it.

src/p_lkrg_main.h
oshogbo: Sure. Done.

Adam-pi3 (approved):

solardiz:
> On ubuntu machine with custom 5.11.6 kernel build.
> I used p_vuln to implement overwriting the selinux_status struct fields, and checked if the kernel will panic and if the original value will be restored.

@oshogbo You might want to rerun these tests for the merged code now, as it's slightly different. Thanks!

PR #62, opened 2021-04-07
CI: Boot test using Ubuntu mainline kernel build

vt-alt:
```
Ubuntu have daily builds of mainline kernels at
  https://wiki.ubuntu.com/Kernel/MainlineBuilds
They are built on groovy.
```
Example actions https://github.com/vt-alt/lkrg/actions

I decided to create 2 workflows (instead of single one with 3 element matrix, that I tried too), (**pros**) because -- this will allow to disable one of workflows (for example mainline one) from GA web UI; and allows to run mainline workflow also from cron schedule (daily). -- (**cons**) This will create two "lines" in Actions for every commit -- this is bad UI design, because single run is not grouped, but cant be helped (every other repo with complex workflows have that); and this is code duplication, but this is overall style of GA and cant be helped too.

I don't understand completely who will receive email notifications from cron builds - some documentation says that only me (last person who edited 'cron' line), which is not what is desiring.

Lets wait couple of days before merging (in case there is fixes).

### DIFF
--- a/.github/workflows/mkosi-mainline.yml
+++ b/.github/workflows/mkosi-mainline.yml
@@ -1,20 +1,18 @@
-name: mkosi boot (ubuntu)
+name: mkosi boot (mainline)
 
-on: [push, pull_request]
+on:
+    push:
+    pull_request:
+    schedule:
+        - cron: '0 10 * * *'
 
 jobs:
-    mkosi-boot:
+    mkosi-mainline:
         runs-on: ubuntu-20.04
-        strategy:
-            fail-fast: false
-            matrix:
-                release:
-                    - focal
-                    - groovy
         steps:
             - uses: actions/checkout@v2
             - run: sudo apt-get update
-            - run: sudo apt-get install -y debootstrap qemu-system-x86 systemd-container expect
+            - run: sudo apt-get install -y debootstrap qemu-system-x86 systemd-container expect sysvbanner
             - name: Install mkosi from git
               # Native focal package seems to be too old (v5) and not even
               # able to build images properly.
@@ -23,7 +21,9 @@ jobs:
                   echo /usr/local/bin >> $GITHUB_PATH
 
             - name: Create bootable image using mkosi
-              run: sudo mkosi -r ${{ matrix.release }} --cache=mkosi.cache
+              run: |
+                  .github/workflows/ubuntu-kernel-daily.sh
+                  sudo mkosi -r groovy -p '!linux-virtual' --cache=mkosi.cache --prepare-script=.github/workflows/dpkg-i.sh
 
             - name: Boot image on qemu
               run: |
@@ -44,7 +44,7 @@ jobs:
                   grep 'Linux version' boot.log
                   grep 'LKRG initialized successfully' boot.log
             - name: Check that boot.log does not contain problems
-              run: egrep 'Panic|BUG:|WARNING:|Oops|Call Trace' boot.log && false || true
+              run: "! egrep 'Panic|BUG:|WARNING:|Oops|Call Trace' boot.log"
             - name: Check that boot.log contains shutdown sequence
               run: |
                   grep 'Reached target.*Power-Off' boot.log
--- a/.github/workflows/dpkg-i.sh
+++ b/.github/workflows/dpkg-i.sh
@@ -1,0 +1,9 @@
+#!/bin/bash -eux
+# Non-default mkosi.prepare script (--prepare-script=) to install
+# mainline kernel debs.
+# (It does have network and can run additional installs if needed.)
+
+banner "dpkg -i" >&2
+cd /root/src
+ls -l linux-*.deb
+dpkg -i linux-*.deb
--- a/.github/workflows/ubuntu-kernel-daily.sh
+++ b/.github/workflows/ubuntu-kernel-daily.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# Download daily Ubuntu mainline kernel from
+#   https://wiki.ubuntu.com/Kernel/MainlineBuilds
+#
+# Copyright (c) 2021 Vitaly Chikunov <vt@altlinux.org>.
+#
+
+baseurl="https://kernel.ubuntu.com/~kernel-ppa"
+for listurl in \
+	"$baseurl/mainline/daily/" \
+	"$baseurl/mainline/"
+do
+	echo >&2 "List $listurl"
+	curl -s "$listurl" \
+	| grep -Eio 'href="[^"]+"'   \
+	| grep -Eo '"[^"]+"'         \
+	| grep -Po '[v2][rc\.\d-]+'  \
+	| sort -Vr \
+	| while read subdir; do
+		url="$listurl$subdir/amd64/"
+		echo >&2 "Trying $url"
+		if page=$(curl -s --fail "$url"); then
+			banner $subdir >&2
+			# Show Ubuntu commit and build status.
+			curl --fail "$url/aggregate.yaml" || curl "$url/summary.yaml"
+			echo "$page" \
+			| grep -Eio 'href="[^"]+"' \
+			| grep -o "linux.*deb"     \
+			| grep -v "lowlatency"     \
+			| while read deb; do
+				if [ ! -e "$deb" ]; then
+					echo >&2 "Download $url$deb"
+					curl -O "$url$deb"
+				fi
+			done
+			# Signal success to upper shell.
+			exit 22
+		fi
+	done
+	# Exit if subshell succeeded.
+	[ $? -eq 22 ] && exit
+done
--- a/.gitignore
+++ b/.gitignore
@@ -20,5 +20,6 @@ GRTAGS
 GTAGS
 
 # mkosi
+.mkosi-*
 mkosi.cache
 image.raw*
--- a/mkosi.build
+++ b/mkosi.build
@@ -1,10 +1,10 @@
-#!/bin/bash -eu
+#!/bin/bash -eux
+# mkosi.build - Compile and install module into DESTDIR.
 
-KERNELRELEASE=$(ls -d /lib/modules/*)
+KERNELRELEASE=$(ls -d /lib/modules/* | sort -V | tail -1)
 KERNELRELEASE=${KERNELRELEASE##/lib/modules/}
 export KERNELRELEASE
 
-banner build
-set -x
+banner build $KERNELRELEASE >&2
 make -j$(nproc)
 install -Dpm 644 p_lkrg.ko $DESTDIR/lib/modules/$KERNELRELEASE/extra/p_lkrg.ko
--- a/mkosi.default
+++ b/mkosi.default
@@ -1,9 +1,11 @@
+# mkosi.default - Another way to run a test system with LKRG module.
+
 [Distribution]
 Distribution=ubuntu
 # Linux version per release:
 #   focal  - v5.4
 #   groovy - v5.8
-Release=focal
+Release=groovy
 
 [Output]
 Bootable=yes
@@ -15,16 +17,19 @@ WithUnifiedKernelImages=no
 BootProtocols=bios
 
 [Partitions]
-RootSize=2G
+# 2G was not enough for build with mainline kernels installs.
+RootSize=3G
 
 [Packages]
 BuildPackages=
 	diffutils
 	gcc
 	make
+	linux-base
 	linux-virtual
 Packages=
 	sysvbanner
+	kmod
 
 [Host]
 QemuHeadless=yes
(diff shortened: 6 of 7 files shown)


.github/workflows/mkosi-boot.yml
```diff
@@ -23,15 +23,15 @@ jobs:
...
 
-            - name: Boot image using on qemu
+            - name: Boot image on qemu
               run: |
                   sudo expect <<- EOF | tr -d '\r' | tee boot.log
                   proc abort {} { send_error "ABORT"; exit 1 }
                   spawn mkosi qemu
                   expect "Booting" { send "\r" }
-                  set timeout 120
+                  set timeout 300
```
vt-alt, line 34: I one got timeout on this, thus increased the value.

vt-alt:
Interestingly, it shows here that it tried all new workflows even on this pull request.

.github/workflows/mkosi-boot.yml
```diff
@@ -40,8 +40,14 @@ jobs:
               shell: bash -eo pipefail {0}
 
             - name: Check boot.log for 'LKRG initialized successfully'
-              run: grep 'LKRG initialized successfully' boot.log
+              run: |
+                  grep 'Linux version' boot.log
+                  grep 'LKRG initialized successfully' boot.log
             - name: Check that boot.log does not contain problems
               run: egrep 'Panic|BUG:|WARNING:|Oops|Call Trace' boot.log && false || true
```
solardiz, line 47: Not changed in this PR, but FWIW I doubt this works as desired:

```shell
egrep 'Panic|BUG:|WARNING:|Oops|Call Trace' boot.log && false || true
```

Here's my test:

```shell
$ false && false || true
$ echo $?
0
$ true && false || true
$ echo $?
0
```

.github/workflows/mkosi-mainline.yml
```diff
@@ -0,0 +1,53 @@
+name: mkosi boot (mainline)
+
+on:
+    push:
+    pull_request:
+    schedule:
+        - cron: '0 0 * * *'
```
solardiz, line 7: Maybe schedule this to a weirder time, so that we don't try to run it when many others also run theirs?

solardiz (approved):
No objections to this PR as-is, but I made related comments.

Maybe the `mkosi*` files should be in a subdirectory like `ci/` if our only use of them is for CI.

Also, should we be including any CI stuff in LKRG release tarballs or should we exclude them perhaps in `.gitattributes`?

Also, Adam had added a top-level `test.sh` script - perhaps that one should also be in a subdirectory?

solardiz:
@vt-alt I'm OK merging this first, then addressing my comments separately. What do you prefer? Thanks!

vt-alt:
Let's not merge first.

vt-alt:
> Maybe the `mkosi*` files should be in a subdirectory like `ci/` if our only use of them is for CI.

I thought `mkosi` is usable on itself, similar to how you use `out-of-tree`. That's why I made it into separate commit with appropriate, as I thought, commit message. But, if you reject that usage I can move them into subdirectory, just tell.

> Also, should we be including any CI stuff in LKRG release tarballs or should we exclude them perhaps in `.gitattributes`?

I don't think that matters.

.github/workflows/mkosi-mainline.yml
vt-alt: Ouchie.

solardiz:
> I thought `mkosi` is usable on itself, similar to how you use `out-of-tree`. That's why I made it into separate commit with appropriate, as I thought, commit message. But, if you reject that usage I can move them into subdirectory, just tell.

OK, this makes sense. Maybe have a subdirectory called `mkosi`, then?

> > Also, should we be including any CI stuff in LKRG release tarballs or should we exclude them perhaps in `.gitattributes`?
> 
> I don't think that matters.

OK. Context to my question:
Somehow @ldv-alt excluded his GitHub Actions scripts from passwdqc releases via `.gitattributes`, so the passwdqc 2.0.2 release I made recently doesn't have them in its tarball. There, this exclusion also makes sense in that it simplifies passwdqc license (nothing GPL'ed in tarballs), which is part of the reason why I didn't mind the exclusion. LKRG is GPL'ed anyway, and some of this can also be reused in its forks, so I think is better to include (although ideally any forks would be based on git and not on a release tarball).

Regarding out-of-tree, looks like this framework itself expects the script to be called simply `test.sh` and be at top level, correct @Adam-pi3? If so, we cannot rename or move the script, but we can and should add a comment into it explaining what it is for (something I'd like Adam to do separately, not in this PR).

vt-alt:
> OK, this makes sense. Maybe have a subdirectory called `mkosi`, then?

It's supposed to be top-level. Like person just run mkosi in the top-level dir of a project and gets bootable image. You don't like this? You have `.out-of-tree` top-level too, for example. And it's top-level in `systemd` projects (it's not top-level in `kmod` tests for example, but they have it's use test only).

solardiz:
@vt-alt OK, I'm convinced.

vt-alt:
> Somehow @ldv-alt excluded his GitHub Actions scripts from passwdqc releases via `.gitattributes`, so the passwdqc 2.0.2 release I made recently doesn't have them in its tarball. There, this exclusion also makes sense in that it simplifies passwdqc license (nothing GPL'ed in tarballs), which is part of the reason why I didn't mind the exclusion. LKRG is GPL'ed anyway, and some of this can also be reused in its forks, so I think is better to include (although ideally any forks would be based on git and not on a release tarball).

I don't think any file should be excluded, it's just part of the source. You can think of it as example of CI if it's not directly usable.

vt-alt:
I thought about `mkosi` as "better" *alternative* to `out-of-tree`, since that requires `docker` and `mkosi` doesn't. (Only downside is it's have ubuntu-specific config [now]; and I never used `out-of-tree`.)

solardiz:
> I don't think any file should be excluded, it's just part of the source. You can think of it as example of CI if it's not directly usable.

I agree. I was surprised @ldv-alt's preference was to exclude.

ldv-alt:
On Wed, Apr 07, 2021 at 09:23:14AM -0700, Solar Designer wrote:
> > I don't think any file should be excluded, it's just part of the source. You can think of it as example of CI if it's not directly usable.
> 
> I agree. I was surprised @ldv-alt's preference was to exclude.

There are two reasons.  Firstly, there is little use for passwdqc CI
scripts outside the git repository.  Secondly, passwdqc CI scripts
were forked from strace CI covered by a copyleft license.

Adam-pi3:
> Regarding out-of-tree, looks like this framework itself expects the script to be called simply test.sh and be at top level, correct @Adam-pi3?

From my understanding yes. @jollheef is that  correct?

vt-alt:
I force-pushed changes regarding Solar's review. Cron time changed from `0` to `10`. `... && false || true` bug is fixed with `"! ..."` (which I intended before). I also added `-r groovy` to `mkosi` run so we can change `mkosi.default` (for hypothetical `mkosi` users) w/o changing CI. Tests reported green. This could be merged or we can wait until cron build triggers (never seen it myself).

vt-alt:
> Tests reported green. This could be merged or we can wait until cron build triggers (never seen it myself).

Sec. I found a slightly strange behavior. LKRG should not compile for upstream 5.12-rc, but it is in this test. I am investigating.

vt-alt:
> > Tests reported green. This could be merged or we can wait until cron build triggers (never seen it myself).
> 
> Sec. I found a slightly strange behavior. LKRG should not compile for upstream 5.12-rc, but it is in this test. I am investigating.

Ah, this is commit 523dd2c5b07af6f6dd5116219ae6083c74d3bd9d I somehow missed it in my local tests (since I started experimenting on Mar 21 and never git pulled again). So there are no problems and it could be merged!

solardiz:
> Cron time changed from `0` to `10`.

I'd also change the minutes. Please do along with next update.

Anyway, I'll merge this now.

vt-alt:
> I'd also change the minutes. Please do along with next update.

No need actually, GA delays cron jobs depending on the load.

jollheef:
> > Regarding out-of-tree, looks like this framework itself expects the script to be called simply test.sh and be at top level, correct @Adam-pi3?
> 
> From my understanding yes. @jollheef is that correct?

Yes, that's correct.

jollheef:
@solardiz 
> Regarding out-of-tree, looks like this framework itself expects the script to be called simply `test.sh` and be at top level, correct @Adam-pi3? If so, we cannot rename or move the script, but we can and should add a comment into it explaining what it is for (something I'd like Adam to do separately, not in this PR).

It's also possible to have a test script as ${KERNELMODULEPATH}_test. [Check the example](https://github.com/jollheef/out-of-tree/blob/master/examples/kernel-module/Makefile#L14)

vt-alt:
JFYI. There is appeared [the first cron build](https://github.com/openwall/lkrg/actions/runs/732660619) in [Actions](https://github.com/openwall/lkrg/actions):
```
mkosi boot (mainline)
mkosi boot (mainline) #4: Scheduled
 21 hours ago
 9m 33s
```
So, it's working. But, I never received any email about it.

vt-alt:
I never received any new build failure email from GA, but today I checked the Actions page and found there are two failures:
* https://github.com/openwall/lkrg/runs/2361227895?check_suite_focus=true
This one is triggered lkrg somehow
```
[   42.567193] [p_lkrg] ALERT !!! MODULE'S <kvm> HASH IS DIFFERENT it is [0x2a58fa680c88611] and should be [0x80fe0107fb787c51] !!!
[   42.567193] [p_lkrg] [KOBJ] ALERT !!! MODULE'S <kvm> HASH IS DIFFERENT it is [0x2a58fa680c88611] and should be [0x80fe0107fb787c51] !!!
[   42.567193] [p_lkrg] ALERT !!! SYSTEM HAS BEEN COMPROMISED - DETECTED DIFFERENT 1 CHECKSUMS !!!
```
* https://github.com/openwall/lkrg/runs/2332324643?check_suite_focus=true
And this is of unknown cause. (Maybe timeout?)

solardiz:
> I never received any new build failure email from GA

I did. The first time I took a look - and like you say, the cause was unclear and could be a timeout. The second time I decided that if the next time the check passes, I wouldn't bother looking. Thank you for looking!

I think we need to open a separate issue for the `MODULE'S <kvm> HASH IS DIFFERENT` on shutdown. I will.

vt-alt:
@solardiz Thanks for confirming that emails go in the right direction!

solardiz:
> GA delays cron jobs depending on the load.

Yet we could help by not trying to schedule to a time where the delaying would be needed. I see that so far the job ran at 10:05 or 10:06, so perhaps it was started when the load barely permitted that. At your convenience and perhaps along with some other change, please reschedule to a time like 10:12. Thanks!

vt-alt:
> At your convenience and perhaps along with some other change, please reschedule to a time like 10:12. Thanks!

No.

vt-alt:
See https://docs.github.com/en/actions/guides/about-continuous-integration#notifications-for-workflow-runs

> _"Notifications for scheduled workflows are sent to the user who initially created the workflow. If a different user updates the cron syntax in the workflow file, subsequent notifications will be sent to that user instead. If a scheduled workflow is disabled and then re-enabled, notifications will be sent to the user who re-enabled the workflow rather than the user who last modified the cron syntax."_

So perhaps you should update it so it doesn't start to send me notifications (by some obscure logic).

vt-alt:
> I did. The first time I took a look - and like you say, the cause was unclear and could be a timeout. The second time I decided that if the next time the check passes, I wouldn't bother looking. Thank you for looking!

Now I looked more closely into this. it seems timeout, but on the second boot, (because there was no `-no-reboot` option qemu rebooted guest instead of just terminating). And the first (reboot) cause is
```
root@localhost:~# systemctl poweroff
root@localhost:~#          Stopping Session 1 of user root.
         Stopping Session 2 of user root.
[  OK  ] Removed slice system-modprobe.slice.
[   42.855599] [p_lkrg] ALERT !!! MODULE'S <kvm> HASH IS DIFFERENT it is [0x3fd1877442241f75] and should be [0x75de4a171b5ae5b9] !!!
[   42.855599] [p_lkrg] ALERT !!! MODULE LIST HASH IS DIFFERENT !!! - it is [0x9735e31bd42bba70] and should be [0x83a61ef86e92c546] !!!
[   42.855599] [p_lkrg] ALERT !!! MODULE KOBJ HASH IS DIFFERENT !!! - it is [0xe29295fe81ac0726] and should be [0xac4388a327c3c2ee] !!!
[   42.855599] [p_lkrg] [KOBJ] ALERT !!! MODULE'S <kvm> HASH IS DIFFERENT it is [0x3fd1877442241f75] and should be [0x75de4a171b5ae5b9] !!!
[   42.855599] [p_lkrg] ALERT !!! SYSTEM HAS BEEN COMPROMISED - DETECTED DIFFERENT 4 CHECKSUMS !!!
[   42.855599] Kernel panic - not syncing: [p_lkrg] Kernel Integrity verification failed! Killing the kernel...
```

Thus, it's the same issue all the time. We should add `-no-reboot` to qemu run, though.

PR #68, opened 2021-04-18
Add support for Debian pbuilder/dpkg-buildpackage

morfikov:
Basically the `debian/` dir allows users to build the LKRG dkms packages via `pbuilder`/`dpkg-buildpackage` on Debian based distos, which can then be installed in the system using `apt`/`apt-get`/`aptitude/dpkg`. Since `lintian` doesn't complain much, this package could be sent to the Debian main repo, but I don't really know how to do this, so at the moment the files in question can only help to build the Debian packages locally.

As a result of the building process, there are 3 packages:
 - `lkrg`         -- this one is a virtual package, and it's purpose is just to install the `lkrg-dkms` package. This package is created for a better user experience, since people (including myself) tend to install LKRG just by issuing `apt-get install lkrg` , but all DKMS modules in Debian have to have the `-dkms` suffix.
 - `lkrg-dkms`    -- this is the package that installs the LKRG source and build the kernel module from it using the DKMS mechanism.
 - `lkrg-systemd` -- this is the package that installs LKRG systemd's service. This one probably could/should be removed. (see below for explanation).

When a user installs just the `lkrg` package, it will also pull `lkrg-dkms` and (depending on the system configuration) `lkrg-systemd`. The `lkrg-systemd` package is set as recommended, since not all systems are systemd-based. When a user installs `lkrg-dkms` , it won't install the virtual `lkrg` package. When a user installs `lkrg-systemd` , it will pull only the `lkrg-dkms`.

After the `lkrg-dkms` is installed, the sources will be placed under the `/usr/src/lkrg-*/ dir`, and the compilation process will take place. The package `lkrg-dkms ` depends on `dkms` , so there's no need for any other deps to be specified. The DKMS configuration doesn't have to be updated with every LKRG release so the module could be registered with proper version. It automatically picks up the version based on the one that was set in the `changelog` file.

The systemd's service file that comes with the `lkrg-systemd` package is a little bit useless. The service loads and unloads the LKRG module. The loading functionality can be achieved via the files in the `/etc/modules-load.d/` dir. All options that should be set at loading time can also be passed using the `/etc/modprobe.d/` dir. The configuration of the module can be set via the files in the `/etc/sysctl.d/` dir. As LKRG is a kernel module, it should be loaded as soon as possible when system boots, and in the case of systemd, the `systemd-modules-load.service` and `systemd-sysctl.service` take care of loading and configuring the kernel modules. Also, there's no need to unload the loaded kernel modules at system shutdown. That's why the systemd service isn't really necessary and I think it should be dropped altogether. When a user wants to remove the kernel module, he can simply issue the `modprobe -r` command. There's no real advantage of having the LKRG systemd's service. Besides, other systems (inits) also have to load and configure the module, and systemd's service file is of no use in those cases, so some more general approach is needed. I don't know whether other inits use the files in the `/etc/modules-load.d/` , `/etc/modprobe.d/` and `/etc/sysctl.d/` dirs, but systemd does.

The systemd's service is installed but it's not activated at the moment. A user can enable the service manually, but it's not necessary.

The `lintian` tool reports the following messages:

```
$ lintian -EviIL +pedantic lkrg_0.9.0-1_amd64.changes
N: Using profile debian/main.
N: Starting on group lkrg/0.9.0-1
N: Finished processing group lkrg/0.9.0-1
I: lkrg-systemd: package-supports-alternative-init-but-no-init.d-script lib/systemd/system/lkrg.service
N:
I: package-supports-alternative-init-but-no-init.d-script
N:
N:   The package provides daemon, but contains no init.d script Packages
N:   that provide services (daemons), like cron daemon or web servers, must
N:   provide init.d script for starting that services with sysvinit.
N:   Optionally, packages can also provide integration with alternative
N:   init systems.
N:
N:   Package in question provides integration with some alternative init
N:   system, but corresponding init.d script is absent.
N:
N:   See init-d-script(5) for one of possible ways writing init.d scripts.
N:
N:   Refer to Debian Policy Manual section 9.11 (Alternate init systems)
N:   for details.
N:
N:   Severity: info
N:
N:   Check: init.d
N:
I: lkrg-systemd: systemd-service-file-missing-documentation-key lib/systemd/system/lkrg.service
N:
I: systemd-service-file-missing-documentation-key
N:
N:   The systemd service file does not contain a Documentation key.
N:
N:   Documentation for systemd service files can be automatically viewed
N:   using systemctl help servicename if this field is present.
N:
N:   Refer to the systemd.unit(5) manual page for details.
N:
N:   Severity: info
N:
N:   Check: systemd
N:
```

So basically it's a clean package with two exceptions concerning the LKRG systemd's service. But if you decide to drop the LKRG systemd's service and go with the `/etc/modules-load.d/` , `/etc/modprobe.d/` and `/etc/sysctl.d/` dirs, there will be no `lintian` messages at all.

I'm not sure whether all rules in the `rules` file under `override_dh_install:` are good, since Debian Helper has the [dh_installmodules](https://manpages.debian.org/unstable/debhelper/dh_installmodules.1.en.html) target, which can only support `debian/package.modprobe` files. The other files `.modules` and `.sysctl` were created manually based on other packages, and the rules under `override_dh_install:` were added to move the files to the right place under the right names.

I'm not sure whether all dates and names in the `copyright` file are accurate, so place check that and let me know if something has to be fixed.

Since LKRG source is signed, the public key was also added so the `uscan` utility can verify automatically whether the source was properly signed. The current source of the `tar.gz` package is set to to https://www.openwall.com/lkrg/ , since you don't plan to provide releases on github.

I think that's all. If I missed/forgot something (or something is unclear), please ask.


debian/changelog
```diff
@@ -0,0 +1,5 @@
+lkrg (0.9.0-1) unstable; urgency=medium
+
+  * Initial Release (Closes: #123456)
```
solardiz, line 3: This would be `#944476`, not `#123456`.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=944476

debian/control
```diff
@@ -0,0 +1,53 @@
...
+Package: lkrg
+Architecture: all
+Depends: ${misc:Depends},
+ lkrg-dkms (= ${source:Version})
+Recommends: lkrg-systemd (= ${source:Version})
+Description: Linux Kernel Runtime Guard (LKRG)
+ Linux Kernel Runtime Integrity Checking and Exploit Detection.
+ .
+ LKRG provides security through diversity.
+ Similar to running an uncommon operating system (kernel) would.
```
solardiz, line 22: This is part of `CONCEPTS`, whereas our own brief description of LKRG is:

```
LKRG performs runtime integrity checking of the Linux kernel and detection of
security vulnerability exploits against the kernel.

LKRG is a kernel module (not a kernel patch), so it can be built for and loaded
on top of a wide range of mainline and distros' kernels, without needing to
patch those.
```

I think it's more appropriate to use this here (and in descriptions of the subpackages), and to package `CONCEPTS` as part of documentation.

debian/control
```diff
@@ -0,0 +1,53 @@
...
+ .
+ It's only a dependency package to install the LKRG kernel module and also some
```
solardiz, line 24: Replace "It's" with "This is", wrap the line at a length consistent with what you use in other descriptions here.

debian/copyright
```diff
@@ -0,0 +1,29 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: lkrg
+Upstream-Contact: Adam Zabrocki <pi3@itsec.pl>
+Source: https://www.openwall.com/lkrg/
+
+Files: *
+Copyright: 2018-2021 Adam Zabrocki <pi3@itsec.pl>
+License: GPL-2.0
```
solardiz, line 8: We don't use Adam's `itsec.pl` e-mail address anywhere in LKRG, so I think we shouldn't introduce it here. Perhaps use his address from git commits and `.mailmap`, which is `Adam 'pi3' Zabrocki <pi3@pi3.com.pl>`.

Also, we have other people's contributions that are non-trivial enough that they're subject to copyright. Our `LICENSE` says:

```
Copyright (c) 2015-2020 Adam 'pi3' Zabrocki
Copyright (c) 2020 Mariusz Zaborski
Copyright (c) 2020 Solar Designer
```

(BTW, we forgot to update the copyright years to reflect.)

A few source files mention:

```
 * Author:
 *  - Mariusz Zaborski (https://oshogbo.vexillium.org/)
```

Two contributors provided `Signed-off-by` tags, which I think imply copyright claim:

```
$ git log | fgrep Signed-off-by | sort -u
```

although as I recall Vitaly's non-trivial enough contributions are only for mkosi/GA, not LKRG core code.

In `debian/copyright`, can you just refer to the copyright holder list in `LICENSE` or do you have to duplicate it?

debian/lkrg-dkms.sysctl
```diff
@@ -0,0 +1,21 @@
...
+#lkrg.smep_validate = 1
+#lkrg.smep_enforce = 2
+#lkrg.smap_validate = 0
+#lkrg.smap_enforce = 0
+#lkrg.umh_validate = 1
+#lkrg.umh_enforce = 1
+#lkrg.pcfi_validate = 2
+#lkrg.pcfi_enforce = 1
+#lkrg.profile_validate = 3
+#lkrg.profile_enforce = 2
```
solardiz, line 21: I suggest you move the two `profile` lines to the top, so that someone can easily use a profile and then override some of its settings.

debian/upstream/metadata
```diff
@@ -0,0 +1,12 @@
+---
+Archive: github
+Name: lkrg
+Contact: Adam Zabrocki <pi3@itsec.pl>
+Security-Contact: Adam Zabrocki <pi3@itsec.pl>
```
solardiz, line 5: Maybe point `Contact` at lkrg-users mailing list (archive, where there's also subscription button) and `Security-Contact` at Adam's other address?

solardiz (commented):
I'm not fully qualified to review this as I never packaged anything for Debian, but I made some comments. I'd also appreciate reviews by @oshogbo and @adrelanos.

debian/changelog
morfikov: I forgot about this. Yes, this should be changed to `#944476`.

debian/control
morfikov: Can be done.

debian/control
morfikov: Ok.

debian/copyright
morfikov: To be honest, I don't know. I usually put here just the name of author/authors. If there are also contributors which provide some additional files, then those people can be listed along with the files. 

I think contributors should be listed in the project itself as they don't really own rights to the files, and the `copyright` file is to just about the copyrights.  

Basically every file should have an owner, So that's why you see something like:

```
Files: *
Copyright: 2018-2021 Adam Zabrocki <pi3@itsec.pl>
License: GPL-2.0

Files: debian/*
Copyright: 2020-2021 Mikhail Morfikov <mmorfikov@gmail.com>
License: GPL-2.0 
```
And you can add as many as you need. Of course many authors can be listed in each of such section.

debian/lkrg-dkms.sysctl
morfikov: It's just an example in which I wanted to list the possible config options.

debian/upstream/metadata
morfikov: [Here's the info on this file.](https://wiki.debian.org/UpstreamMetadata#Fields) if you needed any additional fields.

I can make the changes you ask.

debian/copyright
morfikov: > Copyright (c) 2015-2020 Adam 'pi3' Zabrocki
> Copyright (c) 2020 Mariusz Zaborski
> Copyright (c) 2020 Solar Designer

I think emails should also be listed along with the names.

morfikov:
Should be good now. I added also the `debian/lkrg-dkms.docs` to list the doc files.

debian/control
```diff
@@ -0,0 +1,59 @@
...
+Description: Linux Kernel Runtime Guard (LKRG) Source Code and DKMS
+ LKRG performs runtime integrity checking of the Linux kernel and detection of
+ security vulnerability exploits against the kernel.
+ .
+ LKRG is a kernel module (not a kernel patch), so it can be built for and loaded
+ on top of a wide range of mainline and distros' kernels, without needing to
+ patch those.
+ .
+ This package uses DKMS to automatically build the LKRG kernel
+ module.
```
solardiz, line 44: You can unwrap this line.

debian/lkrg-dkms.sysctl
```diff
@@ -0,0 +1,23 @@
...
+#lkrg.trigger = 0
+#lkrg.smep_validate = 1
+#lkrg.smep_enforce = 2
+#lkrg.smap_validate = 0
+#lkrg.smap_enforce = 0
+#lkrg.umh_validate = 1
+#lkrg.umh_enforce = 1
+#lkrg.pcfi_validate = 2
+#lkrg.pcfi_enforce = 1
+
```
solardiz, line 23: Trailing empty line here.

debian/patches/series
```diff
@@ -0,0 +1 @@
+
```
solardiz, line 1: Is the `series` files required when there are no patches? Why put empty line into it rather than keep it zero-sized?

debian/watch
```diff
@@ -0,0 +1,4 @@
+version=4
+https://www.openwall.com/lkrg/lkrg-(.+).tar.gz
+opts=pgpsigurlmangle=s/$/.sign/ https://www.openwall.com/lkrg/lkrg-(.+).tar.gz
+
```
solardiz, line 4: Trailing empty line.

debian/copyright
solardiz: E-mail addresses must be optional: someone might be a copyright holder, but not even have an e-mail address.

solardiz (approved):
I made some nitpick comments, but overall I'm fine getting this merged.

solardiz:
Oh, you could also update `CHANGES` noting the addition of the `debian/` subdirectory. If you do, please start it with:

```
The following major changes have been made between LKRG 0.9.0 and current:
```

and separate the added list of changes (with just one change by you so far) from the current content of the file with two empty lines.

debian/control
morfikov: I'll fix that.

debian/lkrg-dkms.sysctl
morfikov: I'll fix that.

debian/patches/series
morfikov: Basically the whole `debian/patches/` dir isn't required. I can remove it since there's no patches so far.

debian/watch
morfikov: I'll fix that.

morfikov:
> Oh, you could also update `CHANGES`

I think there's no need for that as this generally describes the project changes. The `debian/` dir just make it easier to build the source on Debian based distros.

solardiz:
`CHANGES` for 0.9.0 includes several "similar" changes:

```
 *) Integrate LKRG with out-of-tree (a tool to assist kernel module testing)
 *) Integrate LKRG with mkosi (systemd's tool for generating a test boot image)
 *) Continuous Integration setup: boot tests on GitHub Actions using mkosi
    (with Ubuntu's release kernels and their daily builds of mainline kernels)
```

and we need to be mentioning things like that in release announcements. So please add yours to `CHANGES`, too. Thanks!

debian/copyright
morfikov: Yes, that's true.

morfikov:
> `CHANGES` for 0.9.0 includes several "similar" changes:
> 
> ```
>  *) Integrate LKRG with out-of-tree (a tool to assist kernel module testing)
>  *) Integrate LKRG with mkosi (systemd's tool for generating a test boot image)
>  *) Continuous Integration setup: boot tests on GitHub Actions using mkosi
>     (with Ubuntu's release kernels and their daily builds of mainline kernels)
> ```
> 
> and we need to be mentioning things like that in release announcements. So please add yours to `CHANGES`, too. Thanks!

Ok, I think it's good now.

Any other suggestions?

solardiz:
If you do end up making other changes, please also fix this:

> separate the added list of changes (with just one change by you so far) from the current content of the file with two empty lines.

Other than that, if no one else requests changes now, I think we can merge. Thanks!

morfikov:
Ok, I now it should be good. :)

solardiz:
Thank you, @morfikov! If others who are into Debian packaging suggest changes, we'll take care of those separately.

adrelanos:
Mind porting features from https://github.com/Whonix/lkrg/tree/master/debian to here?

morfikov:
> Mind porting features from https://github.com/Whonix/lkrg/tree/master/debian to here?

What features?

adrelanos:
May I kindly suggest looking through these files one by one and compare?

(Because if I search/describe the features, I guess I spend almost the same time as creating (a) pull request(s). Just wondering if you're up for it since you wrote "Any other suggestions?")

solardiz:
@adrelanos I took a brief look at https://github.com/Whonix/lkrg/tree/master/debian earlier today and didn't see anything we'd obviously want to take from there. But like I said, I'm not into Debian packaging. So I think ideally you'd suggest what good there is for us to take. Just having @morfikov pick stuff from there doesn't feel like an obvious improvement to me. Thanks!

solardiz:
OK, here's some of what I see in Whonix packaging:

`make -j1` in DKMS when building in a low-RAM VM. Well, maybe, although this isn't really LKRG specific.

Workaround for VirtualBox/KVM. This one shouldn't currently be needed, because we've since set the default for `msr_validate` to 0.

Some workaround for `tirdad` - I don't know whether this is still needed (was this ever reported to us?)

```
## Load tirdad before LKRG so LKRG does not judge tirdad to be malicious.
softdep p_lkrg pre: tirdad
```

`lkrg-dkms.install` somehow picking individual files to copy to `/usr/src`.

Do we need any of this?

morfikov:
Before I created my own `debian/` dir, I tried to use the Whonix one. It was a long time ago. Basically I had to cut many things out of their `debian/` dir to be able to use `pbuilder` in the way I usually do. So the `debian/` dir that I "pasted` here is just all that user needs to build the LKRG module in the Debian way using its tools. So that's why I asked what features Whonix has which LKRG already doesn't have?

adrelanos:
`debian/control`

> Architecture: linux-amd64

Because lkrg won't run on i386, arm, power9, etc. also it won't work with other kernels such as gnu-hurd.

----

https://github.com/Whonix/lkrg/blob/master/debian/lkrg-dkms.dkms#L19 has a feature to reduce number of cores during compilation to 1 if less than 1950 MB RAM, because otherwise compilation would freeze in VirtualBox (maybe in other situations too) due to a kernel bug under RAM pressure.

----

That's as far as I got for now. Will write more soonish.

solardiz:
> lkrg won't run on i386, arm

Actually, it is supposed to now, and does in Adam's testing. So we need to list the 4 supported archs there.

> due to a kernel bug under RAM pressure.

Not really a kernel bug, just not enough RAM for the requested parallelism when you e.g. proxy your host's many vCPUs into a VM, but only give it 1 GB RAM.

morfikov:
> lkrg-dkms.install somehow picking individual files to copy to /usr/src.

This isn't needed as the config I provided is automated in terms of building the kernel module. Whonix has issues with this, and [they even stated this in the dkms file](https://github.com/Whonix/lkrg/blob/master/debian/lkrg-dkms.dkms). That's why I had to make some changes to make this a fully automated thing.

adrelanos:
> OK, here's some of what I see in Whonix packaging:
> 
> `make -j1` in DKMS when building in a low-RAM VM. Well, maybe, although this isn't really LKRG specific.

LKRG isn't the root cause indeed. LKRG packaging is just a user of DKMS. But I wouldn't know how else to work around this.

> Not really a kernel bug, just not enough RAM for the requested parallelism when you e.g. proxy your host's many vCPUs into a VM, but only give it 1 GB RAM.

1 GB (or even less) RAM usually works fine in (VirtualBox) VMs. That is a common and "acceptable" configuration, I suppose (think servers). The only non-functional thing was lkrg compilation. It's a kernel bug because even a big swap file doesn't help to not freeze the system. The problem is that DKMS uses 4 cores by default. It's not as if the DKMS default was 1, users set it to 4 with only 1 GB RAM and ten wonder why it's freezing during compilation.

> Some workaround for `tirdad` - I don't know whether this is still needed (was this ever reported to us?)

Never reported. Since http://github.com/Whonix/tirdad modifies the kernel indeed, I am not surprised lkrg detects that.

> ```
> ## Load tirdad before LKRG so LKRG does not judge tirdad to be malicious.
> softdep p_lkrg pre: tirdad
> ```

I guess I can move that to tirdad to keep it out of LKRG.

> lkrg-dkms.install somehow picking individual files to copy to /usr/src.

> Do we need any of this?

It's either in debian/rules or in debian/pkg-name.install. Since the current implementation doesn't hardcode version numbers, the existing implementation (already in this repository) is better.

> Workaround for VirtualBox/KVM. This one shouldn't currently be needed, because we've since set the default for msr_validate to 0.

I guess you mean https://github.com/Whonix/lkrg/blob/master/debian/lkrg-loader isn't required?
Well, I suppose you've looked at the files and followed these links. If none of these module parameters is required (anymore) (since default nowadays?) then I am happy if a virtualizer (host and VM) works out of the box without requirement for any loader detecting best module options.

solardiz:
> I guess you mean https://github.com/Whonix/lkrg/blob/master/debian/lkrg-loader isn't required?

That's what I meant, but upon a closer look I see/recall that besides `msr_validate` for VirtualBox hosts also a change of `pcfi_validate` was needed, and we didn't change the default for this one, instead documenting the problem:

```
  lkrg.pcfi_validate=2 is incompatible with VirtualBox hosts, where you need to
  use at most lkrg.pcfi_validate=1.  However, there's no problem with setting
  lkrg.pcfi_validate=2 on Linux+LKRG guest systems in VirtualBox VMs.
```

I think ideally we'd somehow work around this in LKRG itself, but meanwhile it's either documentation above or a script like Whonix `lkrg-loader`.

adrelanos:
> I think ideally we'd somehow work around this in LKRG itself, but meanwhile it's either documentation above or a script like Whonix `lkrg-loader`.

Is there a (strong) technical reason to have this inside LKRG itself and not in a loader script?

I'd suggest to use scripting languages (easier) whenever reasonably possible.

The loader is already well tested and working great. Easy to understand/modify in future if required.

solardiz:
> Is there a (strong) technical reason to have this inside LKRG itself and not in a loader script?

Mostly user experience. We cannot expect every LKRG user to only load LKRG via a provided script. So we have to also document `pcfi_validate` vs. VirtualBox like in the quote I included above, and even then there will be users not reading documentation and running into this problem.

We had a "similar" (for user experience) issue with overlayfs[2] loading time, which we worked around in LKRG itself - and I think it's great that we did - much better than having to document it.

solardiz:
> Is there a (strong) technical reason to have this inside LKRG itself and not in a loader script?

Another reason is what condition we check. Right now, Whonix' loader script checks whether VirtualBox is installed, not whether it's in use. This makes sense if the check is only done proactively and only once. LKRG itself could instead check for VirtualBox host's module being inserted into the kernel, so it'd only weaken LKRG protection if and when this happens. Optionally, it could also revert the weakening when the module is removed from the kernel.

> The problem is that DKMS uses 4 cores by default.

Oh, my guess was it was based on actual vCPU count, no? If it's indeed a hard-coded default, then we could do better by determining and using max(min(vCPUs, GBs), 1) on our own.

PR #76, opened 2021-04-20
Enhance systemd service, introduce lkrg.conf

wladmis:
lkrg-helper is aimed to help service manager to start and stop lkrg
service. Its main purpose to store and restore lkrg settings state
during service restart. It is designed to be service manager agnostic.

This commit also introduces a new semantic for systemctl reload that
restores lkrg default settings.

* libexec/lkrg-helper: New file.
* systemd/lkrg-systemd.sh: Install lkrg-helper.
* systemd/lkrg.service: Use lkrg-helper; add new semantic for reload
  that restores default lkrg settings.

### DIFF
--- a/scripts/bootup/lkrg.conf
+++ b/scripts/bootup/lkrg.conf
@@ -1,0 +1,18 @@
+# lkrg.conf
+# Set LKRG sysctl settings here, so they won't be lost on service restart
+
+# Uncomment to block loading of kernel modules.
+#
+# This feature is meant primarily to prevent unintended user-triggered (or
+# attacker-triggered) auto-loading of maybe-vulnerable modules provided in a
+# distribution after all intended modules have already been loaded.  This
+# feature is not effective (nor is meant to be) against attackers who already
+# have root privileges and try to load a module explicitly (they could simply
+# flip this setting or even unload LKRG first).
+#
+# Please note that enabling this setting (too) early (e.g., using the module
+# parameter or /etc/sysctl.*) may cause the system to fail to complete bootup
+# (if required modules are still being loaded in later stages of bootup, which
+# varies between distributions and system configurations).
+#
+# lkrg.block_modules = 1
--- a/scripts/bootup/systemd/lkrg-systemd.sh
+++ b/scripts/bootup/systemd/lkrg-systemd.sh
@@ -6,6 +6,7 @@
 #  - Adam 'pi3' Zabrocki (http://pi3.com.pl)
 ##
 
+P_SYSCTL_DIR="/etc/sysctl.d"
 P_SYSTEMD_DIR="$(systemctl show -p UnitPath | cut -d " " -f5)"
 
 case "$P_SYSTEMD_DIR" in
@@ -32,6 +33,12 @@ if [ "$1" == "install" ]; then
 	systemctl enable lkrg.service
 	echo -e "       ${P_GREEN}To start ${P_YL}lkrg.service${P_GREEN} please use: ${P_YL}systemctl start lkrg${P_NC}"
     fi
+    if [ ! -e "$P_SYSCTL_DIR/lkrg.conf" ]; then
+	echo -e "       ${P_GREEN}Installing ${P_YL}lkrg.conf${P_GREEN} file under ${P_YL}$P_SYSCTL_DIR${P_GREEN} folder${P_NC}"
+	install -pm 644 -o root -g root scripts/bootup/lkrg.conf "$P_SYSCTL_DIR/lkrg.conf"
+    else
+	echo -e "       ${P_YL}lkrg.conf${P_GREEN} is already installed, skipping${P_NC}"
+    fi
 elif [ "$1" == "uninstall" ]; then
     echo -e "       ${P_GREEN}Stopping ${P_YL}lkrg.service${P_NC}"
     systemctl stop lkrg.service
@@ -39,6 +46,14 @@ elif [ "$1" == "uninstall" ]; then
     systemctl disable lkrg.service
     echo -e "       ${P_GREEN}Deleting ${P_YL}lkrg.service${P_GREEN} file from the ${P_YL}$P_SYSTEMD_DIR${P_GREEN} folder${P_NC}"
     rm "$P_SYSTEMD_DIR/lkrg.service"
+    if cmp -s "$P_SYSCTL_DIR/lkrg.conf" scripts/bootup/lkrg.conf; then
+        echo -e "       ${P_GREEN}Deleting unmodified ${P_YL}lkrg.conf${P_GREEN} file from the ${P_YL}$P_SYSCTL_DIR${P_GREEN} folder${P_NC}"
+        rm "$P_SYSCTL_DIR/lkrg.conf"
+    else
+        echo -e "       ${P_YL}$P_SYSCTL_DIR/lkrg.conf${P_GREEN} was modified, preserve it as ${P_YL}$P_SYSCTL_DIR/lkrg.conf.saved${P_NC}"
+        echo -e "       ${P_GREEN}If you do not need it anymore, delete it manually${P_NC}"
+        mv "$P_SYSCTL_DIR/lkrg.conf"{,.saved}
+    fi
 else
     echo -e "      ${P_RED}ERROR! Unknown option!${P_NC}"
     exit 1
--- a/scripts/bootup/systemd/lkrg.service
+++ b/scripts/bootup/systemd/lkrg.service
@@ -17,6 +17,7 @@ ConditionKernelCommandLine=!nolkrg
 [Service]
 Type=oneshot
 ExecStart=/sbin/modprobe -v p_lkrg
+ExecStartPost=/sbin/sysctl -p /etc/sysctl.d/lkrg.conf
 ExecStop=/sbin/modprobe -v -r p_lkrg
 RemainAfterExit=yes
 


wladmis:
Many distros have a policy to restart services when the packages containing these services are updated. This leads that all the LKRG sysctl settings are reset to defaults. This PR is aimed to solve this problem.

wladmis:
@vt-alt

solardiz:
Thanks @wladmis!

Aside from the main question on whether this functionality is desirable at all (it may well be - you give the rationale), I this possibly has the issue that the settings are not entirely independent. Altering a setting that's part of a profile brings the corresponding profile setting to 9, but that 9 can't be written back into the profile sysctl (will lead to an error, which I guess this script will silently ignore?) Luckily, the underlying settings that had brought the profile to 9 should also be in there, and should have the same effect again. I don't recall us having any more obscure cases where the order of writes would matter.

scripts/bootup/libexec/lkrg-helper
```diff
@@ -0,0 +1,77 @@
...
+}
+
+check()
+{
+	grep -qw p_lkrg /proc/modules
+}
+
+start()
+{
+	[ ! -d "$rundir" ] ||
```
vt-alt, line 31: Why the check if `mkdir -p` will already do almost the same.

scripts/bootup/libexec/lkrg-helper
```diff
@@ -0,0 +1,77 @@
...
+			cat "$sysctl/$state" >"$statedir/$state"
+		done
+	fi
+
+	/sbin/modprobe -v -r p_lkrg
+}
+
+reset()
+{
+	stop reset
```
vt-alt, line 66: This is obscure, what `reset` is supposed to do? If it's called from `ExecReload=` it should `trigger a configuration reload in the service` not just stopping the service.

vt-alt:
> Aside from the main question on whether this functionality is desirable at all 

I agree. Users may expect "restart" to re-reading the config (as `reload`, but with colder service restart) not to retain run-time configuration.

wladmis:
On Tue, Apr 20, 2021 at 03:01:25PM -0700, Vitalio wrote:
> @vt-alt commented on this pull request.
> 
> 
> 
> > +	check || warn_on_exit "Does not run"
> +
> +
> +	if [ "${1-}" != "reset" ] && [ -d "$sysctl" ]; then
> +		mkdir -p "$statedir"
> +		for state in $(cd "$sysctl" && set +f && echo *); do
> +			cat "$sysctl/$state" >"$statedir/$state"
> +		done
> +	fi
> +
> +	/sbin/modprobe -v -r p_lkrg
> +}
> +
> +reset()
> +{
> +	stop reset
> 
> This is obscure, what `reset` is supposed to do? If it's called from
> `ExecReload=` it should `trigger a configuration reload in the
> service` not just stopping the service.

Whoops, that's a bug, fixed!

wladmis:
On Tue, Apr 20, 2021 at 03:04:05PM -0700, Vitalio wrote:
> > Aside from the main question on whether this functionality is desirable at all 
> 
> I agree. Users may expect "restart" to re-reading the config (as
> `reload`, but with colder service restart) not to retain run-time
> configuration.

It's not about the user, it's about of package update. When you update
LKRG (or kernel that also pulls the lkrg module), it forces to restart
LKRG service. Now it resets settings, I assume it as a bug.

To reset settings to default I propose to use systemctl reload lkrg.

vt-alt:
> It's not about the users, it's about of package updates.

I didn't notice where it distinguishes manual restart and restart from package upgrade then.

wladmis:
On Tue, Apr 20, 2021 at 03:23:03PM -0700, Vitalio wrote:
> > It's not about the users, it's about of package updates.
> 
> I didn't notice where it distinguishes manual restart and restart from package upgrade then.

It does not. It's not an issue of this solution, it is an issue of
service and packages management.

vt-alt:
> On Tue, Apr 20, 2021 at 03:23:03PM -0700, Vitalio wrote: > It's not about the users, it's about of package updates. I didn't notice where it distinguishes manual restart and restart from package upgrade then.
> It does not. It's not an issue of this solution, it is an issue of service and packages management.

Do you mean this solution does not affect both service and packages management? What will happen if the user restarts `lkrg.service` after this solution applied (and expects configuration re-load)?

wladmis:
On Tue, Apr 20, 2021 at 03:39:28PM -0700, Vitalio wrote:
> > On Tue, Apr 20, 2021 at 03:23:03PM -0700, Vitalio wrote: > It's not about the users, it's about of package updates. I didn't notice where it distinguishes manual restart and restart from package upgrade then.
> > It does not. It's not an issue of this solution, it is an issue of service and packages management.
> 
> Do you mean this solution does not affect both service and packages
> management?

Opposite: it does affect.

> What will happen if the user restarts `lkrg.service` after this
> solution applied (and expects configuration re-load)?

I think he'll noticed that it does not do what he expected and he should
try something else.

morfikov:
I have a question. Since you edit the `lkrg.service` file anyway, wouldn't be better to add the following to the service file?

`ExecStartPost=/usr/sbin/sysctl -p`

The command is run after the command in `ExecStart=` is executed. In this way, no extra work/scripts is needed, which simplifies the whole process, and you can relay on the sysetmd service alone in any user case, so not only when the package is upgraded but also when a user decides to restart the service in any given time.

wladmis:
On Tue, Apr 20, 2021 at 11:54:11PM -0700, Mikhail Morfikov wrote:
> I have a question. Since you edit the `lkrg.service` file anyway, wouldn't be better to add the following to the service file?
> 
> `ExecStartPost=/usr/sbin/sysctl -p`
> 
> The command is run after the command in `ExecStart=` is executed. In
> this way, no extra work/scripts is needed, which simplifies the whole
> process, and you can relay on the sysetmd service alone in any user
> case, so not only when the package is upgraded but also when a user
> decides to restart the service in any given time.

Hm, this approach causes the lost of changes were made manually, but I
think it's OK. Your solution is much simpler and better. Thank you!

solardiz:
I am working on something else now, so am not really here to comment for real, but to prevent things from going in a wrong direction:

> `ExecStartPost=/usr/sbin/sysctl -p`

I think this is unacceptable in that it'd reset _all_ sysctl's specified in `/etc/sysctl.*`, not only those related to LKRG. This isn't something a package of LKRG is expected to do.

wladmis:
> I am working on something else now, so am not really here to comment for real, but to prevent things from going in a wrong direction:
> 
> > `ExecStartPost=/usr/sbin/sysctl -p`
> 
> I think this is unacceptable in that it'd reset _all_ sysctl's specified in `/etc/sysctl.*`, not only those related to LKRG. This isn't something a package of LKRG is expected to do.

On the second thought, I agree that all changes was made manually will be lost. We cannot use sysctl -p.

wladmis:
Alternatively, LKRG could provide, for example, /etc/sysctl.d/lkrg.conf to store its settings, and use `sysctl -p /etc/sysctl.d/lkrg.conf` in the post-start. In this case strong usage of this file should be documented.

@solardiz what do you think?

morfikov:
> I am working on something else now, so am not really here to comment for real, but to prevent things from going in a wrong direction:
> 
> > `ExecStartPost=/usr/sbin/sysctl -p`
> 
> I think this is unacceptable in that it'd reset _all_ sysctl's specified in `/etc/sysctl.*`, not only those related to LKRG. This isn't something a package of LKRG is expected to do.

I know, but it was just a suggestion to use `sysctl -p` in the service file. The config file can also be passed to load only the lkrg settings (I didn't even know it can do such thing).

scripts/bootup/systemd/lkrg-systemd.sh
```diff
@@ -32,6 +33,18 @@ if [ "$1" == "install" ]; then
...
 	echo -e "       ${P_GREEN}To start ${P_YL}lkrg.service${P_GREEN} please use: ${P_YL}systemctl start lkrg${P_NC}"
     fi
+    if [ ! -e "$P_SYSCTL_DIR/lkrg.conf" ]; then
+	echo -e "       ${P_GREEN}Installing ${P_YL}lkrg.conf${P_GREEN} file under ${P_YL}$P_SYSCTL_DIR${P_GREEN} folder${P_NC}"
+	cat <<EOF >"$P_SYSCTL_DIR/lkrg.conf"
+# lkrg.conf
+# Use this file to store lkrg settings, otherwise
+# it will be lost during service restart
+
+# lkrg.block_modules = 1
```
Adam-pi3, line 43: Why do we have this here?

scripts/bootup/systemd/lkrg.service
```diff
@@ -17,7 +17,9 @@ ConditionKernelCommandLine=!nolkrg
 [Service]
 Type=oneshot
 ExecStart=/sbin/modprobe -v p_lkrg
+ExecStartPost=/sbin/sysctl -p /etc/sysctl.d/lkrg.conf
 ExecStop=/sbin/modprobe -v -r p_lkrg
+ExecReload=/sbin/sysctl -p /etc/sysctl.d/lkrg.conf
```
Adam-pi3, line 22: I don't see `lkrg.conf` in this commit. Is it pushed or smth is wrong on my side? ;-)

Adam-pi3 (changes requested):

wladmis:
On Wed, Apr 21, 2021 at 08:11:20AM -0700, Adam-pi3 wrote:
> @Adam-pi3 requested changes on this pull request.
> 
> 
> 
> > @@ -32,6 +33,18 @@ if [ "$1" == "install" ]; then
>  	systemctl enable lkrg.service
>  	echo -e "       ${P_GREEN}To start ${P_YL}lkrg.service${P_GREEN} please use: ${P_YL}systemctl start lkrg${P_NC}"
>      fi
> +    if [ ! -e "$P_SYSCTL_DIR/lkrg.conf" ]; then
> +	echo -e "       ${P_GREEN}Installing ${P_YL}lkrg.conf${P_GREEN} file under ${P_YL}$P_SYSCTL_DIR${P_GREEN} folder${P_NC}"
> +	cat <<EOF >"$P_SYSCTL_DIR/lkrg.conf"
> +# lkrg.conf
> +# Use this file to store lkrg settings, otherwise
> +# it will be lost during service restart
> +
> +# lkrg.block_modules = 1
> 
> Why do we have this here?

It's just a commented example.

> >  ExecStop=/sbin/modprobe -v -r p_lkrg
> +ExecReload=/sbin/sysctl -p /etc/sysctl.d/lkrg.conf
> 
> I don't see `lkrg.conf` in this commit. Is it pushed or smth is wrong on my side? ;-)

It is generated and installed in the lkrg-systemd.sh.

Adam-pi3:
Ah I see. I did overlooked that , thanks. Does it mean that on `uninstall` you need to delete this file?

wladmis:
On Wed, Apr 21, 2021 at 08:36:31AM -0700, Adam-pi3 wrote:
> Ah I see. I did overlooked that , thanks. Does it mean that on
> `uninstall` you need to delete this file?

Usually config files are preserved after software uninstallation.

morfikov:
> Usually config files are preserved after software uninstallation.

It depends, usually there are two cases (for instance in debian's apt) -- you can `remove` without removing conffiles and `purge` which will also remove conffiles.

morfikov:
> Ah I see. I did overlooked that , thanks. Does it mean that on uninstall you need to delete this file?

Wouldn't be better to provide the necessary conffiles as a part of the project?

scripts/bootup/systemd/lkrg-systemd.sh
```diff
@@ -32,6 +33,28 @@ if [ "$1" == "install" ]; then
 	systemctl enable lkrg.service
 	echo -e "       ${P_GREEN}To start ${P_YL}lkrg.service${P_GREEN} please use: ${P_YL}systemctl start lkrg${P_NC}"
     fi
+    if [ ! -e "$P_SYSCTL_DIR/lkrg.conf" ]; then
+	echo -e "       ${P_GREEN}Installing ${P_YL}lkrg.conf${P_GREEN} file under ${P_YL}$P_SYSCTL_DIR${P_GREEN} folder${P_NC}"
+	cat <<EOF >"$P_SYSCTL_DIR/lkrg.conf"
+# lkrg.conf
+# Use this file to store lkrg settings, otherwise
+# it will be lost during service restart
```
solardiz, line 41: Maybe:

```
# Set LKRG sysctl settings here, so they won't be lost on service restart
```

scripts/bootup/systemd/lkrg-systemd.sh
```diff
@@ -32,6 +33,28 @@ if [ "$1" == "install" ]; then
 	systemctl enable lkrg.service
 	echo -e "       ${P_GREEN}To start ${P_YL}lkrg.service${P_GREEN} please use: ${P_YL}systemctl start lkrg${P_NC}"
     fi
+    if [ ! -e "$P_SYSCTL_DIR/lkrg.conf" ]; then
+	echo -e "       ${P_GREEN}Installing ${P_YL}lkrg.conf${P_GREEN} file under ${P_YL}$P_SYSCTL_DIR${P_GREEN} folder${P_NC}"
+	cat <<EOF >"$P_SYSCTL_DIR/lkrg.conf"
```
solardiz, line 38: With this, permissions on `lkrg.conf` will depend on umask as of `make install`. This is probably undesirable. We should probably explicitly set `umask 077` or `umask 022` before this `cat`.

scripts/bootup/systemd/lkrg-systemd.sh
```diff
@@ -32,6 +33,28 @@ if [ "$1" == "install" ]; then
...
+# lkrg.block_modules = 1
+
+# Unfomment to not validate CPU Model Specific Registers.
+# There are situations where such validation is undesirable, such as if you run
+# LKRG on a host machine that manages VMs and dynamically reconfigures MSRs.
+# This is known to be the case for KVM and VirtualBox hosts, where this setting
+# needs to be disabled.  However, there's no problem with enabling this setting
+# on Linux+LKRG guest systems in VMs on those hosts, and indeed on systems that
+# don't run KVM and VirtualBox.
+# lkrg.msr_validate = 0
```
solardiz, line 53: The current default for `lkrg.msr_validate` is 0, so it is wrong to ask to "Unfomment" (BTW, a typo) that line as if that would make a difference. We could ask to "Uncomment" the opposite kind of line to enable the validation.

The one incompatibility that LKRG has with VirtualBox on host by default is with `lkrg.pcfi_validate`, where it needs to be lowered from the default of 2 to 1. Maybe this needs a commented out line in here.

scripts/bootup/systemd/lkrg-systemd.sh
```diff
@@ -32,6 +33,28 @@ if [ "$1" == "install" ]; then
...
     fi
+    if [ ! -e "$P_SYSCTL_DIR/lkrg.conf" ]; then
+	echo -e "       ${P_GREEN}Installing ${P_YL}lkrg.conf${P_GREEN} file under ${P_YL}$P_SYSCTL_DIR${P_GREEN} folder${P_NC}"
+	cat <<EOF >"$P_SYSCTL_DIR/lkrg.conf"
+# lkrg.conf
+# Use this file to store lkrg settings, otherwise
+# it will be lost during service restart
+
+# Uncomment to block loading of kernel modules.
+# lkrg.block_modules = 1
```
solardiz, line 44: I think we'd need to replicate the rationale and the warning from `README` to here, so that people hopefully won't enable `lkrg.block_modules` without having read those first.

scripts/bootup/systemd/lkrg.service
```diff
@@ -17,6 +17,7 @@ ConditionKernelCommandLine=!nolkrg
 [Service]
 Type=oneshot
 ExecStart=/sbin/modprobe -v p_lkrg
+ExecStartPost=/sbin/sysctl -p /etc/sysctl.d/lkrg.conf
```
solardiz, line 20: Those sysctl settings will end up being applied twice on system bootup. Not pretty.

solardiz (changes requested):
I suggest/request some changes, but since this is an RFC I think we also need opinions on whether this should get in or not.

solardiz:
If we do add a default/suggested `lkrg.conf`, perhaps we should allow for an easy way to take and install it on systems that don't use systemd or where the sysadmin chooses not to install our systemd unit. I don't mind initially providing it as part of the systemd-specific install script, but if so we should plan on moving it out of there later.

solardiz:
> Usually config files are preserved after software uninstallation.

My expectation is that a config file is only preserved on uninstall if the file was modified locally, not remained unchanged since install, and a warning is printed by the package manager about that leftover file. Leaving `/etc/sysctl.d/lkrg.conf` after LKRG is uninstalled would lead to (harmless) errors printed on bootup.

wladmis:
On Sat, Apr 24, 2021 at 04:42:42AM -0700, Solar Designer wrote:
> @solardiz requested changes on this pull request.
> 
> I suggest/request some changes, but since this is an RFC I think we also need opinions on whether this should get in or not.
> 
> > @@ -32,6 +33,28 @@ if [ "$1" == "install" ]; then
>  	systemctl enable lkrg.service
>  	echo -e "       ${P_GREEN}To start ${P_YL}lkrg.service${P_GREEN} please use: ${P_YL}systemctl start lkrg${P_NC}"
>      fi
> +    if [ ! -e "$P_SYSCTL_DIR/lkrg.conf" ]; then
> +	echo -e "       ${P_GREEN}Installing ${P_YL}lkrg.conf${P_GREEN} file under ${P_YL}$P_SYSCTL_DIR${P_GREEN} folder${P_NC}"
> +	cat <<EOF >"$P_SYSCTL_DIR/lkrg.conf"
> +# lkrg.conf
> +# Use this file to store lkrg settings, otherwise
> +# it will be lost during service restart
> 
> Maybe:
> 
> ```
> # Set LKRG sysctl settings here, so they won't be lost on service restart

OK.

> > @@ -32,6 +33,28 @@ if [ "$1" == "install" ]; then
>  	systemctl enable lkrg.service
>  	echo -e "       ${P_GREEN}To start ${P_YL}lkrg.service${P_GREEN} please use: ${P_YL}systemctl start lkrg${P_NC}"
>      fi
> +    if [ ! -e "$P_SYSCTL_DIR/lkrg.conf" ]; then
> +	echo -e "       ${P_GREEN}Installing ${P_YL}lkrg.conf${P_GREEN} file under ${P_YL}$P_SYSCTL_DIR${P_GREEN} folder${P_NC}"
> +	cat <<EOF >"$P_SYSCTL_DIR/lkrg.conf"
> 
> With this, permissions on `lkrg.conf` will depend on umask as of `make install`. This is probably undesirable. We should probably explicitly set `umask 077` or `umask 022` before this `cat`.

OK, I'll do it in the way to not mess the system umask value.

> > +	cat <<EOF >"$P_SYSCTL_DIR/lkrg.conf"
> +# lkrg.conf
> +# Use this file to store lkrg settings, otherwise
> +# it will be lost during service restart
> +
> +# Uncomment to block loading of kernel modules.
> +# lkrg.block_modules = 1
> +
> +# Unfomment to not validate CPU Model Specific Registers.
> +# There are situations where such validation is undesirable, such as if you run
> +# LKRG on a host machine that manages VMs and dynamically reconfigures MSRs.
> +# This is known to be the case for KVM and VirtualBox hosts, where this setting
> +# needs to be disabled.  However, there's no problem with enabling this setting
> +# on Linux+LKRG guest systems in VMs on those hosts, and indeed on systems that
> +# don't run KVM and VirtualBox.
> +# lkrg.msr_validate = 0
> 
> The current default for `lkrg.msr_validate` is 0, so it is wrong to ask to "Unfomment" (BTW, a typo) that line as if that would make a difference. We could ask to "Uncomment" the opposite kind of line to enable the validation.
> 
> The one incompatibility that LKRG has with VirtualBox on host by default is with `lkrg.pcfi_validate`, where it needs to be lowered from the default of 2 to 1. Maybe this needs a commented out line in here.
> 
> > @@ -32,6 +33,28 @@ if [ "$1" == "install" ]; then
>  	systemctl enable lkrg.service
>  	echo -e "       ${P_GREEN}To start ${P_YL}lkrg.service${P_GREEN} please use: ${P_YL}systemctl start lkrg${P_NC}"
>      fi
> +    if [ ! -e "$P_SYSCTL_DIR/lkrg.conf" ]; then
> +	echo -e "       ${P_GREEN}Installing ${P_YL}lkrg.conf${P_GREEN} file under ${P_YL}$P_SYSCTL_DIR${P_GREEN} folder${P_NC}"
> +	cat <<EOF >"$P_SYSCTL_DIR/lkrg.conf"
> +# lkrg.conf
> +# Use this file to store lkrg settings, otherwise
> +# it will be lost during service restart
> +
> +# Uncomment to block loading of kernel modules.
> +# lkrg.block_modules = 1
> 
> I think we'd need to replicate the rationale and the warning from
> `README` to here, so that people hopefully won't enable
> `lkrg.block_modules` without having read those first.

OK

> > @@ -17,6 +17,7 @@ ConditionKernelCommandLine=!nolkrg
>  [Service]
>  Type=oneshot
>  ExecStart=/sbin/modprobe -v p_lkrg
> +ExecStartPost=/sbin/sysctl -p /etc/sysctl.d/lkrg.conf
> 
> Those sysctl settings will end up being applied twice on system bootup. Not pretty.

Not pretty indeed but I don't see how could it be a problem.

morfikov:
> If we do add a default/suggested lkrg.conf, perhaps we should allow for an easy way to take and install it on systems that don't use systemd or where the sysadmin chooses not to install our systemd unit. I don't mind initially providing it as part of the systemd-specific install script, but if so we should plan on moving it out of there later.

I just wanted to note that at least 3 init systems (sysvinit, systemd, openrc) support loading and configuring modules via files in the `/etc/modprobe.d/`, `/etc/modules-load.d/` and  /etc/sysctl.d/ dirs. That's why it's better to provide such files and install them under their proper locations to support all the init systems.

wladmis:
On Sat, Apr 24, 2021 at 04:45:20AM -0700, Solar Designer wrote:
> If we do add a default/suggested `lkrg.conf`, perhaps we should allow
> for an easy way to take and install it on systems that don't use
> systemd or where the sysadmin chooses not to install our systemd unit.
> I don't mind initially providing it as part of the systemd-specific
> install script, but if so we should plan on moving it out of there
> later.

OK.

In fact, in the ALT systems if LKRG is installed, it is loading at early
systemd bootup process as part of initrd [1]. vt-alt@ insisted to make
services as well, the reason was to reload module during upgrade of
module package.

A bit of offtopic: since ALT supports both SysvInit and systemd, it has
also nice SysvInit script which I can contrib in the separate PR to some
contrib/ place as the reference example.

[1] http://git.altlinux.org/gears/m/make-initrd.git?p=make-initrd.git;a=tree;f=features/lkrg;h=795dcbdbe24bdbc7819ede54045ad649e58cb48b;hb=55ff0303f6f823d613557eb7099be032a97a91f0

wladmis:
On Sat, Apr 24, 2021 at 04:48:48AM -0700, Solar Designer wrote:
> > Usually config files are preserved after software uninstallation.
> 
> My expectation is that a config file is only preserved on uninstall if
> the file was modified locally, not remained unchanged since install,
> and a warning is printed by the package manager about that leftover
> file. Leaving `/etc/sysctl.d/lkrg.conf` after LKRG is uninstalled
> would lead to (harmless) errors printed on bootup.

Without context of the real package management it is tricky to
distinguish whether the file was modified or not. I think it would be
enough to print message to clean this file manually if user does not
need it anymore after package deinstallation.

scripts/bootup/systemd/lkrg-systemd.sh
```diff
@@ -32,6 +33,12 @@ if [ "$1" == "install" ]; then
 	systemctl enable lkrg.service
 	echo -e "       ${P_GREEN}To start ${P_YL}lkrg.service${P_GREEN} please use: ${P_YL}systemctl start lkrg${P_NC}"
     fi
+    if [ ! -e "$P_SYSCTL_DIR/lkrg.conf" ]; then
+	echo -e "       ${P_GREEN}Installing ${P_YL}lkrg.conf${P_GREEN} file under ${P_YL}$P_SYSCTL_DIR${P_GREEN} folder${P_NC}"
+	install -pD -m644 scripts/bootup/lkrg.conf "$P_SYSCTL_DIR/lkrg.conf"
```
solardiz, line 38: Let's use consistent options to `install`, which would mean drop `-D` and add `-o root -g root` here.

I understand `-D` may be convenient for a distro package installing LKRG into an otherwise-empty "buildroot", though. Do you suggest we use it here for that reason? If so, please add it to the previous `install` line as well.

scripts/bootup/lkrg.conf
```diff
@@ -0,0 +1,14 @@
...
+#
+# This feature is meant primarily to prevent unintended user-triggered (or
+# attacker-triggered) auto-loading of maybe-vulnerable modules provided in a
+# distribution after all intended modules have already been loaded.  This
+# feature is not effective (nor is meant to be) against attackers who already
+# have root privileges and try to load a module explicitly (they could simply
+# flip this setting or even unload LKRG first).
+#
+# lkrg.block_modules = 1
+
```
solardiz, line 14: Please add my recently added warning paragraph for `block_modules` from `README`, and drop the trailing empty line while you're at it.

solardiz (changes requested):

solardiz:
> Without context of the real package management it is tricky to distinguish whether the file was modified or not. I think it would be enough to print message to clean this file manually if user does not need it anymore after package deinstallation.

I think that even without package management, we can have `make uninstall` use `cmp` against its copy of the file, and remove the file if matches, or otherwise print a message as you suggest.

I understand that this may fail when someone uses `make uninstall` from a tree of a different (perhaps newer) version of LKRG than what's installed. However, such usage isn't guaranteed to work anyway (although it usually mostly does so far).

scripts/bootup/systemd/lkrg-systemd.sh
```diff
@@ -32,13 +33,28 @@ if [ "$1" == "install" ]; then
 	systemctl enable lkrg.service
 	echo -e "       ${P_GREEN}To start ${P_YL}lkrg.service${P_GREEN} please use: ${P_YL}systemctl start lkrg${P_NC}"
     fi
+    if [ ! -e "$P_SYSCTL_DIR/lkrg.conf" ]; then
+	echo -e "       ${P_GREEN}Installing ${P_YL}lkrg.conf${P_GREEN} file under ${P_YL}$P_SYSCTL_DIR${P_GREEN} folder${P_NC}"
+	mkdir -p "$P_SYSCTL_DIR"
+	install -pm 644 -o root -g root scripts/bootup/lkrg.conf "$P_SYSCTL_DIR/lkrg.conf"
```
solardiz, line 39: No, don't add `mkdir -p`. Like I wrote, let's decide on whether we use `-D` or not and then do it consistently for the two uses of `install`. That is, if you like `make install` to be creating system directories, then just add `-D` to both uses of `install`. Unless I am missing something?

scripts/bootup/systemd/lkrg-systemd.sh
```diff
@@ -32,13 +33,28 @@ if [ "$1" == "install" ]; then
...
+    fi
 elif [ "$1" == "uninstall" ]; then
     echo -e "       ${P_GREEN}Stopping ${P_YL}lkrg.service${P_NC}"
     systemctl stop lkrg.service
     echo -e "       ${P_GREEN}Disabling ${P_YL}lkrg.service${P_GREEN} on bootup${P_NC}"
     systemctl disable lkrg.service
     echo -e "       ${P_GREEN}Deleting ${P_YL}lkrg.service${P_GREEN} file from the ${P_YL}$P_SYSTEMD_DIR${P_GREEN} folder${P_NC}"
     rm "$P_SYSTEMD_DIR/lkrg.service"
+    if cmp -s "$P_SYSCTL_DIR/lkrg.conf" scripts/bootup/lkrg.conf; then
+        echo -e "       ${P_GREEN}Deleting ${P_YL}lkrg.conf${P_GREEN} file from the ${P_YL}$P_SYSCTL_DIR${P_GREEN} folder${P_NC}"
```
solardiz, line 51: I'd add the word "unmodified" after "Deleting".

scripts/bootup/systemd/lkrg-systemd.sh
```diff
@@ -32,13 +33,28 @@ if [ "$1" == "install" ]; then
...
+        rm "$P_SYSCTL_DIR/lkrg.conf"
+    else
+        echo -e "       ${P_YL}$P_SYSCTL_DIR/lkrg.service${P_GREEN} was modified, preserve it as ${P_YL}$P_SYSCTL_DIR/lkrg.conf.saved"
```
solardiz, line 54: Just "preserving it" and that's it - don't rename.

scripts/bootup/systemd/lkrg-systemd.sh
```diff
@@ -32,13 +33,28 @@ if [ "$1" == "install" ]; then
...
+        echo -e "       ${P_GREEN}If you do not need it anymore, delete it manually"
+        mv "$P_SYSCTL_DIR/lkrg.conf"{,.saved}
```
solardiz, line 56: Drop this `mv` line. This is confusing: the file is possibly still loaded, whereas the user would think that it is not. Then if LKRG is reinstalled, there will be two files in there.

solardiz (changes requested):

wladmis:
On Sun, Apr 25, 2021 at 02:27:47PM -0700, Solar Designer wrote:
> @solardiz requested changes on this pull request.
> 
> 
> 
> > @@ -32,13 +33,28 @@ if [ "$1" == "install" ]; then
>  	systemctl enable lkrg.service
>  	echo -e "       ${P_GREEN}To start ${P_YL}lkrg.service${P_GREEN} please use: ${P_YL}systemctl start lkrg${P_NC}"
>      fi
> +    if [ ! -e "$P_SYSCTL_DIR/lkrg.conf" ]; then
> +	echo -e "       ${P_GREEN}Installing ${P_YL}lkrg.conf${P_GREEN} file under ${P_YL}$P_SYSCTL_DIR${P_GREEN} folder${P_NC}"
> +	mkdir -p "$P_SYSCTL_DIR"
> +	install -pm 644 -o root -g root scripts/bootup/lkrg.conf "$P_SYSCTL_DIR/lkrg.conf"
> 
> No, don't add `mkdir -p`. Like I wrote, let's decide on whether we use
> `-D` or not and then do it consistently for the two uses of `install`.
> That is, if you like `make install` to be creating system directories,
> then just add `-D` to both uses of `install`. Unless I am missing
> something?

Mm, OK.

> >  elif [ "$1" == "uninstall" ]; then
>      echo -e "       ${P_GREEN}Stopping ${P_YL}lkrg.service${P_NC}"
>      systemctl stop lkrg.service
>      echo -e "       ${P_GREEN}Disabling ${P_YL}lkrg.service${P_GREEN} on bootup${P_NC}"
>      systemctl disable lkrg.service
>      echo -e "       ${P_GREEN}Deleting ${P_YL}lkrg.service${P_GREEN} file from the ${P_YL}$P_SYSTEMD_DIR${P_GREEN} folder${P_NC}"
>      rm "$P_SYSTEMD_DIR/lkrg.service"
> +    if cmp -s "$P_SYSCTL_DIR/lkrg.conf" scripts/bootup/lkrg.conf; then
> +        echo -e "       ${P_GREEN}Deleting ${P_YL}lkrg.conf${P_GREEN} file from the ${P_YL}$P_SYSCTL_DIR${P_GREEN} folder${P_NC}"
> 
> I'd add the word "unmodified" after "Deleting".

OK.

> >  elif [ "$1" == "uninstall" ]; then
>      echo -e "       ${P_GREEN}Stopping ${P_YL}lkrg.service${P_NC}"
>      systemctl stop lkrg.service
>      echo -e "       ${P_GREEN}Disabling ${P_YL}lkrg.service${P_GREEN} on bootup${P_NC}"
>      systemctl disable lkrg.service
>      echo -e "       ${P_GREEN}Deleting ${P_YL}lkrg.service${P_GREEN} file from the ${P_YL}$P_SYSTEMD_DIR${P_GREEN} folder${P_NC}"
>      rm "$P_SYSTEMD_DIR/lkrg.service"
> +    if cmp -s "$P_SYSCTL_DIR/lkrg.conf" scripts/bootup/lkrg.conf; then
> +        echo -e "       ${P_GREEN}Deleting ${P_YL}lkrg.conf${P_GREEN} file from the ${P_YL}$P_SYSCTL_DIR${P_GREEN} folder${P_NC}"
> +        rm "$P_SYSCTL_DIR/lkrg.conf"
> +    else
> +        echo -e "       ${P_YL}$P_SYSCTL_DIR/lkrg.service${P_GREEN} was modified, preserve it as ${P_YL}$P_SYSCTL_DIR/lkrg.conf.saved"
> 
> Just "preserving it" and that's it - don't rename.

Why?

> >  elif [ "$1" == "uninstall" ]; then
>      echo -e "       ${P_GREEN}Stopping ${P_YL}lkrg.service${P_NC}"
>      systemctl stop lkrg.service
>      echo -e "       ${P_GREEN}Disabling ${P_YL}lkrg.service${P_GREEN} on bootup${P_NC}"
>      systemctl disable lkrg.service
>      echo -e "       ${P_GREEN}Deleting ${P_YL}lkrg.service${P_GREEN} file from the ${P_YL}$P_SYSTEMD_DIR${P_GREEN} folder${P_NC}"
>      rm "$P_SYSTEMD_DIR/lkrg.service"
> +    if cmp -s "$P_SYSCTL_DIR/lkrg.conf" scripts/bootup/lkrg.conf; then
> +        echo -e "       ${P_GREEN}Deleting ${P_YL}lkrg.conf${P_GREEN} file from the ${P_YL}$P_SYSCTL_DIR${P_GREEN} folder${P_NC}"
> +        rm "$P_SYSCTL_DIR/lkrg.conf"
> +    else
> +        echo -e "       ${P_YL}$P_SYSCTL_DIR/lkrg.service${P_GREEN} was modified, preserve it as ${P_YL}$P_SYSCTL_DIR/lkrg.conf.saved"
> +        echo -e "       ${P_GREEN}If you do not need it anymore, delete it manually"
> +        mv "$P_SYSCTL_DIR/lkrg.conf"{,.saved}
> 
> Drop this `mv` line. This is confusing: the file is possibly still
> loaded, whereas the user would think that it is not. Then if LKRG is
> reinstalled, there will be two files in there.

No, sysctl does not load files from /etc/sysctl.d if they end with
something else than `.conf'.

solardiz:
> > Just "preserving it" and that's it - don't rename.
> Why?

Such as to ease preserving LKRG configuration across LKRG version upgrades. And also I thought `sysctl` would possibly read the file despite of the rename anyway, but you say this isn't so - then renaming isn't that bad.

Do you want this PR merged now?

wladmis:
> > > Just "preserving it" and that's it - don't rename.
> > > Why?
> 
> Such as to ease preserving LKRG configuration across LKRG version upgrades. And also I thought `sysctl` would possibly read the file despite of the rename anyway, but you say this isn't so - then renaming isn't that bad.
> 
> Do you want this PR merged now?

I think it is OK now to merge.

PR #78, opened 2021-04-26
scripts/bootup/systemd/lkrg-systemd.sh: replace 'folder' with 'diretoory' in messages

wladmis:
$subj


solardiz:
I already have pending commits modifying this script in various ways, so your changes would clash. Thank you for trying to help, though!

PR #80, opened 2021-04-27
Do not treat installed proper lkrg.service as an error

wladmis:
$subj

Also, use diff instead of cmp cause diff is line-oriented while cmp is byte-oriented.

### DIFF
--- a/scripts/bootup/systemd/lkrg-systemd.sh
+++ b/scripts/bootup/systemd/lkrg-systemd.sh
@@ -45,7 +45,7 @@ elif [ "$1" == "uninstall" ]; then
 	systemctl disable lkrg.service
 	echo -e "       ${P_GREEN}Deleting ${P_YL}lkrg.service${P_GREEN} file from ${P_YL}$P_SYSTEMD_DIR${P_GREEN} directory${P_NC}"
 	rm "$P_SYSTEMD_DIR/lkrg.service"
-	if cmp -s "$P_SYSCTL_DIR/lkrg.conf" scripts/bootup/lkrg.conf; then
+	if diff -q "$P_SYSCTL_DIR/lkrg.conf" scripts/bootup/lkrg.conf; then
 		echo -e "       ${P_GREEN}Deleting unmodified ${P_YL}lkrg.conf${P_GREEN} file from ${P_YL}$P_SYSCTL_DIR${P_GREEN} directory${P_NC}"
 		rm "$P_SYSCTL_DIR/lkrg.conf"
 	elif [ -e "$P_SYSCTL_DIR/lkrg.conf" ]; then
--- a/scripts/bootup/systemd/lkrg-systemd.sh
+++ b/scripts/bootup/systemd/lkrg-systemd.sh
@@ -24,8 +24,13 @@ echo -e "  ${P_GREEN}[+] ${P_WHITE}Systemd detected${P_NC}"
 
 if [ "$1" == "install" ]; then
 	if [ -e "$P_SYSTEMD_DIR/lkrg.service" ]; then
-		echo -e "       ${P_RED}ERROR! ${P_YL}lkrg.service${P_RED} already exists under ${P_YL}$P_SYSTEMD_DIR${P_RED} directory${P_NC}"
-		exit 1
+		if diff -q "$P_SYSTEMD_DIR/lkrg.service" \
+		   scripts/bootup/systemd/lkrg.service; then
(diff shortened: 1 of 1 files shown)


solardiz:
> Also, use diff instead of cmp cause diff is line-oriented while cmp is byte-oriented.

I don't buy this rationale. A package manager would do an exact binary comparison, and it'd be just right.

I also see no reason not to treat existing service file as an error. Why would we want to allow that?

wladmis:
On Tue, Apr 27, 2021 at 08:30:04AM -0700, Solar Designer wrote:
> > Also, use diff instead of cmp cause diff is line-oriented while cmp is byte-oriented.
> 
> I don't buy this rationale. A package manager would do an exact binary
> comparison, and it'd be just right.

lkrg.conf is more likely a plain text file rather that binary. A package
manager does not know what kind of file it deals with. RPM does not even
compare files, he compares its hashes.

It just common sense that you should use diff for text files and cmp for
binary ones.

> I also see no reason not to treat existing service file as an error.
> Why would we want to allow that?

If user want to update LKRG, he probably just run `make install` rather
than `make deinstall && make install`. I don't see why it must be an
error if installed lkrg.service is equal to one that user want to
install.

solardiz:
>  It just common sense that you should use diff for text files and cmp for binary ones.

I disagree. You should use `diff` in cases where you want to see the differences as text, or/and in cases where you want to ignore some differences (e.g., `diff -w`) or when you need other options specific `diff`. You should use the simpler and thus safer and more reliable program in other cases, which is `cmp`, regardless of file type.

> If user want to update LKRG, he probably just run `make install`

We currently don't support that. Maybe we should? Even if so, with your changes we'd be inconsistent in our approach to replacing the systemd service file (proceed if unchanged, otherwise abort) vs. replacing the kernel module file (replace unconditionally). Right?

solardiz:
What I think we could accept is:

1. Install the systemd service file unconditionally, overwriting the old file. This is like `make install` usually works for other software, and like it'd proceed to do for the kernel module (but please test and confirm).
2. Continue to use `cmp -s` for the `lkrg.conf` file on `make uninstall`.

wladmis:
On Tue, Apr 27, 2021 at 08:52:34AM -0700, Solar Designer wrote:
> >  It just common sense that you should use diff for text files and cmp for binary ones.
> 
> I disagree. You should use `diff` in cases where you want to see the
> differences as text, or/and in cases where you want to ignore some
> differences (e.g., `diff -w`) or when you need other options specific
> `diff`. You should use the simpler and thus safer and more reliable
> program in other cases, which is `cmp`, regardless of file type.

OK.

> > If user want to update LKRG, he probably just run `make install`
> 
> We currently don't support that. Maybe we should? Even if so, with
> your changes we'd be inconsistent in our approach to replacing the
> systemd service file (proceed if unchanged, otherwise abort) vs.
> replacing the kernel module file (replace unconditionally). Right?

Maybe `make install` should ask whether replace of not files if there
are in place to install? But usually `make install` is non-interactive,
so I don't like if this idea. Some `make install` just replaces files
silently, some saves them with '.old' suffix (with warnings).

Do you like make `make install` save old files with suffix and warnings?

solardiz:
I think we should just close this PR now, and later rewrite the whole `make install` thing. We have more issues with the current approach - e.g., installing `lkrg.conf` has nothing to do with systemd in particular, so shouldn't be in the systemd-specific script. The color output may be user-friendly, but isn't expected when we're not running on a tty (e.g. are running from a package manager's build script with output going to a log file). And probably more.

When we rework `make install`, I think it should be more similar to what this make target does in other projects - install/replace the files unconditionally.

wladmis:
On Tue, Apr 27, 2021 at 09:13:54AM -0700, Solar Designer wrote:
> I think we should just close this PR now, and later rewrite the whole
> `make install` thing. We have more issues with the current approach -
> e.g., installing `lkrg.conf` has nothing to do with systemd in
> particular, so shouldn't be in the systemd-specific script. The color
> output may be user-friendly, but isn't expected when we're not running
> on a tty (e.g. are running from a package manager's build script with
> output going to a log file). And probably more.
> 
> When we rework `make install`, I think it should be more similar to
> what this make target does in other projects - install/replace the
> files unconditionally.

OK, I agree.

solardiz:
BTW, when we work on the new `make install`, I think we should also reduce LKRG source tree's use of subdirectories.

Closing this for now.

PR #83, opened 2021-05-12
Contrib SysVinit script

wladmis:
$subj

### DIFF
--- a/contrib/sysvinit/lkrg
+++ b/contrib/sysvinit/lkrg
@@ -1,0 +1,144 @@
+#!/bin/sh
+#
+# LKRG	Linux Kernel Runtime Guard
+#
+# chkconfig: 2345 01 99
+# description: Linux Kernel Runtime Guard
+# config: /etc/sysctl.d/lkrg.conf
+
+# Do not load RH compatibility interface.
+WITHOUT_RC_COMPAT=1
+
+# Source function library.
+. /etc/init.d/functions
+
+NAME="Linux Kernel Runtime Guard"
+CONFIG="/etc/sysctl.d/lkrg.conf"
+RETVAL=0
+
+check_running()
+{
+	grep -qw p_lkrg /proc/modules
+}
+
+check_nolkrg()
+{
+	grep -qw nolkrg /proc/cmdline
+}
+
+start()
+{
+	if check_running; then
+		msg_already_running "$NAME"
+		echo
+		exit 0
+	fi
+
+	if [ "${1-}" != "force" ] && check_nolkrg; then
+		echo "$NAME: nolkrg kernel parameter passed, do not load"
+		echo "To force service start use force-start command"
+		RETVAL=1
+		return $RETVAL
+	fi
+
+	msg_starting "$NAME"
+	modprobe p_lkrg
+	RETVAL=$?
+
+	if [ "$RETVAL" -eq 0 ] && [ -f "$CONFIG" ]; then
+		sysctl -p "$CONFIG" ||
+			echo 'WARNING: could not set settings'
+	fi
+
+	if [ "$RETVAL" -eq 0 ]; then
+		success "$NAME startup"
+	else
+		failure "$NAME startup"
+	fi
+
+	echo
+
+	return $RETVAL
+}
+
+stop()
+{
+	if ! check_running; then
+		msg_not_running "$NAME"
+		echo
+		exit 0
+	fi
+
+	msg_stopping "$NAME"
+	modprobe -r p_lkrg
+	RETVAL=$?
+
+	if [ "$RETVAL" -eq 0 ]; then
+		success "$NAME stopping"
+	else
+		failure "$NAME stopping"
+	fi
+
+	echo
+
+	return $RETVAL
+}
+
+restart()
+{
+	force=
+	if check_running; then
+		force=force
+	fi
+
+	stop
+	start $force
+}
+
+reload()
+{
+	restart
+}
+
+# See how we were called.
+case "$1" in
+	start)
+		start
+		;;
+	force-start)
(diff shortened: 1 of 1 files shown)


contrib/sysvinit/lkrg
```diff
@@ -0,0 +1,144 @@
...
+#
+# chkconfig: 2345 01 99
+# description: Linux Kernel Runtime Guard
+# config: /etc/sysctl.d/lkrg.conf
+
+# Do not load RH compatibility interface.
+WITHOUT_RC_COMPAT=1
+
+# Source function library.
+. /etc/init.d/functions
```
solardiz, line 13: This assumes not only SysVinit, but also certain shell functions available in the library sourced here. It appears that you use `success`, `failure`, and `msg_*`. Anything else? Are these consistently available across distros, or across which families of distros? Also, do they possibly assume that the shell is bash or not?

solardiz (approved):
This is nice to have (thanks!), but let's decide where we put it, whether we have `make install` possibly install it, and whether/where/how we document it.

What distros/configurations did you test this on? We could want to mention specific ones in the documentation.

Also, need to mention this addition in `CHANGES`. Can be in this same PR.

solardiz:
Incidentally, "mkosi boot (mainline) / mkosi-mainline" is failing since yesterday, @vt-alt.

vt-alt:
@solardiz Thanks. Seems kernel package download script failure. I'll PR fix tomorrow (I hope so).

wladmis:
On Wed, May 12, 2021 at 01:43:35PM -0700, Solar Designer wrote:
> @solardiz approved this pull request.
> 
> This is nice to have (thanks!), but let's decide where we put it,
> whether we have `make install` possibly install it, and
> whether/where/how we document it.
> 
> What distros/configurations did you test this on? We could want to
> mention specific ones in the documentation.

I tested this on ALT Sisyphus. IDK whether it works on other systems,
perhaps I should mention that it is a reference example for SysVinit
script. Therefore it should not be installed.

(P.S. We use slightly more complicated script that takes care about
kernel version, but I decided to provide more simple one to not confuse
the readers.)

> Also, need to mention this addition in `CHANGES`. Can be in this same
> PR.

OK.

> > @@ -0,0 +1,144 @@
> +#!/bin/sh
> +#
> +# LKRG	Linux Kernel Runtime Guard
> +#
> +# chkconfig: 2345 01 99
> +# description: Linux Kernel Runtime Guard
> +# config: /etc/sysctl.d/lkrg.conf
> +
> +# Do not load RH compatibility interface.
> +WITHOUT_RC_COMPAT=1
> +
> +# Source function library.
> +. /etc/init.d/functions
> 
> This assumes not only SysVinit, but also certain shell functions
> available in the library sourced here. It appears that you use
> `success`, `failure`, and `msg_*`. Anything else? Are these
> consistently available across distros, or across which families of
> distros?

IDK, but I saw smth like these in PLD Linux scripts.

> Also, do they possibly assume that the shell is bash or not?

I don't see any bash specific features in the script, and shellcheck as
well, but yeah, old RedHat Linuces^WLinux distros used bash as /bin/sh,
but Debian uses (d)ash for /bin/sh IIRC.

-- 
   WBR,
   Vladimir D. Seleznev

solardiz:
Searching for `msg_*`, these appear to be specific to PLD and ALT.

I think it'd make more sense to contribute a more generic init script and place it in `scripts/bootup` and have `lkrg-bootup.sh` install it when the distro doesn't use systemd.

Or just have it in `scripts/bootup` as an example, even if it's unused yet.

Which are currently the most common distros without systemd (or configured that way)? Maybe a Debian configuration? Perhaps the script should work on those out-of-the-box.

Also need to mention this in `LICENSE`,  `debian/copyright`, and `CHANGES`.

While we could accept whatever into `contrib/`, we already have various contributed stuff in the tree that isn't in `contrib/`, so introducing that subdirectory now would be somewhat inconsistent.

solardiz:
> have `lkrg-bootup.sh` install it when the distro doesn't use systemd.

If so, we'd also need to edit the "Installation" section in `README`.

wladmis:
> Searching for `msg_*`, these appear to be specific to PLD and ALT.

OK.

> I think it'd make more sense to contribute a more generic init script 

Agree. I wrote perhaps the most generic init script that is possible to write some time ago. Should I create a new PR for this?

> and place it in `scripts/bootup` and have `lkrg-bootup.sh` install it when the distro doesn't use systemd.

I think it is wrong to assume that if the systemd does not use systemd it uses sysvinit. There are some others options available such as OpenRC or runit/daemontools-like. 

> Or just have it in `scripts/bootup` as an example, even if it's unused yet.
>
> Which are currently the most common distros without systemd (or configured that way)?

It's hard to say but I assume the most common systemd-less distros do not use sysvinit.

> Maybe a Debian configuration? Perhaps the script should work on those out-of-the-box.
> 
> Also need to mention this in `LICENSE`, `debian/copyright`, and `CHANGES`.

OK.

> While we could accept whatever into `contrib/`, we already have various contributed stuff in the tree that isn't in `contrib/`, so introducing that subdirectory now would be somewhat inconsistent.

OK.

solardiz:
> I wrote perhaps the most generic init script that is possible to write some time ago. Should I create a new PR for this?

I think only if the PR is complete enough for its inclusion to benefit LKRG users.

> I think it is wrong to assume that if the systemd does not use systemd it uses sysvinit. There are some others options available such as OpenRC or runit/daemontools-like.

On one hand, yes, this isn't a reliable assumption. On the other hand, it is reasonable to have a fallback choice that might or might not be correct. Then possibly add checks for other init systems before reaching that fallback.

PR #91, opened 2021-05-18
Keep just one sysctl config file

### LINKED ISSUE #88: Keep just one sysctl config file
Due to recent contributions, we have `scripts/bootup/lkrg.conf` and `debian/lkrg-dkms.sysctl`. This is redundant. Since `debian/lkrg-dkms.sysctl` is more complete, perhaps it should be renamed/moved over `scripts/bootup/lkrg.conf` and that file used by the Debian packaging as well.

Further, `debian/lkrg-dkms.sysctl` has a copy of our sysctl descriptions from `README`. When making edits to those, are we going to manually keep them in sync between these two files or should we possibly drop them from one of the two files and refer from it to the other?

morfikov:
I moved the content of the `debian/lkrg-dkms.sysctl` file to the `scripts/bootup/lkrg.conf` file. I also removed the header of the file saying where the file should be put. Now the debian build system picks up the `scripts/bootup/lkrg.conf` and installs it under `/etc/sysctl.d/30-lkrg-dkms.conf`.

(Closes #88)


solardiz (approved):
Looks good to me.

Where does the style with two spaces between the `#` characters and the actual comments come from, though? Is it some convention we need to follow, or can we get rid of it? (It looks weird to me.)

morfikov:
> Where does the style with two spaces between the `#` characters and the actual comments come from, though? Is it some convention we need to follow, or can we get rid of it? (It looks weird to me.)

I just copied it from the README file. I just inserted the `#` at the beginning of the lines.

solardiz:
> I just copied it from the README file. I just inserted the `#` at the beginning of the lines.

Oh, in there the two spaces were to align with the list items. This doesn't apply here. So please change it to one space and force-push. Thanks!

morfikov:
Should be good now.

solardiz:
> Should be good now.

Oh, I was too quick to merge. Turns out you removed the double spaces not only after the `#`, but also between sentences. I suggest you restore those (by starting from the previous revision of this file and only removing the right double spaces from it)... with a separate PR now.

solardiz:
> with a separate PR now.

Or actually, I'll fix that myself.

solardiz:
It's a matter of style of English text. I'd like to be the same as the documentation's.

solardiz:
@morfikov I've amended your commit and force-pushed. I hope this was quick enough (that no one cloned the tree in those minutes), and is OK with you.

morfikov:
To be honest, I've never heared about this "multiple spaces style of English text" . I've seen multiple `-` but spaces? Never.

morfikov:
I've just read about this double spaces, and a[ccording to the wik](https://en.wikipedia.org/wiki/Sentence_spacing)i:

> The desired or correct sentence spacing is often debated but some sources now claim an additional space is not necessary.[9] From around 1950, single sentence spacing became standard in books, magazines, and newspapers,[10] and the majority of style guides that use a Latin-derived alphabet as a language base now prescribe or recommend the use of a single space after the concluding punctuation of a sentence.[11]

So I would throw the double spaces away and stick with the single ones. :)

solardiz:
Then you haven't seen much yet. ;-) This is a typewriter thing that carried over to computers when using fixed-width fonts. As usage of fixed-width font became rare, this habit grew out of date. I think in our case it's still relevant, and helps keep the texts more readable. I don't use double spaces e.g. when writing GitHub comments or when writing HTML web pages (they wouldn't propagate anyway) or when using a modern WYSIWYG word processor, but I do in text files and commit messages.

PR #98, opened 2021-06-15
CI: Build on different OS using Docker containers

vt-alt:
Test build on Fedora, CentOS, and ALT.

https://github.com/openwall/lkrg/pull/61#issuecomment-813011209
> I think we should keep Ubuntu, but also add perhaps CentOS 7 and 8. Instead of CentOS 8, it can be (and will eventually have to be) AlmaLinux or (later) RockyLinux. Can also have Fedora or CentOS Stream - would probably let us catch issues (typically back-ports that LKRG needs to adapt to) before they appear in a RHEL/CentOS/AlmaLinux/RockyLinux release.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -1,0 +1,30 @@
+name: Multi-OS container builds
+on: [push, pull_request]
+
+jobs:
+    build:
+        runs-on: ubuntu-20.04
+        strategy:
+            fail-fast: false
+            matrix:
+                include:
+                    - image: centos:7
+                      run: yum install -y kernel-devel kernel gcc make elfutils-libelf-devel
+                    - image: centos:8
+                      run: yum install -y kernel-devel kernel gcc make elfutils-libelf-devel
+                    - image: quay.io/centos/centos:stream8
+                      run: yum install -y kernel-devel kernel gcc make elfutils-libelf-devel
+                    - image: fedora:latest
+                      run: dnf install -y kernel-devel kernel gcc make elfutils-libelf-devel
+                    - image: alt:sisyphus
+                      run: apt-get update; apt-get install -y kernel-headers-modules-un-def gcc make libelf-devel
+
+        container:
+            image: ${{ matrix.image }}
+
+        steps:
+            - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


vt-alt:
I also thinking about building on ARM. But, there is three options:
- Cross-build inside of Docker (con: need to manually download and extract full kernel headers).
- qemu-user build in Docker (con: slower, because there is no KVM in GA, but faster than qemu-system).
- Use free Travis-CI on ARM (as suggested-by @ldv-alt).

solardiz:
Thank you, @vt-alt!

>     * Use free Travis-CI on ARM (as suggested-by @ldv-alt).

This started often failing on `sudo m-a prepare` for me (missing package with proper kernel headers version), which is why I disabled it.

vt-alt:
> This started often failing on `sudo m-a prepare` for me (missing package with proper kernel headers version), which is why I disabled it.

Maybe it was requiring `apt-get update`? (Outdated APT index can be confusing. Maybe they created VM from the local (Travis mirror) repo, but Module Assistant goes to the upstream repo which is not 100% synchronized with the local. Just a theory.)

Another idea is just to install `linux-virtual` and not use `m-a` at all (like I did for `mkosi`).

solardiz:
> Maybe it was requiring `apt-get update`?

I think we already had that (so it didn't help or at least wasn't sufficient to avoid that problem):

```
addons:
  apt:
    update: true
    packages:
      - module-assistant
```

solardiz:
> Another idea is just to install `linux-virtual` and not use `m-a` at all (like I did for `mkosi`).

Maybe you'd give this a try in your fork of the repo, then send us a PR if it helps?

vt-alt:
Btw, should LKRG support OpenVZ's vzkernel? If yes which versions (6-8)?

solardiz:
> should LKRG support OpenVZ's vzkernel? If yes which versions (6-8) and distros?

Yes. We previously tested using Virtuozzo 7 upstream kernel builds. Currently, I'd expect LKRG to work with 7 and 8, including hopefully with custom builds of those. If anything doesn't build/work there, it will likely be easy to fix in LKRG.

Issue #100, opened 2021-07-20
LKRG fails to load on Debian 10


gnd:
Hi, 

we are using LKRG on a Debian 10 machine since a year without any problems. However after a recent reboot, we have acquired a new kernel and thus recompiled LKRG from the latest github. This latest version will not load because:

```
[Tue Jul 20 13:17:28 2021] p_lkrg: loading out-of-tree module taints kernel.
[Tue Jul 20 13:17:28 2021] p_lkrg: module verification failed: signature and/or required key missing - tainting kernel
[Tue Jul 20 13:17:28 2021] p_lkrg: Unknown symbol __module_address (err -2)
[Tue Jul 20 13:17:28 2021] p_lkrg: Unknown symbol __module_text_address (err -2)
[Tue Jul 20 13:18:40 2021] p_lkrg: Unknown symbol __module_address (err -2)
[Tue Jul 20 13:18:40 2021] p_lkrg: Unknown symbol __module_text_address (err -2)
[Tue Jul 20 13:21:28 2021] p_lkrg: Unknown symbol __module_address (err -2)
[Tue Jul 20 13:21:28 2021] p_lkrg: Unknown symbol __module_text_address (err -2)
```

Do you have any ideas how to proceed using LKRG ? Thanks a lot. 

```
uname -a
Linux 4.19.0-17-cloud-amd64 #1 SMP Debian 4.19.194-2 (2021-06-21) x86_64 GNU/Linux
```

solardiz:
Hi @gnd. Thanks for reporting. This looks similar to #93. Perhaps upstream back-ported this change even further, to 4.19 kernels. We'll look into that.

oshogbo:
Just out of curiosity can you run:
```
$ cat /proc/kallsyms | grep __module_address
```

solardiz:
I think the tasks here are to review upstream 4.19.* ChangeLogs (maybe also for even older branches) or stable list postings to confirm the guess and determine the exact kernel revisions where the changes were made. Then adjust LKRG accordingly. Alternatively, we could enable our hack on all kernels (or maybe starting with the oldest currently maintained upstream branch) so that this issue won't bite us again.

@oshogbo Do you possibly have time for this now? ;-)

gnd:
```

oshogbo:
@solardiz I can look into that. Unfortunately I don't have such kernel so it will take me a moment.

solardiz:
[...]

PR #101, opened 2021-07-20
Fix LKRG builds on new kernels

oshogbo:
### Description
Deduplicate use of __module_text_address.
Add more kernels where the unexported __module_address.

### How Has This Been Tested?
Build and loaded on 5.8.0-45

This should solve #100

### DIFF
--- a/src/p_lkrg_main.h
+++ b/src/p_lkrg_main.h
@@ -116,6 +116,13 @@
 #define P_SELINUX_VERIFY
 #endif
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,9,0) || \
+ (LINUX_VERSION_CODE >= KERNEL_VERSION(5,4,118) && LINUX_VERSION_CODE < KERNEL_VERSION(5,5,0)) || \
+ (LINUX_VERSION_CODE >= KERNEL_VERSION(4,19,191) && LINUX_VERSION_CODE < KERNEL_VERSION(4,20,0)) || \
+ (LINUX_VERSION_CODE >= KERNEL_VERSION(4,14,233) && LINUX_VERSION_CODE < KERNEL_VERSION(4,15,0))
+#define P_LKRG_UNEXPORTED_MODULE_ADDRESS
+#endif
+
 #define nitems(val)      (sizeof(val) / sizeof(val[0]))
 
 typedef struct _p_lkrg_global_conf_structure {
@@ -219,9 +226,7 @@ typedef struct _p_lkrg_global_symbols_structure {
    void (*p_native_write_cr4)(unsigned long p_val);
  #endif
 #endif
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,9,0) || \
- (LINUX_VERSION_CODE >= KERNEL_VERSION(5,4,118) && LINUX_VERSION_CODE < KERNEL_VERSION(5,5,0))
-#define P_LKRG_UNEXPORTED_MODULE_ADDRESS
+#ifdef P_LKRG_UNEXPORTED_MODULE_ADDRESS
    struct module* (*p_module_address)(unsigned long p_val);
    struct module* (*p_module_text_address)(unsigned long p_val);
 #endif
--- a/src/p_lkrg_main.h
+++ b/src/p_lkrg_main.h
@@ -242,6 +242,14 @@ typedef struct _p_lkrg_global_symbols_structure {
 
 } p_lkrg_global_syms;
 
+#ifdef P_LKRG_UNEXPORTED_MODULE_ADDRESS
+#define LKRG_P_MODULE_ADDRESS(p_addr)      P_SYM(p_module_address)(p_addr)
+#define LKRG_P_MODULE_TEXT_ADDRESS(p_addr) P_SYM(p_module_text_address)(p_addr)
+#else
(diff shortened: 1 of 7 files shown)


solardiz:
> Introduce LKRG_P_MODULE_ADDRESS and LKRG_P_MODULE_TEXT_ADDRESS.

In this commit, perhaps also patch the other instance in `p_kmod.c` and completely get rid of `P_LKRG_UNEXPORTED_MODULE_ADDRESS`?

src/p_lkrg_main.h
```diff
@@ -116,6 +116,13 @@
 #define P_SELINUX_VERIFY
 #endif
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,9,0) || \
+ (LINUX_VERSION_CODE >= KERNEL_VERSION(5,4,118) && LINUX_VERSION_CODE < KERNEL_VERSION(5,5,0)) || \
+ (LINUX_VERSION_CODE >= KERNEL_VERSION(4,19,191) && LINUX_VERSION_CODE < KERNEL_VERSION(4,20,0)) || \
+ (LINUX_VERSION_CODE >= KERNEL_VERSION(4,14,233) && LINUX_VERSION_CODE < KERNEL_VERSION(4,21,0))
```
solardiz, line 122: I think it should be `LINUX_VERSION_CODE < KERNEL_VERSION(4,15,0))` on the last line here.

Also, "kernels" is mistyped in the commit message. And I'd make it "stable kernels".

Oh, and please update `CHANGES`.

oshogbo:
Ugh, sorry.

solardiz:
This looks acceptable for merging now, but perhaps you can improve these two things:

1. Completely drop `P_LKRG_UNEXPORTED_MODULE_ADDRESS`, unless I'm missing a reason why you still have it? That is, instead of conditionally defining this symbol do (only) your if/else replacement defines right away.
2. Update `CHANGES` to reflect that these specific new versions in these two additional stable branches are now supported.

Change this line:

```
 *) Support new stable kernels 5.4.118+
```

to list the two additional kernel revisions as well.

Adam-pi3:
@oshogbo I'm wondering if you tried to run these changes on exactly the same environment as #100 

Btw. I've just seen that `mkosi` is failing with the following problem:

```
[   10.494659] p_lkrg: loading out-of-tree module taints kernel.
[   10.495813] p_lkrg: module verification failed: signature and/or required key missing - tainting kernel
[   10.500162] [p_lkrg] Loading LKRG...
[   10.500625] [p_lkrg] System does NOT support SMEP. LKRG can't enforce SMEP validation :(
[   10.501364] [p_lkrg] System does NOT support SMAP. LKRG can't enforce SMAP validation :(
[   10.588498] Freezing user space processes ... (elapsed 0.002 seconds) done.
[   10.591480] OOM killer disabled.
[   11.781970] [p_lkrg] [kretprobe] register_kretprobe() for <__x32_compat_sys_keyctl> failed! [err=-2]
[   11.782550] [p_lkrg] ERROR: Can't hook x32_sys_keyctl :(
[   12.742620] [p_lkrg] Can't initialize exploit detection features! Exiting...
```

This is unrelated bug to the one which this PR addresses but I'm wondering if @oshogbo would you have some time to take a look on this as well?

Adam-pi3:
I think it is related to:
https://github.com/torvalds/linux/commit/2e958a8a510d956ec8528f0bd20e309b5bb5156c

oshogbo:
@solardiz 
1. I'm not sure if we can get rid of P_LKRG_UNEXPORTED_MODULE_ADDRESS. We still used it in p_lkrg_main.h. In case of this file - I prefer this define over moving everything to the structure definition or to have multiple checks with LINUX_VERSION. But if you prefer moving everything to the structure definition I can do that.
2. Addressed.

@Adam-pi3
> I'm wondering if you tried to run these changes on exactly the same environment as #100
No, I have not. I verified only on  5.8.0-45.

> This is unrelated bug to the one which this PR addresses but I'm wondering if @oshogbo would you have some time to take a look on this as well?
I can look into the other bug as well.

solardiz:
> Btw. I've just seen that `mkosi` is failing with the following problem:

Does it ultimately report failure to GitHub? If not, then this would be something for @vt-alt to fix.

> [   11.781970] [p_lkrg] [kretprobe] register_kretprobe() for <__x32_compat_sys_keyctl> failed! [err=-2]
> [   11.782550] [p_lkrg] ERROR: Can't hook x32_sys_keyctl :(

Please open a separate issue for this.

vt-alt:
> Does it ultimately report failure to GitHub? If not, then this would be something for @vt-alt to fix.

(In quotation) the kernel seems booted normally. So, what to fix?

solardiz:
> the kernel seems booted normally.

Yes, but without LKRG:

```
[   12.742620] [p_lkrg] Can't initialize exploit detection features! Exiting...
```

> So, what to fix?

I was thinking that maybe we were not detecting that LKRG didn't load successfully. However, as I recall you did include a check for its messages on successful loading.

Adam-pi3:
@oshogbo I've just pushed the fix for the COMPAT naming problem here:
https://github.com/openwall/lkrg/commit/3fd154d334b263a2590c1380cf7d29cade379d97

oshogbo:
@Adam-pi3 Thanks. From some reason I didn't check the 5.14... and I wasn't able to reproduce the problem. But I'm happy you fixed it!

PR #102, opened 2021-07-21
Fix building kernel 5.13.X line.

oshogbo:
### Description

In the c6d93ae9df2519a66a6d5ec9e226f8458c51815c we mentioned that
the task_struct had been changed, but actually the change is since
5.14. The task_struct structure is the same still in 5.13.4.

Source:
https://elixir.bootlin.com/linux/v5.13.2/source/include/linux/sched.h#L666
https://elixir.bootlin.com/linux/v5.13.3/source/include/linux/sched.h#L666
https://elixir.bootlin.com/linux/v5.13.4/source/include/linux/sched.h#L666
https://elixir.bootlin.com/linux/v5.14-rc1/source/include/linux/sched.h#L669

### How Has This Been Tested?
Build on:
`
Linux lkrg-ubu 5.13.2-051302-generic #202107141437 SMP Wed Jul 14 18:40:02 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
`

### DIFF
--- a/src/modules/exploit_detection/p_exploit_detection.h
+++ b/src/modules/exploit_detection/p_exploit_detection.h
@@ -147,7 +147,7 @@
 
 static inline unsigned int p_get_task_state(struct task_struct *p_task) {
 
-#if LINUX_VERSION_CODE > KERNEL_VERSION(5,13,0)
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,14,0)
    return READ_ONCE(p_task->__state);
 #else
    return READ_ONCE(p_task->state);


solardiz:
Oh, now everything makes sense. Thanks!

(We'll also need to update `CHANGES` and `README` to declare 5.14 support, but it's too early for that because of the other issue preventing LKRG from loading on 5.14. So shouldn't be in this PR.)

solardiz:
BTW, Adam, on further occasions let's not use the "5.13+" syntax to mean anything other than "5.13 inclusive, and above" (I'm referring to your commit message). In special cases like what we thought we had there, we can use clear wording like "kernels above 5.13, but not 5.13 itself". Yes, I recall you also included a clarification like that, but the "+" syntax is inconsistent with that. Thanks!

Adam-pi3:
I've just pushed the commit where I tried to avoid it (https://github.com/openwall/lkrg/commit/3fd154d334b263a2590c1380cf7d29cade379d97)

Adam-pi3 (approved):

solardiz:
> I've just pushed the commit where I tried to avoid it ([3fd154d](https://github.com/openwall/lkrg/commit/3fd154d334b263a2590c1380cf7d29cade379d97))

Actually, I think it's a case where "5.14+" would be appropriate, because it literally means "5.14 and above" - the wording you used. I guess I somehow managed to confuse you. %-)

PR #103, opened 2021-07-24
delete superfluous file debian/files

adrelanos:
Why?

* It's superfluous.
* I don't remember seeing it in other packages.
* Its content are versioned. Unnecessary maintenance overhead to keep updated.
* Its contents change during build using debuild.
* Its contents are autogenerated.

### DIFF
--- a/debian/files
+++ b/debian/files
@@ -1,1 +1,0 @@
-lkrg_0.9.1-2_source.buildinfo kernel optional


solardiz:
@morfikov Please comment on this. Thanks!

morfikov:
It can be deleted.

PR #104, opened 2021-07-24
Debian buster packaging compatibility

adrelanos:
Fixes this issue when building with cowbuilder on Debian buster.

```
The following packages have unmet dependencies:
 pbuilder-satisfydepends-dummy : Depends: debhelper-compat (= 13) which is a virtual package, provided by:
                                          - debhelper (12.1.1) provides debhelper-compat=12, but 12.1.1 is installed
                                          - debhelper (12.1.1) provides debhelper-compat=11, but 12.1.1 is installed
                                          - debhelper (12.1.1) provides debhelper-compat=10, but 12.1.1 is installed
                                          - debhelper (12.1.1) provides debhelper-compat=9, but 12.1.1 is installed

Unable to resolve dependencies!  Giving up...
```


morfikov:
I wouldn't do that, since v13 is the current debian compatibility level. You can read more about it here:

https://manpages.debian.org/testing/debhelper/debhelper.7.en.html#COMPATIBILITY_LEVELS

adrelanos:
> https://manpages.debian.org/testing/...


'testing' isn't 'stable'. It's unreleased still.

morfikov:
If you look closely, in the `debian/changelog` file you have `lkrg (0.9.1-2) unstable;` , so this is the build directory for unstable/testing Debian branch.

solardiz:
@adrelanos @morfikov I assume this one PR is on hold until the two of you hopefully arrive at a common opinion. Thanks!

morfikov:
There's not much to add from my side. The packaging flow in debian is from unstable->testing->stable. Which is basically why all packages are built for unstable in the first place. Then after some time, the package goes to the testing branch and sits there for a little bit longer period of time. Once in a while, they create a stable release from the testing branch. But you always should aim to package some project to unstable first and if some dependencies are broken between stable and testing/unstable, [you should backport the package.](https://backports.debian.org/) . But in the case of lkrg, I think the debian/ dir is in the shape as it should be and nothing should be changed.

solardiz:
@morfikov I don't know if those aspects you mention are relevant to us or perhaps currently not. This isn't a package _in_ Debian, but is a package _for_ Debian(-like) distros. If this PR fixes a real issue for @adrelanos, then why not merge it? Does merging this PR break anything for anyone?

morfikov:
The purpose of the debian/ dir is to debianize the project source in order to build the .deb package and upload it to some repository (either the project's one or the main debian repo), so the end users could install it via apt/dpkg. 

Building the project from the source on any given system is a totally different matter -- you don't even need the debian/ dir at all to build the project from source, even on some ancient 10y-old debian system. 

When I prepared the debian/ dir, I wanted to follow the current guidelines from the debian reference. This includes silencing every lintian warnings/error which appeared during preparation of the dir. That's why you have the debian/ dir in the current shape. If you don't want to follow the current debian standards, you can remove the debian/ dir altogether since sooner or later it will become a mess that prevents building the deb package on any system including the current one. But it's up to you whether you want to have consistent and following current standard debian/ dir, or not.

solardiz:
I'm not into Debian packaging and don't have much say on any of this, but my gut feeling is that ideally the `debian/` directory in the project should be useful to the maximum number of users and packagers. If we can make it more useful to @adrelanos by merging this PR without hurting anyone else's use, then I think we should. So once again, "Does merging this PR break anything for anyone?"

morfikov:
I don't know, we should ask [lintian](https://lintian.debian.org/). :)

morfikov:
Just to let you know: [bullseye release planned on 2021-08-14](https://lists.debian.org/debian-devel-announce/2021/07/msg00003.html)

adrelanos:
> Just to let you know: [bullseye release planned on 2021-08-14](https://lists.debian.org/debian-devel-announce/2021/07/msg00003.html)

Yes, in this specific case not important since `buster` will become `old-stable`, obsolete soon but the general issue will happen again soon.

> @morfikov I don't know if those aspects you mention are relevant to us or perhaps currently not. This isn't a package _in_ Debian, but is a package _for_ Debian(-like) distros.

That's the critical point. What's the goal of the the `/debian` folder? For me:

* Widest usefulness for most users as easy as possible (no further changes required).
* A template with ideally zero delta for creation of [prebuild packages](https://www.whonix.org/wiki/Linux_Kernel_Runtime_Guard_LKRG).

Some who use LKRG on Debian might want to use any prebuild package such as from the Whonix repository once that's updated again. That version of the `/debian` folder could be forked for the current stable release. Though, I'd like to keep the delta with the LKRG upstream repository ideally zero or minimal.

For other users of Debian `stable` who wish the build the LKRG Debian package from source code in the the future the `/debian` folder would be unusable because of following the latest standards of Debian, i.e. Debian `testing`.

In Debian this is solved by maintaining multiple git branches. One per suite, i.e. `stable`, `testing`. (They use specific suite names such as `buster`, `bullseye` instead of generic suite names such as `stable`, `testing`.) That approach however seems overkill for me inside the LKRG repository.

Hence my approach was making it compatible with the current Debian `stable` release.

Otherwise how could Debian `stable` users make use of it? Then there would have to be some instructions "but as a user of Debian `stable` you first need to do this and this".

Adam-pi3:
@morfikov and @adrelanos are you agreeing that merging this PR is fine?
Same as @solardiz I don't have expertise to make the correct decision in that field.

morfikov:
I would leave it as it is. In two weeks they release the new Debian stable, and everything will be working just fine.

adrelanos:
I can live without Debian buster compatibly in this upstream repository but general question what the purpose of the `/debian` folder should be presented in https://github.com/openwall/lkrg/pull/104#issuecomment-887069841 remains.

morfikov:
Since the new stable Debian was released a few days ago. I think this can be closed.

Issue #74, opened 2021-04-20
Criticism of LKRG systemd service


solardiz:
https://linuxreviews.org/Linux_Kernel_Runtime_Guard includes this criticism:

> Run sudo make install to install it. It will very impolitely try to enable itself with systemd the moment it is installed by means of a lkrg.service file installed in /etc/systemd/system. If will fail since it doesn't run systemctl daemon-reload to make systemd aware of it and print out an error saying Failed to enable unit: Unit file lkrg.service does not exist. [...]
> 
> [...]
> 
> You may want to eradicate /etc/systemd/system/lkrg.service if you plan on loading it using a file in /etc/modprobe.d/ or other non-systemd means. Preventing the systemd service from loading with systemctl mask lkrg.service won't work since LKRG foolishly puts a file, not a symbolic link, in /etc/systemd/system/.

Is any of this criticism valid? Should we address it, how, and when - probably not in a rushed 0.9.1 release yet (as we might introduce new issues if we do, so need some more testing by users before release)?

morfikov:
To me the LKRG systemd's service doesn't make any sense. You could probably pick up the `modprobe`, `modules` and `sysctl` files from the `debian/` dir and rely on the init to load and configure the module. No extra service is needed here, at least in the case of systemd. If you decide to get rid of the LKRG systemd's service, the config in `debian/` dir has to be updated (in the `debian/rules` file) to use the provided files instead.

solardiz:
We had discussed this on lkrg-users, and IIRC the conclusion was that we needed more control over when exactly the LKRG module is loaded into the kernel and to ensure any sysctl changes would be applied shortly thereafter. Using those separate mechanisms, the timings could be worse and there would be more distro-specific uncertainty. However, we might need to re-read that discussion.

solardiz:
In that context, it is arguably a bug that `debian/` would load LKRG at a different time than `make install` does. Right now, I'd say the bug is in `debian/`.

solardiz:
For the timings, we have this:

```
After=systemd-modules-load.service
Before=systemd-sysctl.service
Before=sysinit.target shutdown.target
```

This currently looks to me quite similar to loading via `modules`, except that we explicitly make LKRG load _after_ all of the other modules listed in there. I am not currently aware of specific reasons for that to be required, but it can probably help avoid triggering some race condition issues between modules going forward.

Possibly more importantly, we also have this:

```
ConditionKernelCommandLine=!nolkrg
```

[...]

PR #107, opened 2021-07-25
fix path to /etc/sysctl.d/30-lkrg-dkms.conf

adrelanos:
https://github.com/openwall/lkrg/blob/main/debian/rules#L22 install [`scripts/bootup/lkrg.conf`](https://github.com/openwall/lkrg/blob/main/scripts/bootup/lkrg.conf) it to `/etc/sysctl.d/30-lkrg-dkms.conf`.

(The prefix `30-lkrg` seems right for a `d.` directory, i.e. for `/etc/sysctl.d/`.)

Not to `/etc/sysctl.d/lkrg.conf` as referenced in [`scripts/bootup/systemd/lkrg.service`](https://github.com/openwall/lkrg/blob/main/scripts/bootup/systemd/lkrg.service).

So in order to fix the Debian package (make LKRG work after installation) some change is required.

Currently this issue is happening:

```
user@localhost:~$ sudo systemctl status lkrg
```
```
● lkrg.service - Linux Kernel Runtime Guard
   Loaded: loaded (/lib/systemd/system/lkrg.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Sun 2021-07-25 15:40:37 UTC; 4s ago
  Process: 5539 ExecStart=/sbin/modprobe -v p_lkrg (code=exited, status=0/SUCCESS)
  Process: 5540 ExecStartPost=/sbin/sysctl -p /etc/sysctl.d/lkrg.conf (code=exited, status=255/EXCEPTION)
 Main PID: 5539 (code=exited, status=0/SUCCESS)

Jul 25 15:40:37 localhost.localdomain systemd[1]: Starting Linux Kernel Runtime Guard...
Jul 25 15:40:37 localhost.localdomain sysctl[5540]: sysctl: cannot open "/etc/sysctl.d/lkrg.conf": No such file or directory
Jul 25 15:40:37 localhost.localdomain systemd[1]: lkrg.service: Control process exited, code=exited, status=255/EXCEPTION
Jul 25 15:40:37 localhost.localdomain systemd[1]: lkrg.service: Failed with result 'exit-code'.
Jul 25 15:40:37 localhost.localdomain systemd[1]: Failed to start Linux Kernel Runtime Guard.
```

Issue being `sysctl: cannot open "/etc/sysctl.d/lkrg.conf": No such file or directory`.

### DIFF
--- a/scripts/bootup/systemd/lkrg.service
+++ b/scripts/bootup/systemd/lkrg.service
@@ -18,7 +18,7 @@ ConditionKernelCommandLine=!nolkrg
 [Service]
 Type=oneshot
 ExecStart=/sbin/modprobe -v p_lkrg
-ExecStartPost=/sbin/sysctl -p /etc/sysctl.d/lkrg.conf
+ExecStartPost=/sbin/sysctl -p /etc/sysctl.d/01-lkrg.conf
 ExecStop=/sbin/modprobe -v -r p_lkrg
 RemainAfterExit=yes
 
--- a/debian/rules
+++ b/debian/rules
@@ -17,9 +17,9 @@ override_dh_install:
 	dh_install -p lkrg-systemd scripts/bootup/systemd/lkrg.service lib/systemd/system
 	mkdir -p debian/tmp/etc/sysctl.d
 	mkdir -p debian/tmp/etc/modules-load.d
-	cp scripts/bootup/lkrg.conf debian/tmp/etc/sysctl.d/30-lkrg-dkms.conf
+	cp scripts/bootup/lkrg.conf debian/tmp/etc/sysctl.d/01-lkrg.conf
 	cp debian/lkrg-dkms.modules debian/tmp/etc/modules-load.d/lkrg-dkms.conf
-	dh_install -p lkrg-dkms    debian/tmp/etc/sysctl.d/30-lkrg-dkms.conf etc/sysctl.d
+	dh_install -p lkrg-dkms    debian/tmp/etc/sysctl.d/01-lkrg.conf etc/sysctl.d
 	dh_install -p lkrg-dkms    debian/tmp/etc/modules-load.d/lkrg-dkms.conf etc/modules-load.d
 
 override_dh_dkms:
--- a/scripts/bootup/systemd/lkrg-systemd.sh
+++ b/scripts/bootup/systemd/lkrg-systemd.sh
@@ -28,11 +28,11 @@ if [ "$1" == "install" ]; then
 		echo -e "       ${P_GREEN}To start ${P_YL}lkrg.service${P_GREEN} please use: ${P_YL}systemctl start lkrg${P_NC}"
 		echo -e "       ${P_GREEN}To enable ${P_YL}lkrg.service${P_GREEN} on bootup please use: ${P_YL}systemctl enable lkrg.service${P_NC}"
 	fi
-	if [ -e "$P_SYSCTL_DIR/lkrg.conf" ]; then
-		echo -e "       ${P_YL}lkrg.conf${P_GREEN} is already installed, skipping${P_NC}"
+	if [ -e "$P_SYSCTL_DIR/01-lkrg.conf" ]; then
+		echo -e "       ${P_YL}01-lkrg.conf${P_GREEN} is already installed, skipping${P_NC}"
 	else
-		echo -e "       ${P_GREEN}Installing ${P_YL}lkrg.conf${P_GREEN} file under ${P_YL}$P_SYSCTL_DIR${P_GREEN} directory${P_NC}"
-		install -pm 644 -o root -g root scripts/bootup/lkrg.conf "$P_SYSCTL_DIR/lkrg.conf"
+		echo -e "       ${P_GREEN}Installing ${P_YL}01-lkrg.conf${P_GREEN} file under ${P_YL}$P_SYSCTL_DIR${P_GREEN} directory${P_NC}"
+		install -pm 644 -o root -g root scripts/bootup/lkrg.conf "$P_SYSCTL_DIR/01-lkrg.conf"
 	fi
 elif [ "$1" == "uninstall" ]; then
 	echo -e "       ${P_GREEN}Stopping ${P_YL}lkrg.service${P_NC}"
@@ -41,13 +41,13 @@ elif [ "$1" == "uninstall" ]; then
 	systemctl disable lkrg.service
 	echo -e "       ${P_GREEN}Deleting ${P_YL}lkrg.service${P_GREEN} file from ${P_YL}$P_SYSTEMD_DIR${P_GREEN} directory${P_NC}"
 	rm "$P_SYSTEMD_DIR/lkrg.service"
-	if cmp -s "$P_SYSCTL_DIR/lkrg.conf" scripts/bootup/lkrg.conf; then
-		echo -e "       ${P_GREEN}Deleting unmodified ${P_YL}lkrg.conf${P_GREEN} file from ${P_YL}$P_SYSCTL_DIR${P_GREEN} directory${P_NC}"
-		rm "$P_SYSCTL_DIR/lkrg.conf"
-	elif [ -e "$P_SYSCTL_DIR/lkrg.conf" ]; then
-		echo -e "       ${P_YL}$P_SYSCTL_DIR/lkrg.conf${P_GREEN} was modified, preserving it as ${P_YL}$P_SYSCTL_DIR/lkrg.conf.saved${P_NC}"
+	if cmp -s "$P_SYSCTL_DIR/01-lkrg.conf" scripts/bootup/lkrg.conf; then
+		echo -e "       ${P_GREEN}Deleting unmodified ${P_YL}01-lkrg.conf${P_GREEN} file from ${P_YL}$P_SYSCTL_DIR${P_GREEN} directory${P_NC}"
+		rm "$P_SYSCTL_DIR/01-lkrg.conf"
+	elif [ -e "$P_SYSCTL_DIR/01-lkrg.conf" ]; then
+		echo -e "       ${P_YL}$P_SYSCTL_DIR/01-lkrg.conf${P_GREEN} was modified, preserving it as ${P_YL}$P_SYSCTL_DIR/01-lkrg.conf.saved${P_NC}"
 		echo -e "       ${P_GREEN}If you do not need it anymore, delete it manually${P_NC}"
-		mv "$P_SYSCTL_DIR/lkrg.conf"{,.saved}
+		mv "$P_SYSCTL_DIR/01-lkrg.conf"{,.saved}
 	fi
 else
 	echo -e "      ${P_RED}ERROR! Unknown option!${P_NC}"


solardiz (changes requested):
This PR makes `scripts/bootup/systemd/lkrg.service` consistent with `debian/rules`, but inconsistent with `scripts/bootup/systemd/lkrg-systemd.sh`. IOW, it fixes one problem, but introduces another.

@adrelanos @morfikov Please propose a fix that doesn't introduce new problems and that both of you are comfortable with. Thanks!

morfikov:
I'm in favor of removing the systemd service altogether as it's useless in the case of kernel module loading. Removing this service and relaying on the native sysctl/modprobe mechanisms would solve all the problems.

solardiz:
> native sysctl/modprobe mechanisms would solve all the problems.

... but would re-introduce the maybe-problem of having the module loaded on bootup after package installation, without needing to enable it explicitly after testing.

solardiz:
@morfikov Please see also discussion in #124. Things don't look that simple to me.

morfikov:
> > native sysctl/modprobe mechanisms would solve all the problems.
> 
> ... but would re-introduce the maybe-problem of having the module loaded on bootup after package installation, without needing to enable it explicitly after testing.

Could you put this in some other words, because I can't really get what you mean by that. :)

adrelanos:
> ... but would re-introduce the maybe-problem of having the module loaded on bootup after package installation, without needing to enable it explicitly after testing.

On Debian that is distribution culture. Any installed package with enable the service/daemon/kernel module by default. For example, `sudo apt install nginx` and `nginx` will be already running.

May I suggest to do it as other kernel modules are usually doing it for the specific linux distributions?

solardiz:
> > the maybe-problem of having the module loaded on bootup after package installation, without needing to enable it explicitly after testing.
> 
> Could you put this in some other words, because I can't really get what you mean by that. :)

As I recall, in #74 part of the criticism was that it was too easy to make a system unbootable by installing LKRG if the version/build of LKRG and the kernel turn out to be incompatible in a bad way. We recommend testing of LKRG before installing it, but as a further safeguard we now also don't enable LKRG by default, so that if after installation the person starts the LKRG service (but does not enable it yet) and the system locks up, a reboot is sufficient to get the system working again.

Anyway, maybe we can accept reintroducing this maybe-problem once we've taken care of #125.

solardiz:
> On Debian that is distribution culture. Any installed package with enable the service/daemon/kernel module by default. For example, `sudo apt install nginx` and `nginx` will be already running.
> 
> May I suggest to do it as other kernel modules are usually doing it for the specific linux distributions?

OK. So apparently this will differ between e.g. Debian and Alpine.

Anyway, as it relates to this one PR, @adrelanos will you revise it in some way (how?) or do we just close it? We shouldn't merge it as-is because it breaks things for non-Debian and even for Debian non-packaged installs.

morfikov:
> > > the maybe-problem of having the module loaded on bootup after package installation, without needing to enable it explicitly after testing.
> > 
> > 
> > Could you put this in some other words, because I can't really get what you mean by that. :)
> 
> As I recall, in #74 part of the criticism was that it was too easy to make a system unbootable by installing LKRG if the version/build of LKRG and the kernel turn out to be incompatible in a bad way. We recommend testing of LKRG before installing it, but as a further safeguard we now also don't enable LKRG by default, so that if after installation the person starts the LKRG service (but does not enable it yet) and the system locks up, a reboot is sufficient to get the system working again.
> 
> Anyway, maybe we can accept reintroducing this maybe-problem once we've taken care of #125.

Oh, I see now. My response to this is simple. There are many things, which can break a system. In all of those cases you have at least three options in order to recover:
- use `init=/bin/bash`  (supported by all)
- use `break=premount` (supported by all)
- use `systemd.unit=emergency.target` (this one is systemd only)

So you have at least 3 kernel cmdline parameters, which can help you with system that doesn't boot for any reason, including loading of the faulty lkrg module. Should you really introduce another parameter? 

With the `modprobe` files, to recover from a broken system is simple as using one of the parameters listed above and commenting the lkrg module out in the right file. It's easy, simple, fast and really straightforward way to follow.

solardiz:
> Should you really introduce another parameter?

I think yes. None of those you listed instantly result in a fully usable system - they just provide one with a way to manually disable LKRG and reboot again.

Besides, the kernel itself has some `no*` parameters, we now support linking LKRG into the kernel, and there could be many reasons to (temporarily) disable LKRG (not only in case LKRG misbehaves, but also e.g. to test an exploit) while booting up with an LKRG-enabled kernel image.

> commenting the lkrg module out in the right file.

Unfortunately, I expect that some people trying LKRG out (especially via a distro package) would have difficulty doing that for the very first time, especially while booted up in one of those ways. It's much more familiar for them to have the system boot up fully into their configured desktop environment or whatever, and to simply uninstall LKRG from there (uninstall the package or use our `make uninstall`).

morfikov:
> > Should you really introduce another parameter?
> 
> I think yes. None of those you listed instantly result in a fully usable system - they just provide one with a way to manually disable LKRG and reboot again.

You don't have to reboot again after disabling the the module, you can continue booting at least with `break=premount` and `systemd.unit=emergency.target` -- you just comment the module out in the right file, and `exit` , and the boot will continue as usual, but without loading of the module, which fixes the system for good, i.e. till you decide to enable the module again.

> 
> > commenting the lkrg module out in the right file.
> 
> Unfortunately, I expect that some people trying LKRG out (especially via a distro package) would have difficulty doing that for the very first time, especially while booted up in one of those ways. It's much more familiar for them to have the system boot up fully into their configured desktop environment or whatever, and to simply uninstall LKRG from there (uninstall the package or use our `make uninstall`).

I'm not really against the parameter, but it shouldn't affect the idea of removing the systemd service. You don't have to explicitly support this init system in any other way than you would with sysvinit or openrc. That's why I've always been for removing the service as it was always useless and introduced really weird problems.

adrelanos:
> > Should you really introduce another parameter?
> 
> I think yes. None of those you listed instantly result in a fully usable system - they just provide one with a way to manually disable LKRG and reboot again.
> 
> Besides, the kernel itself has some `no*` parameters, we now support linking LKRG into the kernel, and there could be many reasons to (temporarily) disable LKRG (not only in case LKRG misbehaves, but also e.g. to test an exploit) while booting up with an LKRG-enabled kernel image.

https://github.com/openwall/lkrg/issues/125

adrelanos:
> Anyway, as it relates to this one PR, @adrelanos will you revise it in some way (how?) or do we just close it?

Renaming the file makes sense anyhow. The prefix `30-lkrg` seems right for a `d.` directory, i.e. for `/etc/sysctl.d/`.

I suggest to rename `/etc/sysctl.d/lkrg.conf` to `/etc/sysctl.d/30-lkrg-dkms.conf`.

Sounds good?

Should be easy, just string search replace in all files?

solardiz:
> I suggest to rename `/etc/sysctl.d/lkrg.conf` to `/etc/sysctl.d/30-lkrg-dkms.conf`.
> 
> Sounds good?

Can we do `/etc/sysctl.d/30-lkrg.conf` (without `-dkms` in there)? LKRG can be installed in a variety of ways, not always with DKMS, so it's wrong to have that in the generic filename.

solardiz:
> Should be easy, just string search replace in all files?

Yes, although then new version's `make uninstall` wouldn't remove `/etc/sysctl.d/lkrg.conf` that might have been created by a previous `make install`. I guess we'd accept that, expecting people to use `make uninstall` from whatever version was installed in order to have it uninstalled reliably and cleanly.

Why `30`, though? Don't we want LKRG configuration applied at almost the earliest time possible, so maybe `00` or `01`? (Assuming that these are loaded in alphabetically sorted order. I haven't checked whether that is actually guaranteed or not - the `sysctl` man page here leaves that unspecified.)

morfikov:
I think 30 is just fine.

> Why `30`, though? Don't we want LKRG configuration applied at almost the earliest time possible, so maybe `00` or `01`? (Assuming that these are loaded in alphabetically sorted order. I haven't checked whether that is actually guaranteed or not - the `sysctl` man page here leaves that unspecified.)

There's info in man sysctl.d :

> 
>        All configuration files are sorted by their filename in lexicographic order, regardless of which of the directories they reside in. If
>        multiple files specify the same option, the entry in the file with the lexicographically latest name will take precedence. Thus, the
>        configuration in a certain file may either be replaced completely (by placing a file with the same name in a directory with higher
>        priority), or individual settings might be changed (by specifying additional settings in a file with a different name that is ordered
>        later).

solardiz:
> I think 30 is just fine.

It may be, but if we have to choose a number, we better have reasons. For example, review what files other packages install on typical distros and how LKRG's will fit in with those. Without such review, I think the lowest number makes most sense, so `00`.

adrelanos:
`00` is bad because then other distributions or system administrators have a hard time to preempt it if needed. It can be a low number as this results in a high priority (early processing of settings) but shouldn't be `00` as this would be unnecessarily the highest priority.

Since sysctl is fully parsed anyhow before loading any kernel module (LKRG) (which ever mechanism used, systemd unit or otherwise), LKRG's priority shouldn't matter. Systemd clearly orders sysctl before module load. Anything else would be a bug. That needs to be sequential and not parallelized in the boot order and is indeed. Even `99` would early enough because loading kernel modules before sysctl is fully parsed would be a bug.

There are no internet standards on how numbers in `.d` config folders should be sorted. Only conventions if you will. Best you can do is look at other `.d` folders and then attempt to pick a good number to fit in. To play nice, it is good to allow people to cut the line or to append to the line.

Lexicographic order was always been stable for any `.d` folder I've encountered.

> Can we do /etc/sysctl.d/30-lkrg.conf (without -dkms in there)?

Yes.

solardiz:
> `00` is bad because then other distributions or system administrators have a hard time to preempt it if needed.

I think this reasoning doesn't apply: we're talking variable assignments, so the latest stays in effect, not the earliest.

> Since sysctl is fully parsed anyhow before loading any kernel module (LKRG) (which ever mechanism used, systemd unit or otherwise), LKRG's priority shouldn't matter. Systemd clearly orders sysctl before module load.

Oh. It doesn't make sense to set LKRG sysctls before the LKRG module is loaded. That would simply fail, because the sysctls don't exist yet. So are you saying we have a bug now, and have no way to fix it? Some in what you wrote above should be wrong.

My guess is we actually avoid that problem due to the systemd unit now, and that's a reason to keep the systemd unit and init system awareness in general.

> Anything else would be a bug.

Why? I don't see it that way. Arguably, it's the other way around.

morfikov:
The sysctl settings are applied after all modules are loaded. 

In systemd, `systemd-modules-load.service` is loaded before `systemd-sysctl.service` , which makes sense because you first load a module, and then you apply settings to it. If some module was missing and you tried to apply settings, you would get `systemd-sysctl[1252]: Couldn't write '0' to 'some_example_file', ignoring: No such file or directory` error.

When you put the lkrg module in `/etc/modules-load.d/` , then it will be loaded by `systemd-modules-load.service` and no extra services are needed to load the module. The settings for the module will be applied shortly after the `systemd-modules-load.service` is loaded by `systemd-sysctl.service` , which is run just right after it.

So the sysctl/modprobe is fully supported by systemd, and also by openrc and sysvinit. There's really no need to explicitly support systemd by making a separate unit.

solardiz:
OK, so I assume that @morfikov's description of the order in which services are started is correct, and @adrelanos is not.

For the filename, I propose `01-lkrg.conf`. While I think 00 would also be reasonable, 01 leaves more of a possibility for another package to provide LKRG(-related) settings that would only take effect if not overridden by LKRG package's file (and other files at 01 and higher). I doubt that will ever be needed, but why not.

@adrelanos Would you like to revise this PR accordingly, or should I take care of this change separately from this PR?

morfikov:
Can be `01-` . But also change the number in [debian/rules](https://github.com/openwall/lkrg/blob/main/debian/rules#L20) to be in sync.

Adam-pi3:
Hm... thanks for this educational discussion. What I'm reading here there is a race window after LKRG is loaded and before configuration is applied. If user decides to not enforce kernel integrity feature (e.g., he only wants to have process integrity) then in this short race window, his configuration won't be taken into account. In that timeframe, default LKRG's internal sysctl configuration would be taken into account. @solardiz is such scenario acceptable?

solardiz:
@Adam-pi3 This time window of LKRG running with default configuration is implied when (mis)using sysctl's to provide initial configuration. We actually intended the module parameters for initial configuration, and sysctl's for subsequent changes. However, we accepted such (mis)use of sysctl's when we accepted prior PRs with initial `/etc/sysctl.d/lkrg.conf`.

For cases when running with default configuration even for a moment is highly undesirable, there should be a mechanism to provide module parameters as well. I think that mechanism is the `options` line in a file under `/etc/modprobe.d`. Maybe we need to separately (not in this PR) provide a sample file in there as well, with something like `#options p_lkrg some-examples-here` in it, commented out with the `#`.

Or maybe (also not in this PR) we need to reconsider and stop providing an `/etc/sysctl.d` file in LKRG.

solardiz:
I took the liberty to replace the contents of this PR with what I think we agreed on above. Can each one of you please take a look and confirm? Thanks!

solardiz:
> Can be `01-` . But also change the number in [debian/rules](https://github.com/openwall/lkrg/blob/main/debian/rules#L20) to be in sync.

I updated `debian/rules`, but didn't bother adding to `debian/changelog` - I assume you can take care of that separately if you like.

solardiz (approved):

wladmis:
Some says that it is bad idea to install files to /etc/sysctl.d dir. The idea is that it is the place is destined for admins to edit files, meanwhile software should install its configurations to system directories, to /lib/sysctl.d dir in that particular case.

adrelanos:
> @wladmis I guess in our case `/etc` is more appropriate than `/lib` because we install the file for possible local edits to it, not to provide package-imposed settings.

Local edits by system administrator should happen in /etc. But indeed /lib is better for packages to place files in. /lib is the new/proper way. /etc is legacy only if not possible/available in the interface used.

(Interfaces would be module load, sysctl, systemd, grub.)

That local system administrator would edit the file does not mean /etc should be preferred.

scripts/bootup/systemd/lkrg.service
```diff
@@ -18,7 +18,7 @@ ConditionKernelCommandLine=!nolkrg
 [Service]
 Type=oneshot
 ExecStart=/sbin/modprobe -v p_lkrg
-ExecStartPost=/sbin/sysctl -p /etc/sysctl.d/lkrg.conf
+ExecStartPost=/sbin/sysctl -p /etc/sysctl.d/01-lkrg.conf
```
adrelanos, line 21: This `ExecStartPost=/sbin/sysctl -p` seems weird to me to begin with. Violating principle of least surprise. Haven't ever seen anything similar. It might result in loading that config snippet but ignoring any others the user might have set.

Not a blocker. As said elsewhere, I am not a big fan of a systemd unit is this very particular vase anyhow.

solardiz: I think it's not this `ExecStartPost=/sbin/sysctl -p /etc/sysctl.d/01-lkrg.conf` that violates the principle of least surprise, but rather any kind of use of sysctl to provide initial LKRG configuration might be confusing. When we do go for such use, we actually have to have that `ExecStartPost` in order to reduce the surprise - LKRG settings that were applied on system bootup should not be lost on LKRG service restart (such as after LKRG is upgraded but prior to a reboot).

> It might result in loading that config snippet but ignoring any others the user might have set.

I don't know what you mean, but maybe you misunderstand how sysctl works. The rest of the previously set sysctl settings remain intact (in kernel variables). Sure, if there are other LKRG settings (not other kernel sysctls in general) set via sysctl elsewhere, they would be lost on LKRG service restart (module removal and re-insertion), but what can we do, or what are we supposed to do? I think there's no better action we can take here.

The fix would be stopping to use sysctl to provide initial LKRG configuration.

Adam-pi3:
> Local edits by system administrator should happen in /etc. But indeed /lib is better for packages to place files in. /lib is the new/proper way. /etc is legacy only if not possible/available in the interface used.

Just minor comment about the `/lib` is that some distros are using `/usr/lib` and we would need to have a reliable way of finding the right path. Seems like this problem won't exist if we use `/etc`

solardiz:
@morfikov

> Since it's not a real debian package (in the repo), the changelog doesn't have to be updated often. [...] But if the new version showed up, the info about renaming the conffile should be added.

While I wish you took care of that, we're now about to make a release and can't wait on you, so I'm about to add this:

```diff
+++ b/debian/changelog
@@ -1,3 +1,11 @@
+lkrg (0.9.2-1) unstable; urgency=medium
+
+  * New upstream release.
+  * debian/rules:
+    + Install the sysctl settings in /etc/sysctl.d/01-lkrg.conf.
+
+ -- Solar Designer <solar@openwall.com>  Tue, 28 Dec 2021 23:59:59 +0100
+
 lkrg (0.9.1-2) unstable; urgency=medium
 
   * Drop the debian/lkrg-dkms.sysctl file altogether and use the provided
```

Looks right?

solardiz:
There was also this:

```
$ git log control 
commit c54b71546304231ca64f88e2a211ef56bb3babab
Author: Patrick Schleizer <adrelanos@whonix.org>
Date:   Sat Aug 21 12:44:26 2021 -0400

    stop recommending lkrg-systemd
    
    https://github.com/openwall/lkrg/issues/108#issuecomment-890534005
```

but it'd be weird to claim Patrick's change in my name nor to add a `changelog` entry in his name. So I won't bother now, but perhaps the two of you should in fact maintain this `changelog` properly going forward.

morfikov:
> Looks right?

Yes

The updated `debian/changelog` file is needed only for people to build the updated deb package.  So to be able to build the new version of LKRG as the deb package, you just have to bump the version in the file and just add the `* New upstream release.` entry.  That's basically it. 

The `debian/changelog` is for Debian users and changes put in here should concern mostly the internals of Debian (for instance, drop some dependency, switch to some other version of other deps, rename something, etc). LKRG has its own changelog and all changes made to the project itself should be documented in the project's changelog.

morfikov:
Also I've noticed you changed to project URLs (here on GH and the website). You updated the files in the `debian/` dir, but this change for sure should go to the debian's changelog, at least it the package was in the Debian repo, :D

Issue #108, opened 2021-07-25
LKRG Debian package - systemd unit does not automatically start - lkrg.service is a disabled or a static unit, not starting it.


adrelanos:
> sudo apt install lkrg
```
...
Setting up lkrg (0.9.1-3) ...
Setting up lkrg-systemd (0.9.1-3) ...
lkrg.service is a disabled or a static unit, not starting it.
```

It requires a manual `sudo systemctl enable lkrg` and reboot or `sudo systemctl start lkrg`.

I am wondering, why does it do that?

The Debian default is to enable systemd units after installation.

adrelanos:
https://github.com/Whonix/lkrg/blob/old-master/debian/lkrg-dkms.service did not have this issue.

Some change in https://github.com/openwall/lkrg/blob/main/scripts/bootup/systemd/lkrg.service might be causing this issue.

old:
`WantedBy=multi-user.target`

new:
`WantedBy=sysinit.target`

Wondering if this is causing it.

* [When should a systemd unit file use sysinit.target vs multi-user.target?](https://unix.stackexchange.com/questions/556923/when-should-a-systemd-unit-file-use-sysinit-target-vs-multi-user-target)
* https://www.freedesktop.org/software/systemd/man/bootup.html

In my understanding, `WantedBy=multi-user.target` does not invalidate `Before=sysinit.target`.

adrelanos:
Even replacing the systemd unit https://github.com/Whonix/lkrg/blob/old-master/debian/lkrg-dkms.service does not fix this. So I guess it must be something else in the Debian packaging.

adrelanos:
For debugging maybe [`debian/lkrg-systemd.postinst`](https://github.com/Whonix/lkrg/blob/debian-packaging-systemd-debugging/debian/lkrg-systemd.postinst) can help.

(Not suggested for merge. Only manual use for debugging.)

This is the debug output:

```
sudo dpkg -i lkrg-systemd_0.9.1-3_all.deb 
```

```
(Reading database ... 328917 files and directories currently installed.)
Preparing to unpack .../lkrg-systemd_0.9.1-3_all.deb ...
Unpacking lkrg-systemd (0.9.1-3) over (0.9.1-3) ...
Setting up lkrg-systemd (0.9.1-3) ...
+ set -e
+ true '
#####################################################################
## INFO: BEGIN: lkrg-systemd postinst configure' '0.9.1-3
#####################################################################
'
[...]

PR #114, opened 2021-08-21
stop recommending lkrg-systemd

adrelanos:
https://github.com/openwall/lkrg/issues/108#issuecomment-890534005

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -14,7 +14,6 @@ Package: lkrg
 Architecture: all
 Depends: ${misc:Depends},
  lkrg-dkms (= ${source:Version})
-Recommends: lkrg-systemd (= ${source:Version})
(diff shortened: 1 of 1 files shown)


solardiz:
Re-reading the comments in #108, @morfikov seemed OK with this change. I don't have an opinion on it, but since both @adrelanos and @morfikov are OK with it and they're the ones maintaining the Debian packaging so far, I'll proceed to merge this.

solardiz:
For future occasions, please prefix commit titles with subsystem name where relevant. In this case, I'd use "debian/control: " or "debian: " or "Debian packaging: ". Also, instead of referring to GitHub issue comments (which might not stay around), include the rationale in the commit messages themselves (if and when desired).

PR #124, opened 2021-11-14
disable systemd scripts on systems without systemd

itoffshore:
* small `Makefile` change to disable the boot scripts on systems without `systemd`

* tested on Alpine Linux as part of the [new aport for `lkrg`](https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/27429)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,10 @@ P_OUTPUT = output
 P_PWD ?= $(shell pwd)
 P_KVER ?= $(shell uname -r)
 P_BOOTUP_SCRIPT ?= scripts/bootup/lkrg-bootup.sh
+P_SYSTEMD = /lib/systemd/systemd
+P_SYS_ERR = 'skipping systemd boot scripts'
 TARGET := p_lkrg
+
 ifneq ($(KERNELRELEASE),)
     KERNEL := /lib/modules/$(KERNELRELEASE)/build
 else
@@ -101,10 +104,10 @@ all:
 install:
 	$(MAKE) -C $(KERNEL) M=$(P_PWD) modules_install
 	depmod -a
-	$(P_PWD)/$(P_BOOTUP_SCRIPT) install
+	[ -f $(P_SYSTEMD) ] && $(P_PWD)/$(P_BOOTUP_SCRIPT) install || echo $(P_SYS_ERR)
 
 uninstall:
-	$(P_PWD)/$(P_BOOTUP_SCRIPT) uninstall
+	[ -f $(P_SYSTEMD) ] && $(P_PWD)/$(P_BOOTUP_SCRIPT) uninstall || echo $(P_SYS_ERR)
 
 clean:
 	$(MAKE) -C $(KERNEL) M=$(P_PWD) clean


solardiz:
@itoffshore Thanks, but this looks wrong to me, in several ways:

1. While `scripts/bootup/lkrg-bootup.sh` currently only supports systemd, it is a wrapper script meant to potentially invoke future scripts supporting other init systems. "Teaching" the top-level `Makefile` about systemd breaks this abstraction layer.
2. Why is this PR's change even needed? `scripts/bootup/lkrg-bootup.sh` looks like it'd already detect the lack of systemd on a system and print a message and terminate without a non-zero exit code, so `make` should be happy. Am I missing something?
3. If this PR is needed, then it should contain just one commit, no extra `Merge pull request #1 from itoffshore/alpine`.

itoffshore:
I see `lkrg-bootup.sh` should have a `case` added for OpenRC so closing.

solardiz:
> I see `lkrg-bootup.sh` should have a `case` added for OpenRC so closing.

That was the original plan for that script, but there are suggestions e.g. in #108 that we should actually drop the systemd unit, and if we do then perhaps we wouldn't support any init systems - instead relying on `/etc/modules-load.d`. Is the latter also supported by systems that use OpenRC?

I don't mind us adding OpenRC support though, even if it might end up being temporary.

Another open question is whether installing LKRG should also automatically enable its start on system bootup or not. We used to have that for systemd, but we dropped this in 3c2b7b2e20b2891a86eba35c6fd8a76461c1b505 so that if testing of LKRG locks the system up that issue does not persist after a reboot. Someone installing LKRG manually is supposed to test it first, and only enable its automatic startup upon successful testing. So if you do add OpenRC support, can you do the same - have the service installed, but not yet enabled?

itoffshore:
Alpine has `/etc/modules-load.d` & `/etc/modprobe.d`

If `lkrg` already uses sane defaults with zero configuration `/etc/modules-load.d/lkrg.conf` is enough. Fine tuning is then done by the user with an `/etc/sysctl.d` file.

This will work for every system & removes the maintenance of various boot scripts & unexpected behaviour. For a kernel module service scripts should not be necessary.

solardiz:
> For a kernel module service scripts should not be necessary.

Right, however please see this comment I just posted https://github.com/openwall/lkrg/issues/108#issuecomment-968357020 and in particular:

If we start using `/etc/modules-load.d`, then we'd effectively be enabling LKRG on bootup by default, thus reintroducing the problem/criticism mentioned in #74.

itoffshore:
enabling the module is probably best left to the user or the Package Maintainer so any show stopping issues are identified in testing.

on Alpine there are no packages in `main` that [enable modules](https://pkgs.alpinelinux.org/contents?file=&path=%2Fetc%2Fmodules-load.d&name=&branch=edge)

I think a good compromise would be an info message of how to enable the `p_lkrg` module - I'm sure a lot of people will in error try `modprobe lkrg` (I know I did when I first started using `lkrg`)

solardiz:
> enabling the module is probably best left to the user or the Package Maintainer so any show stopping issues are identified in testing.

How do you suggest we achieve that if we use `/etc/modules-load.d`?

> on Alpine there are no packages in `main` that [enable modules](https://pkgs.alpinelinux.org/contents?file=&path=%2Fetc%2Fmodules-load.d&name=&branch=edge)

Do I get it right that the packages in `main` that include kernel modules don't have corresponding `/etc/modules-load.d` lines?

> I think a good compromise would be an info message of how to enable the `p_lkrg` module

That's what we have for systemd now. We print a message on usage of `systemctl`.

> - I'm sure a lot of people will in error try `modprobe lkrg` (I know I did when I first started using `lkrg`)

Do you mean missing the `p_` prefix?

> being able to enable or disable `lkrg` on the kernel boot line would also be useful in cases of an unbootable system

We have the `nolkrg` option processed by the systemd unit now. Perhaps we need to introduce it into LKRG itself, so that it'd also work with other boot systems.

itoffshore:
In Alpine the policy is not to have unexpected behaviour so very little configuration is done for you. Services are not enabled when you install a package for instance (unlike in Debian)

* Creating `/etc/modules-load.d/lkrg.conf.disabled` containing `p_lkrg` is another idea (in Alpine this is done with `/etc/profile.d/color_prompt.disabled` for example )

This will help stop any confusion with the slightly different module name to `modprobe` (`p_lkrg` being required to modprobe not `lkrg`) 

I think an info message is more appropriate than creating a file. Then it is the responsibility of the user or Package Maintainer to test & enable the module.

A module parameter that can be set or overridden on the boot command line to enable / disable `lkrg` solves the broken system issues (I did get caught once with an unbootable system)

solardiz:
> This will help stop any confusion with the slightly different module name to `modprobe` (`p_lkrg` being required to modprobe not `lkrg`)

I think the best fix for that confusion would be to rename the module from `p_lkrg` to `lkrg`, if @Adam-pi3 agrees.

PR #153, opened 2022-01-30
Fix build on opensuse/leap

vt-alt:
SUSE version `#if`s added to a set in a similar way as RHEL.

`CONFIG_SUSE_PRODUCT_CODE` `1` seems to be for Leap, `3` for Tumbleweed.


vt-alt:
Force pushed to add `#` to the `Fixes` tag.

solardiz:
Thank you, @vt-alt! And I agree with your reasoning that we need to reconsider our approach to these.

Issue #137, opened 2022-01-13
Intermittent "_STEXT MEMORY BLOCK HASH IS DIFFERENT" on Linux 5.16.0-051600daily20220113-generic


solardiz:
As seen in https://github.com/solardiz/lkrg/runs/4801643170 for LKRG effectively at d75c7a48c3097123c760579eff5f0d94b8077c8b (December 9, 2021, after which we made no relevant fixes yet, so the issue probably applies to our latest as well):

```
[   28.858645] systemd[1]: Mounting Huge Pages File System...
         Mounting Huge Pages File System...
[   28.880341] systemd[1]: Mounting POSIX Message Queue File System...
         Mounting POSIX Message Queue File System...
[   28.900705] systemd[1]: Mounting Kernel Debug File System...
         Mounting Kernel Debug File System...
[   28.937468] systemd[1]: Mounting Kernel Trace File System...
         Mounting Kernel Trace File System...
[   28.951572] systemd[1]: Stopped Journal Service.
[  OK  ] Stopped Journal Service.
[   28.985877] [p_lkrg] ALERT !!! _STEXT MEMORY BLOCK HASH IS DIFFERENT - it is [0x51240158d2cff830] and should be [0x3c94ef031bb8b45c] !!!
[   29.026047] [p_lkrg] ALERT !!! SYSTEM HAS BEEN COMPROMISED - DETECTED DIFFERENT 1 CHECKSUMS !!!
[   29.026047] Kernel panic - not syncing: [p_lkrg] Kernel Integrity verification failed! Killing the kernel...
[   29.026047] CPU: 1 PID: 292 Comm: kworker/u4:4 Tainted: G           OE     5.16.0-051600daily20220113-generic #202201122112
[   29.026047] Hardware name: QEMU Standard PC (Q35 + ICH9, 2009), BIOS 1.13.0-1ubuntu1.1 04/01/2014
[   29.026047] Workqueue: events_unbound p_check_integrity [p_lkrg]
[   29.026047] Call Trace:
[   29.026047]  <TASK>
[   29.026047]  show_stack+0x52/0x58
[   29.026047]  dump_stack_lvl+0x4c/0x63
[   29.026047]  dump_stack+0x10/0x12
[   29.026047]  panic+0x109/0x2e1
[   29.026047]  p_check_integrity.cold+0xa9b/0x1b0f [p_lkrg]
[   29.026047]  process_one_work+0x21f/0x3f0
[...]

PR #154, opened 2022-01-31
CI: Enable P_LKRG_JUMP_LABEL_STEXT_DEBUG in mainline test

vt-alt:
As discussed in https://github.com/lkrg-org/lkrg/issues/137

>  It might be useful to always enable P_LKRG_JUMP_LABEL_STEXT_DEBUG on mainline build.

### How Has This Been Tested?
https://github.com/vt-alt/lkrg/runs/5013353445

### DIFF
--- a/.github/workflows/mkosi-mainline.yml
+++ b/.github/workflows/mkosi-mainline.yml
@@ -25,6 +25,11 @@ jobs:
             - name: Download mainline kernel from Kernel PPA
               run: .github/workflows/ubuntu-kernel-daily.sh
 
+            - name: Enable LKRG debugging options
+              run: |
+                  sed -i '/P_LKRG_JUMP_LABEL_STEXT_DEBUG/s/\/\///' src/modules/print_log/p_lkrg_log_level_shared.h
+                  git diff
+
             - name: Create bootable image using mkosi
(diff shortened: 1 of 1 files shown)


Adam-pi3:
It's unrelated to this PR, but it looks like there is problem with docker build for centos:8
```
Run yum install -y kernel-devel kernel gcc make elfutils-libelf-devel
CentOS Linux 8 - AppStream                      178  B/s |  38  B     00:00    
Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
Error: Process completed with exit code 1.
```

vt-alt:
> It's unrelated to this PR, but it looks like there is problem with docker build for centos:8

Seems accidental error and no reproducible. Maybe it was some temporary network issue.

solardiz:
The CentOS 8 build keeps failing like that. It failed after this PR was merged, then failed two more times as I tried to re-run the job.

vt-alt:
Ah, I mis-thought it's centos stream. Yes, plain centos 8 is failing and other projects are removing it, such as LTP https://lists.linux.it/pipermail/ltp/2022-February/027417.html
Perhaps, we should do the same.

Issue #158, opened 2022-02-04
Built-in on GCC 11.1 leaving orphan sections


sempervictus:
When building in the kernel tree, not as a module, on GCC 11.1.0, this warning is thrown at the end of the compilation process:
```
  LD      .tmp_vmlinux.kallsyms1'
ld: warning: orphan section `.p_lkrg_read_only' from `security/lkrg/p_lkrg_main.o' being placed in section `.p_lkrg_read_only'
  KSYMS   .tmp_vmlinux.kallsyms1.S
```
The tree in which this is happening is @ 5.10.96 with the `linux-hardened` patches (not adding major GCC plugins or anything like that).

sempervictus:
Hmm, something's not right here. On the subsequent 5.10.100, built as a module, and loaded at runtime, i get this fine mess:
```
# modprobe p_lkrg 
modprobe: ERROR: could not insert 'p_lkrg': No buffer space available

```
and in dmesg:
```
[ 6474.831003] [p_lkrg] Loading LKRG...
[ 6474.831620] [p_lkrg] System does NOT support SMEP. LKRG can't enforce SMEP validation :(
[ 6474.832685] [p_lkrg] System does NOT support SMAP. LKRG can't enforce SMAP validation :(
[ 6474.836621] Freezing user space processes ... (elapsed 0.001 seconds) done.
[ 6474.837918] OOM killer disabled.
[ 6474.850588] [p_lkrg] [ED] ERROR: Can't find 'put_seccomp_filter' function :( Exiting...
[ 6474.851636] [p_lkrg] Can't initialize exploit detection features! Exiting...
[ 6474.916225] OOM killer enabled.
[ 6474.916226] Restarting tasks ... done.
# uname -r
5.10.100
```

Adam-pi3:
I don't see anything wrong with this:
```
ld: warning: orphan section `.p_lkrg_read_only' from `security/lkrg/p_lkrg_main.o' being placed in section `.p_lkrg_read_only'
```
LKRG does created a new section called `.p_lkrg_read_only` which probably matches the attribute of the kernel's `.rodata` section and this warning is printed.

The problem with initialization is related to:
```
[p_lkrg] [ED] ERROR: Can't find 'put_seccomp_filter' function :( Exiting...
```
which is exactly the same as https://github.com/lkrg-org/lkrg/issues/135. Please read the fix for that issue in my comment:
[...]

Issue #135, opened 2021-12-31
[5.15.12] load error


Pheidologeton:
The following error occurs when trying to load a kernel module.
modprobe
~~~
sudo modprobe p_lkrg
modprobe: ERROR: could not insert 'p_lkrg': No buffer space available
~~~
dmesg
~~~
sudo dmesg -W
[46512.556753] [p_lkrg] Loading LKRG...
[46512.556764] [p_lkrg] System does NOT support SMAP. LKRG can't enforce SMAP validation :(
[46512.561193] Freezing user space processes ... (elapsed 0.002 seconds) done.
[46512.563858] OOM killer disabled.
[46512.583300] [p_lkrg] [ED] ERROR: Can't find 'put_seccomp_filter' function :( Exiting...
[46512.583302] [p_lkrg] Can't initialize exploit detection features! Exiting...
[46512.630179] OOM killer enabled.
[46512.630182] Restarting tasks ... done.
~~~
Arch linux, latest zen kernel.

solardiz:
@Pheidologeton Thank you for reporting this. Can you please show which `CONFIG_SECCOMP*` options your kernel has? My guess is it somehow has `CONFIG_SECCOMP`, but not `CONFIG_SECCOMP_FILTER`. Maybe we need to add support for that. My other guess is you're inserting LKRG into a kernel configured differently from the kernel for which LKRG was built. Let's figure this out.

Pheidologeton:
My CONFIG_SECCOMP
~~~
zcat /proc/config.gz | grep "CONFIG_SECCOMP"
CONFIG_SECCOMP=y
CONFIG_SECCOMP_FILTER=y
# CONFIG_SECCOMP_CACHE_DEBUG is not set
~~~

solardiz:
I'm not familiar with Arch and don't know what "latest zen kernel" is, but perhaps it's not the same kernel that "LKRG 0.9.2-dkms from AUR" was built against? And, what kernel _is_ the LKRG package built against?

I can see how we'd have this problem if the LKRG package is built against a kernel older than 5.9, but I doubt that's the case?

@rgacogne @itoffshore Can you comment?

Pheidologeton:
This is a DKMS package which is built from source code for the installed kernel. LKRG is not provided in binary.
My kernel
~~~
uname -r
5.15.12-zen1-1-zen
~~~

itoffshore:
I have the same `SECCOMP` config on `linux-hardened`:

```
[...]

PR #161, opened 2022-02-17
Add kernel diff (5.10) to export seccomp symbols

sempervictus:
Custom kernels compiled with -O3 (or other optimization) appear to
break the runtime symbol resolution in LKRG, as per Adam's comment:
github.com/lkrg-org/lkrg/issues/135#issuecomment-1018693257

In order to maintain access to these symbols in the optimized O3
binary, explicit export annotations are required in the kernel at
build-time.

Implemented exports for do_seccomp, change_page_attr_set_clr, and
do_seccomp on 5.10.98-hardened. Extracted the resulting diff into
the scripts/builtin directory of this repo for application by users
who patch their own kernels already. This won't help people pulling
pre-built binaries out of repositories, but AUR users and similar
compile-on-demand setups should be able to add the diff to their
PKGBUILD/analogous build blueprint.

Testing:
  Arch Linux VM on 5.10.101 loading p_lkrg as a module succeeds.

### Description
Related to #158 and #135

### How Has This Been Tested?
Booted Arch Linux on 5.10.101 with this patch applied to the kernel sources and loaded the LKRG kernel module successfully:
```
[   49.980824] [p_lkrg] Loading LKRG...
[   64.531548] cfg80211: failed to load regulatory.db
[   64.532098] Freezing user space processes ... (elapsed 0.001 seconds) done.
[   64.533364] OOM killer disabled.
[   64.540020] [p_lkrg] LKRG can't enforce SELinux validation (CONFIG_GCC_PLUGIN_RANDSTRUCT detected)
[   64.755986] [p_lkrg] [kretprobe] register_kretprobe() for <ovl_create_or_link> failed! [err=-2]
[   64.756029] [p_lkrg] Can't hook 'ovl_create_or_link' function. This is expected if you are not using OverlayFS.
[   64.859440] [p_lkrg] [kretprobe] register_kretprobe() for <lookup_fast> failed! [err=-2]
[   64.859487] [p_lkrg] LKRG won't enforce pCFI validation on 'lookup_fast'
[   65.046324] [p_lkrg] LKRG initialized successfully!
[   65.046347] OOM killer enabled.
[   65.046348] Restarting tasks ... done.
```

### DIFF
--- a/scripts/builtin/export_symbols_from_optimized_kernels.diff
+++ b/scripts/builtin/export_symbols_from_optimized_kernels.diff
@@ -1,0 +1,32 @@
+diff --git a/arch/x86/mm/pat/set_memory.c b/arch/x86/mm/pat/set_memory.c
+index 40baa90e74f4..c2cf7cea5846 100644
+--- a/arch/x86/mm/pat/set_memory.c
++++ b/arch/x86/mm/pat/set_memory.c
+@@ -1775,6 +1775,7 @@ static int change_page_attr_set_clr(unsigned long *addr, int numpages,
+ out:
+        return ret;
+ }
++EXPORT_SYMBOL(change_page_attr_set_clr);
+ 
+ static inline int change_page_attr_set(unsigned long *addr, int numpages,
+                                       pgprot_t mask, int array)
+diff --git a/kernel/seccomp.c b/kernel/seccomp.c
+index 305f0eca163e..fa0d391aef04 100644
+--- a/kernel/seccomp.c
++++ b/kernel/seccomp.c
+@@ -448,6 +448,7 @@ static void __put_seccomp_filter(struct seccomp_filter *orig)
+                seccomp_filter_free(freeme);
+        }
+ }
++EXPORT_SYMBOL(__put_seccomp_filter);
+ 
+ static void __seccomp_filter_release(struct seccomp_filter *orig)
+ {
+@@ -1705,6 +1706,8 @@ static long do_seccomp(unsigned int op, unsigned int flags,
+        }
+ }
+ 
++EXPORT_SYMBOL(do_seccomp);
++
+ SYSCALL_DEFINE3(seccomp, unsigned int, op, unsigned int, flags,
+


sempervictus:
@solardiz/@Adam-pi3 - is `[kretprobe] register_kretprobe() for <lookup_fast> failed! [err=-2]` a concern here? Is that something else that needs to be exported?

solardiz:
I doubt the problem is isolated to `-O3`, so this shouldn't be part of the filename.

> is `[kretprobe] register_kretprobe() for <lookup_fast> failed! [err=-2]` a concern here? Is that something else that needs to be exported?

A relatively minor concern, but yes.

sempervictus:
@solardiz - for my own edification, could you possibly elaborate on what else could be causing this? Looks like all of the folks reporting issues have some version of the GCC specialization patch (targets the compiler to specific CPUs) and i'm pretty sure those performance patch sets carry the O3 patch as well. My understanding of that patch is that its supposed to instruct the compiler that it is safe to use certain newer instructions in lieu of complex instruction chains for optimization as opposed to inlining things or changing call structure.
Will rename the file of course, just curious about what's happening under the hood.

sempervictus:
Do we want to put something in the readme to help people find it? Should be good to merge though in terms of functionality

solardiz:
> Do we want to put something in the readme to help people find it?

I think we'd first need to find a more reasonable place in the directory tree for both the `copy-builtin.sh` script and this patch, and then document both.

> Should be good to merge though in terms of functionality

OK, I'll merge this as-is for now. We can revisit this later.

solardiz:
@Adam-pi3 Somehow I don't see your comment here (got it via e-mail) - deleted? Sorry if I was too quick to merge this PR. I didn't want to spend much time on it, nor to discourage @sempervictus. Having PRs just hang around isn't good for our project, but maybe I should have waited a day to see if you have any objections.

I view this patch as obscure documentation/hint. Something we can refer to and polish later. It isn't applied by anything by default, and isn't referred to in any documentation we have - it's just a file in the tree.

solardiz:
Upon further discussion on the commit, it's now un-merged. We should ideally have something cleaner and more useful and more generic than that undocumented and misplaced patch (for which we don't currently have a good place in the tree).

solardiz:
@sempervictus We could have a separate repo for the contributions, but for this specific one I think you can fairly easily polish it enough that Adam would be happy to have it merged - I made specific suggestions. @Adam-pi3 please correct me if I'm wrong.

solardiz:
> I made specific suggestions.

For others reading this later - I referred to comments made on the commit referenced here.

Adam-pi3:
I agree with @solardiz suggestions and I believe that the best approach would be to invoke a separate script from `copy-builtin.sh`. `add-exports.sh` name sounds about right. @sempervictus can you do that? It also requires some testing if everything works fine (e.g., compile the kernel in-tree as well as out-of-tree module).

PR #163, opened 2022-02-17
Add explicit critical function exports to kernel sources via shell script

sempervictus:
### Description
Implements fixes from #161 via `echo` method proposed by @solardiz 

### How Has This Been Tested?
It hasnt, per se, but the original diff from which the lines `echo`d originate works like a charm

### DIFF
--- a/scripts/add-exports.sh
+++ b/scripts/add-exports.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+set -e
+KDIR="${KDIR:="/usr/src/linux"}"
+cd "$KDIR"
+# Add critical exports to prevent them from being optimized-out
+echo "EXPORT_SYMBOL(change_page_attr_set_clr);" >> arch/x86/mm/pat/set_memory.c
+echo "EXPORT_SYMBOL(lookup_fast);" >> fs/namei.c
+# keep symbol export inside the ifdef block defining the symbol
+sed -i '/static void __seccomp_filter_release.*/iEXPORT_SYMBOL(__put_seccomp_filter);\n' kernel/seccomp.c
+echo "EXPORT_SYMBOL(do_seccomp);" >> kernel/seccomp.c


solardiz:
@sempervictus Thanks. As discussed, this should be a separate script `add-exports.sh`, so that it's also usable when intending to build/load LKRG as a module.

Please also take care of `lookup_fast`.

Also, you define `EXPORT_SYMBOLS`, but never use it - either drop it or use it.

And there's that potential issue with the functions being within some `#ifdef` blocks - please check whether this is the case for any, and if so you'll need to be adding `#ifdef ... EXPORT_SYMBOL ... #endif` blocks (e.g. with 3 `echo` lines each or with `cat << EOF`). Hopefully, this is simply not the case, though.

Finally, you'll need to test this before we merge.

solardiz:
@sempervictus Thank you for making some of the changes per my feedback. I think you need to start the new script with:

```bash
#!/bin/bash

set -e

KDIR="${KDIR:="/usr/src/linux"}"
cd "$KDIR"
```

(these are the relevant lines from `copy-builtin.sh`).

Please also take care of `lookup_fast`.

Please look into the potential issue of some of these functions possibly being in `#if` blocks and at least document your findings in here.

Finally, please test this before we merge.

solardiz:
BTW, the commit description says that "Several kernel patches appear to be optimizing-out the symbols", but actually the problem is not limited to kernel patches (unfortunately). In general, while I appreciate the level of detail in your commit descriptions, the more detailed they are, the more likely they are also subtly wrong or subjective. Maybe that's a reason to focus on describing "what" and briefly "why" rather than making an attempt at "exactly why".

solardiz:
> If this happens in unpatched upstream builds too,

So far to a very limited extent - e.g., I've seen this issue for `lookup_fast`, but that's only an optional dependency for LKRG - however, we're concerned the problem may worsen over time.

> then that's a pretty serious issue for LKRG's design

Yes, it is.

> and the whole "resolve symbols at compile-time" thing starts to seem a lot more important

Yes, but this would not address the problem for loading LKRG as a module, so we (also) need some other solution/workaround.

solardiz:
@sempervictus Why did you choose to add only:

```bash
#!/bin/bash
set -e
```

and not also:

```bash
KDIR="${KDIR:="/usr/src/linux"}"
cd "$KDIR"
```

While at it, also drop the trailing empty line.

You're also still to address `lookup_fast`. ;-)

solardiz:
@sempervictus Just like your grep probably already found, `lookup_fast` is in `fs/namei.c`, and no it is not conditional.

You don't need `pushd` / `popd` when the script runs in its own subshell, like it normally will. `cd` only changes the current directory in the subshell, not in the parent shell. Anyway, `pushd` / `popd` would make sense in case the script is ever sourced into another script, rather than invoked from it. The combination of `pushd .` followed by `cd some-path` can be optimized to `pushd some-path`, so please at least do that (or just drop the `pushd` / `popd`).

Oh, I now see you simply copied this weird usage of `pushd` / `popd` from `copy-builtin.sh`. You could want to simplify it there with a separate commit.

sempervictus:
Does the `sed`-based insertion work here or should i do it at the bottom of the file in several lines?

solardiz:
> Does the `sed`-based insertion work here or should i do it at the bottom of the file in several lines?

Either way works, and either can break later. So OK to keep with `sed` if you've tested this.

scripts/add-exports.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+KDIR="${KDIR:="/usr/src/linux"}"
+cd "$KDIR"
+# Add critical exports to prevent them from being optimized-out
+echo "EXPORT_SYMBOL(change_page_attr_set_clr);" >> arch/x86/mm/pat/set_memory.c
+echo "EXPORT_SYMBOL(do_seccomp);" >> kernel/seccomp.c
+echo "EXPORT_SYMBOL(fast_lookup);" >> fs/namei.c
```
Adam-pi3, line 8: it should be `lookup_fast` not `fast_lookup`

scripts/add-exports.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+KDIR="${KDIR:="/usr/src/linux"}"
+cd "$KDIR"
+# Add critical exports to prevent them from being optimized-out
+echo "EXPORT_SYMBOL(change_page_attr_set_clr);" >> arch/x86/mm/pat/set_memory.c
+echo "EXPORT_SYMBOL(do_seccomp);" >> kernel/seccomp.c
```
Adam-pi3, line 7: `do_seccomp` also depends on `CONFIG_SECCOMP`

Adam-pi3:
I don't see changes in `scripts/copy-builtin.sh` file which should invoke this script (`add-exports.sh`). Did I miss it?

scripts/add-exports.sh
solardiz: I think this entire file is not included in build when `CONFIG_SECCOMP=n`, or at least I didn't see an `#if` like that in there. So I think we're fine doing the export "unconditionally". However, I don't mind someone confirming this. ;-)

solardiz:
> I don't see changes in `scripts/copy-builtin.sh` file which should invoke this script (`add-exports.sh`). Did I miss it?

You didn't miss - I think this is not implemented, and I'm fine with that. Not everyone using `copy-builtin.sh` will also need the extra exports.

scripts/add-exports.sh
solardiz: @sempervictus You missed this: "swap lines so that `do_seccomp` goes after it and becomes adjacent to the `sed` of the same file."

(How many GitHub comments does it take to collaboratively write a ~10-line shell script?)

solardiz:
> ~43s to load a module

This is weird, and I think is an issue we haven't seen so far. Somehow already the "Freezing user space processes" line is delayed, so maybe the delay isn't in LKRG, but in some concurrent activity in the kernel.

sempervictus:
A bit of sleep does wonders. Read that again this morning and realized you weren't asking me to insert adjacent to where the `sed` call was inserting the statement but to move the actual line in the script file down. moved that and the `lookup_fast` thing was fixed last night.

solardiz:
@sempervictus Thanks! This looks good enough to me now, so I'd merge, but let's wait for @Adam-pi3's OK.

sempervictus:
thank you, sorry for the rigmarole.

Adam-pi3:
Thanks @solardiz and @sempervictus. LGTM

PR #164, opened 2022-02-18
copy-builtin.sh: Drop unneeded pushd/popd

sempervictus:
### Description
Cleanup the copy script per @solardiz to remove pushd/popd calls since they're not needed.

### How Has This Been Tested?
Has not been tested, can't see how it could blow up


solardiz:
In the commit message, please describe the change, not who suggested it. So e.g.:

```
copy-builtin.sh: Drop unneeded pushd/popd
```

Issue #132, opened 2021-12-29
CI: Ubuntu Groovy failing


solardiz:
We started getting this:

```
I: Retrieving InRelease 
I: Retrieving Release 
E: Failed getting release file http://archive.ubuntu.com/ubuntu/dists/groovy/Release
‣ Detaching namespace
‣ Setting up package cache…
‣ Setting up package cache /home/runner/work/lkrg/lkrg/mkosi.cache complete
‣ Setting up temporary workspace.
‣ Temporary workspace set up in /var/tmp/mkosi-h726qpf8
‣ Running first (development) stage…
‣  Creating image with partition table…
‣  Created image with partition table as /home/runner/work/lkrg/lkrg/.mkosi-j5lzdb18
‣  Attaching /home/runner/work/lkrg/lkrg/.mkosi-j5lzdb18 as loopback…
‣  Attached /dev/loop3
‣  Formatting root partition…
‣  Mounting image…
‣  Setting up basic OS tree…
‣  Mounting Package Cache
‣  Installing Ubuntu…
‣  (Unmounting Package Cache)
‣  (Unmounting image)
‣  (Detaching /dev/loop3)
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.8/dist-packages/mkosi/__main__.py", line 32, in <module>
    main()
  File "/usr/local/lib/python3.8/dist-packages/mkosi/__main__.py", line 26, in main
    run_verb(a)
  File "/usr/local/lib/python3.8/dist-packages/mkosi/__init__.py", line 7812, in run_verb
    manifest = build_stuff(args)
  File "/usr/local/lib/python3.8/dist-packages/mkosi/__init__.py", line 7222, in build_stuff
    image = build_image(args, root, do_run_build_script=True)
  File "/usr/local/lib/python3.8/dist-packages/mkosi/__init__.py", line 6968, in build_image
    install_distribution(args, root, do_run_build_script, cached_tree)
  File "/usr/local/lib/python3.8/dist-packages/mkosi/__init__.py", line 3115, in install_distribution
    install[args.distribution](args, root, do_run_build_script)
  File "/usr/lib/python3.8/contextlib.py", line 75, in inner
    return func(*args, **kwds)
[...]

PR #167, opened 2022-02-22
CI: Add Bionic and fix KVM misdetection 

vt-alt:
As discussed in https://github.com/lkrg-org/lkrg/issues/132
Tested: https://github.com/vt-alt/lkrg/actions/runs/1884073929 all OK.

While testing there was failure on `jammy`: https://github.com/vt-alt/lkrg/runs/5294495248?check_suite_focus=true I don't know if it's related to LKRG, and it's disappeared on next run.
```
[   12.240310] BUG: kernel NULL pointer dereference, address: 0000000000000000
```

### DIFF
--- a/.github/workflows/mkosi-boot.yml
+++ b/.github/workflows/mkosi-boot.yml
@@ -9,6 +9,7 @@ jobs:
             fail-fast: false
             matrix:
                 release:
+                    - bionic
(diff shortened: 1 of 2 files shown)


vt-alt:
Force-pushed to correct typo in commit message.

Issue #176, opened 2022-04-11
P_LKRG_STRONG_KPROBE_DEBUG incompatible with P_LKRG_DEBUG_BUILD


solardiz:
Uncommenting these lines in Makefile:

```
# ccflags-m := -ggdb -DP_LKRG_DEBUG_BUILD -finstrument-functions
# ccflags-y := ${ccflags-m}
# p_lkrg-objs += src/modules/print_log/p_lkrg_debug_log.o
```

is incompatible with also enabling:

```diff
+++ b/src/modules/print_log/p_lkrg_log_level_shared.h
@@ -41,7 +41,7 @@
 
 // Debug every time we enter/exit *kprobed* function?
 // not recommended - will be very noisy...
-//#define P_LKRG_STRONG_KPROBE_DEBUG
+#define P_LKRG_STRONG_KPROBE_DEBUG
```

Resulting in:

```
  CC [M]  /home/solar/lkrg/lkrg-build/src/modules/print_log/p_lkrg_debug_log.o
/home/solar/lkrg/lkrg-build/src/modules/print_log/p_lkrg_debug_log.c:119:29: error: 'p_compat_sys_add_key_entry' undeclared here (not in a function)
    P_LKRG_DEBUG_RULE_KPROBE(p_compat_sys_add_key),
                             ^
/home/solar/lkrg/lkrg-build/src/modules/print_log/p_lkrg_debug_log.c:21:47: note: in definition of macro 'P_LKRG_DEBUG_RULE'
 #define P_LKRG_DEBUG_RULE(fname) { (uintptr_t)fname, #fname }
                                               ^
/home/solar/lkrg/lkrg-build/src/modules/print_log/p_lkrg_debug_log.c:119:4: note: in expansion of macro 'P_LKRG_DEBUG_RULE_KPROBE'
    P_LKRG_DEBUG_RULE_KPROBE(p_compat_sys_add_key),
    ^
/home/solar/lkrg/lkrg-build/src/modules/print_log/p_lkrg_debug_log.c:119:29: error: 'p_compat_sys_add_key_ret' undeclared here (not in a function)
    P_LKRG_DEBUG_RULE_KPROBE(p_compat_sys_add_key),
                             ^
/home/solar/lkrg/lkrg-build/src/modules/print_log/p_lkrg_debug_log.c:21:47: note: in definition of macro 'P_LKRG_DEBUG_RULE'
[...]

PR #178, opened 2022-04-13
debug: fix build with P_LKRG_STRONG_KPROBE_DEBUG

oshogbo:
### Description
In commit 9c5ef16cb51d51947c62e9116c6b52a3c3618bd0 we changed these
function to be untraceable. Respect that and remove them from
the list of traced functions.

This should address issue https://github.com/lkrg-org/lkrg/issues/176.

### How Has This Been Tested?
Building with debug flag and P_LKRG_STRONG_KPROBE_DEBUG enabled.


oshogbo:
I fixed the reference.

Adam-pi3:
@solardiz it's a good question. I think, ideally would be to keep `notrace` attribute in normal build but remove it from the DEBUG build. @oshogbo would it be possible? (it should ;p)

solardiz:
Thank you, @oshogbo!

There's a typo in "Mart". Other than that, @Adam-pi3 can you take another look and perhaps approve merging this?

Adam-pi3:
LGTM

PR #181, opened 2022-04-26
CI: Use Docker to create bootable multi-arch images

vt-alt:
Currently i386 and arm64v8. (I didn't find a way to create image for arm32v7 yet.)
x86_64 is just an example to prove it's working, since other arches are failing.
Perhaps we should delete x86_64 target.
And maybe we should delete arm64v8 either, it's slower than i386 and wasn't requested.

### DIFF
--- a/.github/workflows/docker-boot.sh
+++ b/.github/workflows/docker-boot.sh
@@ -1,0 +1,93 @@
+#!/bin/bash
+
+PATH=$PATH:/sbin
+
+image=ubuntu:jammy
+qemu=qemu-system-x86_64
+opts="-bios bios.bin"
+console=ttyS0
+root=/dev/sda
+append=no_timer_check
+init=/lkrg/.github/workflows/run-boot-tests.sh
+kernel=linux-virtual
+[ -n "$1" ] && declare "$@"
+
+[ -v pkgs ] || pkgs="libelf-dev linux-headers-generic $kernel"
+
+echo "::group::Build image ($image)"
+set -exfu
+
+# Keep it in other directory since we will COPY . .
+td=$(mktemp -d)
+
+# Generate system and build in the Docker.
+docker build --tag test -f - . <<EOF
+# bionic is the latest Ubuntu with i386 support.
+FROM $image
+ENV DEBIAN_FRONTEND="noninteractive"
+RUN apt-get update -y && \
+    apt-get install -y \
+            gcc \
+            git \
+            make \
+            $pkgs
+WORKDIR /lkrg
+COPY . .
+RUN git clean -dxfq
+RUN DESTDIR= ./mkosi.build
+RUN depmod -a \$(cd /lib/modules; ls)
+EOF
+
+# Convert Docker container into QEMU image (first extract rootfs).
+docker rm test-container || :
+docker create --name test-container test
+mkdir -p $td/rootfs
+docker export test-container | tar x -C $td/rootfs
+docker rm test-container
+
+# Since inside of the Docker is qemu-user we need to do as much
+# as possible heavy operations outside of emulation.
+# Also, qemu-system will be run externally to avoid double emulation.
+
+qemu-img create -f raw $td/disk.img 8G
+mkfs.ext4 -q -d $td/rootfs -F $td/disk.img
+# Instead of hurling with the grub will boot using -kernel/-initrd.
+
+set +xe
+echo "::endgroup::"
+
+echo "::group::Boot image ($qemu)"
+# Output command in a handy way for a user.
+set -x
+set -- $qemu \
+	-m 2G \
+	-nographic \
+	-no-reboot \
+	-drive file=$td/disk.img,format=raw \
+	-kernel $(find $td/rootfs/boot -name vmlinuz-* -print -quit) \
+	-initrd $(find $td/rootfs/boot -name initrd.img-* -print -quit) \
+	${dtb+-dtb $(find $td/rootfs/lib/firmware -name $dtb)} \
+	$opts \
+	-append "console=$console root=$root $append panic=-1 oops=panic panic_on_warn softlockup_panic=1 init=$init"
+set +x
+# Different versions of QEMU and kernels are not reliable at powering off ARM32
+# machines, so seek help from expect(1) to terminate qemu when power off is
+# failed.
+expect - <<EOF -- "$@"
+set timeout 300
+spawn -noecho {*}\$argv
+expect {
+	"reboot: Power down" {
+		send_user "ABORT\n"
+		exit
+	} "Reboot failed -- System halted" {
+		send_user "ABORT\n"
+		exit
+	} timeout {
+		send_user "TIMEOUT\n"
+		exit
+	}
+}
+EOF
+rm -rf $td
+echo "::endgroup::"
--- a/.github/workflows/docker-boot.yml
+++ b/.github/workflows/docker-boot.yml
@@ -1,0 +1,63 @@
+name: docker arch boot
+on: [push, pull_request]
+
+jobs:
+    build:
+        runs-on: ubuntu-20.04
+        strategy:
+            fail-fast: false
+            matrix:
+                include:
+                    - args: image=i386/ubuntu:bionic
+                            qemu=qemu-system-i386
+                      install: qemu-system-x86
+
+                    - args: image=arm64v8/ubuntu:impish
+                            opts="-M virt,gic-version=3 -cpu max"
+                            console=ttyAMA0
+                            root=/dev/vda
+                            qemu=qemu-system-aarch64
+                      install: qemu-system-arm
+
+                    # It's possible to use: opts="-M raspi2b" dtb=bcm2709-rpi-2-b.dtb root=/dev/mmcblk0
+                    # but since power-off is unreliable anyway (until QEMU 6.2) use simpler virt machine.
+                    - args: image=arm32v7/ubuntu:jammy
+                            opts="-M virt"
+                            console=ttyAMA0
+                            root=/dev/vda
+                            kernel="linux-raspi linux-modules-extra-raspi"
+                            qemu=qemu-system-arm
+                      install: qemu-system-arm
+
+        steps:
+            - uses: docker/setup-qemu-action@v1
+            - run: sudo apt-get update
+              # It's possible to update QEMU to (6.0) on focal (ubuntu-20.04) with
+              #   add-apt-repository ppa:canonical-server/server-backports
+              # but this is still not enough to boot & power-off raspi2b properly.
+            - run: sudo apt-get install -y qemu-utils e2fsprogs expect ${{ matrix.install }}
+
+            - uses: actions/checkout@v1
+            - name: Enable LKRG debugging options
+              run: |
+                  sed -i '/P_LKRG_JUMP_LABEL_STEXT_DEBUG/s/\/\///' src/modules/print_log/p_lkrg_log_level_shared.h
+                  git diff
+            - name: Build and boot test image
+              run: .github/workflows/docker-boot.sh ${{ matrix.args }} 2>&1 | tr -d '\r' | tee boot.log
+
+            - name: Check boot.log for 'LKRG initialized successfully'
+              run: |
+                  grep 'Linux version' boot.log
+                  grep 'LKRG initialized successfully' boot.log
+            - name: Check that boot.log does not contain problems
+              run:
+                  "! grep -v 'The requested image.s platform .* does not match the detected host platform' boot.log |
+                     grep -E 'Kernel panic|BUG:|WARNING:|Oops|Call Trace'"
+            - name: Check that boot-tests script finished successfully
+              run: grep 'run-boot-tests.sh - SUCCESS' boot.log
+            - name: Check that boot.log contains shutdown sequence
+              run: |
+                  grep 'sysrq: Power Off' boot.log
+                  grep 'reboot: Power down' boot.log
+
+# vim: sw=4
--- a/.github/workflows/mkosi-boot.yml
+++ b/.github/workflows/mkosi-boot.yml
@@ -49,7 +49,7 @@ jobs:
                   grep 'Linux version' boot.log
                   grep 'LKRG initialized successfully' boot.log
             - name: Check that boot.log does not contain problems
-              run: "! egrep 'Kernel panic|BUG:|WARNING:|Oops|Call Trace' boot.log"
+              run: "! grep -E 'Kernel panic|BUG:|WARNING:|Oops|Call Trace' boot.log"
             - name: Check that boot-tests script finished successfully
               run: grep 'run-boot-tests.sh - SUCCESS' boot.log
             - name: Check that boot.log contains shutdown sequence
--- a/.github/workflows/mkosi-mainline.yml
+++ b/.github/workflows/mkosi-mainline.yml
@@ -54,7 +54,7 @@ jobs:
                   grep 'Linux version' boot.log
                   grep 'LKRG initialized successfully' boot.log
             - name: Check that boot.log does not contain problems
-              run: "! egrep 'Kernel panic|BUG:|WARNING:|Oops|Call Trace' boot.log"
+              run: "! grep -E 'Kernel panic|BUG:|WARNING:|Oops|Call Trace' boot.log"
             - name: Check that boot-tests script finished successfully
               run: grep 'run-boot-tests.sh - SUCCESS' boot.log
             - name: Check that boot.log contains shutdown sequence
--- a/.github/workflows/run-boot-tests.sh
+++ b/.github/workflows/run-boot-tests.sh
@@ -8,6 +8,10 @@
 
 set -eux -o pipefail
 
+if [ ! -d /sys/module/p_lkrg ]; then
+	modprobe p_lkrg
+fi
+
 # Trigger loading vhost_vsock module (using UMH call to modprobe).
 # Device numbers are from /lib/modules/*/modules.devname
 mknod /dev/test c 10 241
@@ -19,3 +23,9 @@ sleep 21
 
 # Failed tests will not output this line.
 echo "$0 - SUCCESS"
+
+# If there is no systemd shutdown manually.
+if [ ! -d /run/systemd/system ]; then
+	echo o > /proc/sysrq-trigger
+	sleep 11
+fi
--- a/mkosi.build
+++ b/mkosi.build
@@ -14,4 +14,4 @@ make -j$(nproc) $@
 # Install if compiled.
 [ -e p_lkrg.ko ] && [ -v DESTDIR ] &&
 install -Dpm 644 p_lkrg.ko $DESTDIR/lib/modules/$KERNELRELEASE/extra/p_lkrg.ko &&
-cp -av ./ $DESTDIR/lkrg/
+cp -av ./ $DESTDIR/lkrg/ || :


solardiz:
Thank you, @vt-alt! It looks like this exposed bugs.

@Adam-pi3 Is this something we'd look into before we merge this PR?

i386 is failing with:

```
  [    9.544673] [p_lkrg] Loading LKRG...
  [    9.544979] [p_lkrg] System does NOT support SMEP. LKRG can't enforce SMEP validation :(
  [    9.545542] [p_lkrg] System does NOT support SMAP. LKRG can't enforce SMAP validation :(
  [    9.553544] Freezing user space processes ... (elapsed 0.001 seconds) done.
  [    9.555577] OOM killer disabled.
  [   10.644737] [p_lkrg] [kretprobe] register_kretprobe() for <ovl_create_or_link> failed! [err=-22]
  [   10.645346] [p_lkrg] Trying to find ISRA / CONSTPROP name for <ovl_create_or_link>
  [   10.852290] [p_lkrg] [kretprobe] register_kretprobe() for ovl_create_or_link failed and ISRA / CONSTPROP version not found!
  [   10.852629] [p_lkrg] Can't hook 'ovl_create_or_link' function. This is expected if you are not using OverlayFS.
  [   11.645783] [p_lkrg] <Exploit Detection> process [modprobe | 192] has invalid base for stack pointer!
  [   11.646064] [p_lkrg] <Exploit Detection> process [modprobe | 192] has invalid stack pointer (stack size mismatch)!
  [   11.646800] [p_lkrg] <Exploit Detection> Stack pointer corruption (ROP?) - pCFI violation: process[modprobe | 192] !!!
  [   11.647393] [p_lkrg] <Exploit Detection> Trying to kill process[modprobe | 192]!
  [   11.648307] [p_lkrg] <Exploit Detection> Stack pointer corruption (ROP?) - pCFI violation: process[modprobe | 192] !!!
  [   11.648560] [p_lkrg] <Exploit Detection> Trying to kill process[modprobe | 192]!
  [   13.152004] [p_lkrg] LKRG initialized successfully!
  [   13.152237] OOM killer enabled.
  [   13.152651] Restarting tasks ... 
  [   13.153345] [p_lkrg] <Exploit Detection> process [modprobe | 192] has invalid base for stack pointer!
  [   13.154116] [p_lkrg] <Exploit Detection> process [modprobe | 192] has invalid stack pointer (stack size mismatch)!
  [   13.154739] [p_lkrg] <Exploit Detection> Stack pointer corruption (ROP?) - pCFI violation: process[modprobe | 192] !!!
  [   13.155343] [p_lkrg] <Exploit Detection> Trying to kill process[modprobe | 192]!
  [   13.155984] [p_lkrg] <Exploit Detection> Stack pointer corruption (ROP?) - pCFI violation: process[modprobe | 192] !!!
  [   13.157674] [p_lkrg] <Exploit Detection> Trying to kill process[modprobe | 192]!
  [   13.158746] done.
  /lkrg/.github/workflows/run-boot-tests.sh: line 13:   192 Killed                  modprobe p_lkrg
  [   13.350177] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00008900
  [   13.350177] 
  [   13.350712] CPU: 0 PID: 1 Comm: run-boot-tests. Tainted: G           OE    4.15.0-176-generic #185-Ubuntu
  [   13.351107] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS 1.13.0-1ubuntu1.1 04/01/2014
  [   13.351877] Call Trace:
  [   13.353007]  dump_stack+0x60/0x7a
  [   13.353152]  panic+0x94/0x1e0
  [   13.353543]  do_exit+0x743/0xa60
  [   13.353948]  copy_oldmem_page+0xe0/0xe0
  [   13.354457]  SyS_exit_group+0x16/0x20
  [   13.354881]  do_fast_syscall_32+0x7f/0x200
  [   13.355401]  entry_SYSENTER_32+0x68/0xbb
  [   13.356040] EIP: 0xb7f09d09
  [   13.356217] EFLAGS: 00000296 CPU: 0
  [   13.356672] EAX: ffffffda EBX: 00000089 ECX: 00000000 EDX: b7ed5884
  [   13.357602] ESI: b7ed2210 EDI: 00000000 EBP: b7ed4000 ESP: bfcb089c
  [   13.358536]  DS: 007b ES: 007b FS: 0000 GS: 0033 SS: 007b
  [   13.359824] Kernel Offset: 0x5000000 from 0xc1000000 (relocation range: 0xc0000000-0xf7dfdfff)
```

Looks like our pCFI is broken on i386?

arm64v8 is failing with:

```
  [   60.871210] [p_lkrg] Loading LKRG...
  [   60.919240] Freezing user space processes ... (elapsed 0.002 seconds) done.
  [   60.923377] OOM killer disabled.
  [   61.344795] Bad mode in Synchronous Abort handler detected on CPU0, code 0x34000002 -- BTI
  [   61.346392] CPU: 0 PID: 169 Comm: modprobe Tainted: G           OE     5.13.0-40-generic #45-Ubuntu
  [   61.347370] Hardware name: linux,dummy-virt (DT)
  [   61.348530] pstate: 60400805 (nZCv daif +PAN -UAO -TCO BTYPE=-c)
  [   61.349795] pc : __put_seccomp_filter+0x0/0xc0
  [   61.350928] lr : p_dump_seccomp+0x64/0x80 [p_lkrg]
  [   61.351643] sp : ffff800008ba3950
  [   61.352493] x29: ffff800008ba3950
  [   61.353222]  x28: 0000000000000100
  [   61.353986]  x27: ffff444ac3b9ae80
  [   61.354649] 
  [   61.355058] x26: ffffb82598d4a5c8
  [   61.355945]  x25: ffffb82598d52580
  [   61.356597]  x24: 0000000000000000
  [   61.357252] 
  [   61.357662] x23: ffffb825a16a8000
  [   61.358312]  x22: 0000000000000001
  [   61.358847]  x21: ffffb82598d51000
  [   61.359363] 
  [   61.359945] x20: ffff444ac526e810
  [   61.360488]  x19: ffff444ac3b9ae80
  [   61.361028]  x18: 00000000ffffffff
  [   61.361545] 
  [   61.361917] x17: 000000000000003f
  [   61.362441]  x16: 000000000000000b
  [   61.362981]  x15: 0000000000000dff
  [   61.363496] 
  [   61.364072] x14: 00000000000001bc
  [   61.364645]  x13: ffff444b3fbf30c0
  [   61.365190]  x12: 0000000000000006
  [   61.365707] 
  [   61.366078] x11: ffffb825a10bb2e0
  [   61.366602]  x10: 0000000000000000
  [   61.367171]  x9 : ffffb82598d3f434
  [   61.368022] 
  [   61.368434] x8 : ffff444ac526ea48
  [   61.369104]  x7 : 0000000000000000
  [   61.369785]  x6 : 000000000000003f
  [   61.370400] 
  [   61.370897] x5 : 0000000000000040
  [   61.371597]  x4 : ffff800008ba3998
  [   61.372398]  x3 : 0000000000000001
  [   61.373038] 
  [   61.373477] x2 : 0000000000000001
  [   61.374156]  x1 : ffffb8259f3ca440
  [   61.374777]  x0 : 0000000000000000
  [   61.375328] 
  [   61.378531] Kernel panic - not syncing: bad mode
  [   61.378606] CPU: 0 PID: 169 Comm: modprobe Tainted: G           OE     5.13.0-40-generic #45-Ubuntu
  [   61.378670] Hardware name: linux,dummy-virt (DT)
  [   61.378731] Call trace:
  [   61.378791]  dump_backtrace+0x0/0x1f0
  [   61.378851]  show_stack+0x24/0x30
  [   61.378911]  dump_stack+0xd0/0x128
  [   61.378972]  panic+0x154/0x36c
  [   61.379032]  bad_mode+0x94/0xb0
  [   61.379092]  el1_inv+0x6c/0xcc
  [   61.379152]  el1_sync_handler+0x58/0xd0
  [   61.379212]  el1_sync+0x7c/0x100
  [   61.379273]  __put_seccomp_filter+0x0/0xc0
  [   61.379333]  p_update_ed_process+0xd8/0x1c0 [p_lkrg]
  [   61.379393]  p_dump_task_f+0x48/0x130 [p_lkrg]
  [   61.379454]  p_exploit_detection_init+0x1f8/0x770 [p_lkrg]
  [   61.379515]  p_lkrg_register+0x1f0/0x1000 [p_lkrg]
  [   61.379600]  do_one_initcall+0x50/0x270
  [   61.379661]  do_init_module+0x60/0x280
  [   61.379721]  load_module+0x988/0xb20
  [   61.379781]  __do_sys_finit_module+0xa8/0x11c
  [   61.379841]  __arm64_sys_finit_module+0x2c/0x40
  [   61.379902]  invoke_syscall+0x78/0x100
  [   61.379962]  el0_svc_common.constprop.0+0x54/0x184
  [   61.380023]  do_el0_svc+0x84/0x9c
  [   61.380083]  el0_svc+0x2c/0x54
  [   61.380144]  el0_sync_handler+0xa4/0x130
  [   61.380204]  el0_sync+0x19c/0x1c0
  [   61.380950] Kernel Offset: 0x382597200000 from 0xffff800008000000
  [   61.381015] PHYS_OFFSET: 0xffffbbb580000000
  [   61.381077] CPU features: 0x00000651,2c23085f
  [   61.381137] Memory Limit: none
```

I guess the calling convention we use from our `p_dump_seccomp()` is somehow wrong for this arch?

```c
#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,9,0)
   P_SYM(p_put_seccomp_filter)(p_task->seccomp.filter);
#else
   P_SYM(p_put_seccomp_filter)(p_task);
#endif
```

vt-alt:
Perhaps we should rename workflow from `Multi-arch boot` to `docker boot`. Delete x86_64. Delete aarch64?

Also to note build/boot times:

- x86_64 ~3m
- i386 ~2m
- aarch64 ~13m

solardiz:
@vt-alt Can you explain the below comment in other words? In what case would double emulation occur? I think this comment should be made clearer.

As I understand, you're using Docker to prepare a userland tree, which you then turn into an image bootable without Docker. I don't see another easy way you could have used that Docker-prepared tree differently, do you?

```sh
# Since inside of the docket is qemu-user we need to do as much
# as possible heavy operations outside of emulation.
# qemu-system will be run externally to avoid double emulation.
```

If my understanding is correct, renaming to `docker boot` would be weird - that's not what we're actually doing here, we're merely (ab)using Docker to create a userland tree, not to boot anything.

It's unfortunate the aarch64 test is slow, but it does appear to have identified an issue, so looks valuable to keep (assuming that we get that issue fixed).

vt-alt:
@solardiz That comment was to note that we could run `qemu-system-i386` inside of Docker container, but in that case it would create double emulation (qemu-user run by the Docker runtime to emulate Docker container where `qemu-system-i386` is run to emulates boot of the kernel). So I run `qemu-system-i386` outside of container in the host system.

Overall this is docker build (in compare to mkosi build), and in both cases qemu is used to boot the image So to call it like 'somehow-qemu-related boot' would be non-informative.

vt-alt:
@solardiz BTW, in what distributions on ARM32 LKRG is intended to work?

solardiz:
> in what distributions on ARM32 LKRG is intended to work?

As I recall, @Adam-pi3 tested it on a Raspberry Pi, so it probably was some distro typical for that platform - perhaps Raspbian or such? Where it was tested isn't the same as where it's expected to be used - we just like to support those archs in general, and actual uses may be by some embedded devices with their own distros. However, yes, we do need to pick certain distro(s) for our development and testing use. So let's wait to hear from Adam in case he has suggestions on distros to use.

Adam-pi3:
I was testing LKRG on Raspberry Pi 3 and 4. In both cases I was using Ubuntu server for AArch64 (https://ubuntu.com/download/server/arm) using the top LTS version at that time (if I correctly remember it was version 18.04 and 20...).

solardiz:
@Adam-pi3 I guess one of those wasn't "for AArch64"?

Adam-pi3:
@solardiz correct, sorry my mistake, one of them was ARM (32 bits)

vt-alt:
JFYI (as it takes too much my time without good results)
On `arm32v7/ubuntu:jammy` I am not able to compile lkrg on `linux-generic` kernel dues no absence of `CONFIG_JUMP_LABEL`.
On the same ubuntu, I am able to build and successfully boot (w/o kernel panics) on `linux-raspi` kernel (also with `linux-modules-extra-raspi` required for `virtio-blk`), but this kernel does not shutdown properly on `qemu-system-arm -M virt`. I am not good at debugging `pm_power_off` issues. Many hours of googling didn't help either.

vt-alt:
> I am not good at debugging `pm_power_off`

(JFYI ) `linux-generic` is good at powering off, and it have pointer to `psci_sys_poweroff` at `pm_power_off` (where `linux-raspi` have `0`; looked this up with `/proc/kallsyms` and `x` command in QEMU monitor), this _psci_ thing is built only when `CONFIG_ARM_PSCI` is enabled, which is not enabled for `linux-raspi`.

vt-alt:
Finally, this took the time, to add ARM32 to the tests. Instead of trying to find good QEMU version, emulated machine type, dtb, and the kernel, I decided to wrap things in `expect` to just quit qemu on power-off.
Test time seems ~15-18 minutes. And the test result is green. \o/

solardiz (approved):
It's good to have these tests. My main concerns are that 2 of the 3 archs added here are failing now (should we fix first? or after merging this?) and that we might be running into some resource usage limits at GitHub Actions later by running jobs that are so time-consuming (but I guess let's try and see, we can always disable later).

solardiz:
@Adam-pi3 Do you prefer that we merge this now or (at least try to) fix the two failures first?

Adam-pi3:
Good question. It will take me some time before I can sit-down and analyze these issue - especially on ARM. I don't know how it affects our GitHub resources if we keep getting this errors. I'm OK merging it now.

vt-alt:
I wish we can speed this up somehow. Maybe it's possible to save docker cache (most of the time, it seems, is spent installing packages and generating initrd image), but with infrequent commits (<=1 per day in average) this perhaps will be not too useful and amounts to overengineering.

vt-alt:
Two <20 minute jobs is perhaps no problem for GA. I compared to `strace` CI run
https://github.com/strace/strace/actions/runs/2261850316
there are about 40 jobs of 40-15 minutes and ~2 commits per day in average, and I didn't hear about GA problems from @ldv-alt.

solardiz:
@vt-alt I think it'd make sense to temporarily comment out the i386 and aarch64 sections, and then merge this. We'll then plan on uncommenting them along with making the corresponding code fixes. Makes sense to you, too? If so, can you please make that change to this PR (comment out those sections)? Thank you!

vt-alt:
I am not sure about it. Maybe you can merge and then add commit to comment it (or even delete lines), which can be reverted later?

solardiz:
> [   61.349795] pc : __put_seccomp_filter+0x0/0xc0

I don't immediately see what could be causing this, but FWIW while looking into it I noticed this comment:

```c
/**
 * struct seccomp - the state of a seccomp'ed process
 *
 * @mode:  indicates one of the valid values above for controlled
 *         system calls available to a process.
 * @filter: must always point to a valid seccomp-filter or NULL as it is
 *          accessed without locking during system call entry.
 *
 *          @filter must only be accessed from the context of current as there
 *          is no read locking.
 */
struct seccomp {
        int mode;
        atomic_t filter_count;
        struct seccomp_filter *filter;
};
```

and specifically this part:

```c
 *          @filter must only be accessed from the context of current as there
 *          is no read locking.
```

While most of our calls to `p_update_ed_process` do pass `current` in there, a few don't - and this could potentially be problematic - depending on how that non-current task can (not) be changing under us. I don't think this is the problem seen here as it'd be a subtle race and not something easily reproducible, but nevertheless it's a concern.

solardiz:
> While most of our calls to `p_update_ed_process` do pass `current` in there, a few don't

BTW, it's precisely one of those cases in the backtrace here - however, it's with the processes frozen during LKRG initialization, so should be fine?

solardiz:
I tried adding a debugging commit in my fork of the repo, and when the `put_seccomp_filter` call is skipped (which is safe to skip there because the `filter` is `NULL` anyway) we're getting a similar crash on a different function call made via `P_SYM` later. So maybe our entire `P_SYM` mechanism is currently non-working on aarch64.

solardiz:
> think we need to somehow avoid tail call optimization in LKRG when using `P_SYM`

I've just tried adding:

```c
   __asm__ __volatile__ ("");
```

right after the known problematic calls and before the end of the calling function, but somehow this didn't make a difference. So either this didn't prevent the optimization or my understanding is wrong. I did this test on GitHub Actions, so do not have a copy of the compiled code to check.

vt-alt:
> I did this test on GitHub Actions, so do not have a copy of the compiled code to check.

@solardiz There are two options.
1. I tried to make workflow easily reproducible — just a single script with parameters. You may run it in virtual machine with Ubuntu (it even works on ALT). You will need to install `docker.io` package in addition to what is installed in workflow.
2. Also, workflow can easily be modified to save artifacts with [actions/upload-artifact](https://github.com/actions/upload-artifact) action ([more docs](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts)), and they will be [downloadable](https://docs.github.com/en/actions/managing-workflow-runs/downloading-workflow-artifacts) in Actions pane. (You may save whole `rootfs` exported from docker).

solardiz:
@vt-alt Thanks. I also thought of inserting an `objdump -d` on one of our `.o` files, so that we'd see the generated code in the logs.

solardiz:
I briefly experimented at https://godbolt.org with:

```c
extern void f(void);
extern void g(void);

void fut(int i)
{
    if (i) f(); else g();
    asm("");
}
```

compiling it with `ARM64 gcc trunk`. With the `asm` line in there, this produces "lengthy" code that uses `bl` and `ret` in both branches, even at `-O3`. Without the `asm`, it's the same at `-O1`, but gets optimized to using `b` at `-O2` or higher:

```
fut(int):
        cbz     w0, .L2
        b       f()
.L2:
        b       g()
```

So my `asm` should have worked to prevent this in LKRG.

My new guess is that maybe it's the other way around - the Linux kernel's code for those functions expects them to be jumped to, not called, if that's how the only "calls" to them end up being optimized when building the kernel itself.

Anyhow, this is tricky for us - means we shouldn't hard-code one approach or the other, as the right one can vary per function and across kernel builds. Perhaps we could check the pointed-to instruction at runtime and, if it's a `bti`, call/jump it accordingly.

solardiz:
I've just disabled the two failing tests with a separate commit for now.

solardiz:
@vt-alt Per discussion with Adam we just had, I think we should drop or replace the `-cpu max` in the AArch64 tests, so that for now we'd test on a CPU that doesn't have BTI yet. We know we don't yet support BTI, but we'd like to be testing the rest of our AArch64 support nevertheless. Is this a change you'd make (in a new PR)? Thanks!

vt-alt:
Did you mean to drop from commented line in `workflows/docker-boot.yml`?
```
#                            opts="-M virt,gic-version=3 -cpu max"
```
I would produce extremely dumb commit message—why not change yourself with a good explanation?

solardiz:
I mean uncommenting this section at the same time with removal/edit of that option from there:

```
#                    - args: image=arm64v8/ubuntu:impish
#                            opts="-M virt,gic-version=3 -cpu max"
#                            console=ttyAMA0
#                            root=/dev/vda
#                            qemu=qemu-system-aarch64
#                      install: qemu-system-arm
```

Yes, I could experiment with this myself. I'm not sure removing this option will actually do the trick, or what other `-cpu` option we should ideally put in there. I guess one of us needs to read up on that or/and experiment.

vt-alt:
IC For some reason I didn't realize you want me to uncomment this block.

solardiz:
@vt-alt I've just tried re-enabling this block (in a branch in my fork) and changing the option to `-cpu max,bti=off` (no idea if `bti` is even supported in that option, just a guess based on examples I've seen for other AArch64 optional features). However, this only revealed that the test broke since then:

https://github.com/solardiz/lkrg/actions/runs/3351011388/jobs/5552186308

```
Run .github/workflows/docker-boot.sh image=arm64v8/ubuntu:impish opts="-M virt,gic-version=3 -cpu max,bti=off" console=ttyAMA0 root=/dev/vda qemu=qemu-system-aarch64 2>&1 | tr -d '\r' | tee boot.log
Build image (arm64v8/ubuntu:impish)
  ++ mktemp -d
  + td=/tmp/tmp.OE0O1sYheF
  + docker build --tag test -f - .
  Sending build context to Docker daemon   2.42MB
  Step 1/8 : FROM arm64v8/ubuntu:impish
  impish: Pulling from arm64v8/ubuntu
  166e0294b8df: Pulling fs layer
  166e0294b8df: Verifying Checksum
  166e0294b8df: Download complete
  166e0294b8df: Pull complete
  Digest: sha256:af787f9bfcb52606fbfc5247f9f45dc20f9635762ff7c84839aa75c098c2cc9f
  Status: Downloaded newer image for arm64v8/ubuntu:impish
   ---> bedf4ab662ae
  Step 2/8 : ENV DEBIAN_FRONTEND="noninteractive"
  Warning: rning] The requested image's platform (linux/arm64/v8) does not match the detected host platform (linux/amd64) and no specific platform was requested
   ---> Running in 9129b91cf92e
  Removing intermediate container 9129b91cf92e
   ---> cb3110a6ffd7
  Step 3/8 : RUN apt-get update -y &&     apt-get install -y             gcc             git             make             libelf-dev linux-headers-generic linux-virtual
  Warning: rning] The requested image's platform (linux/arm64/v8) does not match the detected host platform (linux/amd64) and no specific platform was requested
   ---> Running in a71b57e78d46
  Ign:1 http://ports.ubuntu.com/ubuntu-ports impish InRelease
  Ign:2 http://ports.ubuntu.com/ubuntu-ports impish-updates InRelease
  Ign:3 http://ports.ubuntu.com/ubuntu-ports impish-backports InRelease
  Ign:4 http://ports.ubuntu.com/ubuntu-ports impish-security InRelease
  Err:5 http://ports.ubuntu.com/ubuntu-ports impish Release
    404  Not Found [IP: 185.125.190.36 80]
  Err:6 http://ports.ubuntu.com/ubuntu-ports impish-updates Release
    404  Not Found [IP: 185.125.190.36 80]
  Err:7 http://ports.ubuntu.com/ubuntu-ports impish-backports Release
    404  Not Found [IP: 185.125.190.36 80]
  Err:8 http://ports.ubuntu.com/ubuntu-ports impish-security Release
    404  Not Found [IP: 185.125.190.36 80]
  Reading package lists...
  E: The repository 'http://ports.ubuntu.com/ubuntu-ports impish Release' does not have a Release file.
  E: The repository 'http://ports.ubuntu.com/ubuntu-ports impish-updates Release' does not have a Release file.
  E: The repository 'http://ports.ubuntu.com/ubuntu-ports impish-backports Release' does not have a Release file.
  E: The repository 'http://ports.ubuntu.com/ubuntu-ports impish-security Release' does not have a Release file.
  The command '/bin/sh -c apt-get update -y &&     apt-get install -y             gcc             git             make             libelf-dev linux-headers-generic linux-virtual' returned a non-zero code: 100
```

Will you help repair this, please?

FWIW, my attempted commit was: https://github.com/solardiz/lkrg/commit/44c4fe19ed2f43a5b1f49fd426dc4bbc649c5f1b

```diff
commit 44c4fe19ed2f43a5b1f49fd426dc4bbc649c5f1b
Author: Solar Designer <solar@openwall.com>
Date:   Sat Oct 29 12:04:28 2022 +0200

    CI: Re-enable AArch64 test, but without BTI

diff --git a/.github/workflows/docker-boot.yml b/.github/workflows/docker-boot.yml
index 1123c9f..f28e730 100644
--- a/.github/workflows/docker-boot.yml
+++ b/.github/workflows/docker-boot.yml
@@ -11,13 +11,13 @@ jobs:
 #                    - args: image=i386/ubuntu:bionic
 #                            qemu=qemu-system-i386
 #                      install: qemu-system-x86
-#
-#                    - args: image=arm64v8/ubuntu:impish
-#                            opts="-M virt,gic-version=3 -cpu max"
-#                            console=ttyAMA0
-#                            root=/dev/vda
-#                            qemu=qemu-system-aarch64
-#                      install: qemu-system-arm
+
+                    - args: image=arm64v8/ubuntu:impish
+                            opts="-M virt,gic-version=3 -cpu max,bti=off"
+                            console=ttyAMA0
+                            root=/dev/vda
+                            qemu=qemu-system-aarch64
+                      install: qemu-system-arm
 
                     # It's possible to use: opts="-M raspi2b" dtb=bcm2709-rpi-2-b.dtb root=/dev/mmcblk0
                     # but since power-off is unreliable anyway (until QEMU 6.2) use simpler virt machine.
```

solardiz:
@vt-alt Oh, I now vaguely recall we previously found Impish was gone. Now trying Jammy.

solardiz:
OK, on Jammy it failed with:

```
  + set -- qemu-system-aarch64 -m 2G -nographic -no-reboot -drive file=/tmp/tmp.ubbYMsHraJ/disk.img,format=raw -kernel /tmp/tmp.ubbYMsHraJ/rootfs/boot/vmlinuz-5.15.0-52-generic -initrd /tmp/tmp.ubbYMsHraJ/rootfs/boot/initrd.img-5.15.0-52-generic -M virt,gic-version=3 -cpu max,bti=off -append 'console=ttyAMA0 root=/dev/vda no_timer_check panic=-1 oops=panic panic_on_warn softlockup_panic=1 init=/lkrg/.github/workflows/run-boot-tests.sh'
  + set +x
  qemu-system-aarch64: can't apply global max-arm-cpu.bti=off: Property 'max-arm-cpu.bti' not found
```

So I didn't guess that option correctly. Will try specifying an explicit CPU type, then.

solardiz:
This worked:

```
                    - args: image=arm64v8/ubuntu:jammy
                            opts="-M virt,gic-version=3 -cpu cortex-a57"
```

Now to test whether `-cpu max` would possibly also have worked on current Jammy.

solardiz:
BTW, I got `-cpu cortex-a57` from https://wiki.ubuntu.com/ARM64/QEMU

solardiz:
> Now to test whether `-cpu max` would possibly also have worked on current Jammy.

With that, the crash is similar to what we had before. Here's what I got now:

https://github.com/solardiz/lkrg/actions/runs/3351330872/jobs/5552699544

```
  [   66.032578] LKRG: ALIVE: Loading LKRG
  [   66.085474] Freezing user space processes ... (elapsed 0.004 seconds) done.
  [   66.092324] OOM killer disabled.
  [   66.565282] Unhandled 64-bit el1h sync exception on CPU0, ESR 0x34000002 -- BTI
  [   66.565805] CPU: 0 PID: 169 Comm: modprobe Tainted: G           OE     5.15.0-52-generic #58-Ubuntu
  [   66.566044] Hardware name: linux,dummy-virt (DT)
  [   66.566365] pstate: 60400805 (nZCv daif +PAN -UAO -TCO -DIT -SSBS BTYPE=-c)
  [   66.566610] pc : __put_seccomp_filter+0x0/0xc0
  [   66.567176] lr : p_dump_seccomp+0x64/0x7c [lkrg]
  [   66.567853] sp : ffff80000841b8f0
  [   66.567996] x29: ffff80000841b8f0 x28: ffff80000841bbe0 x27: 0000000000000000
  [   66.568392] x26: ffff00000a442f40 x25: ffffa4dafff6f580 x24: 0000000000000000
  [   66.568685] x23: ffffa4db53bf88c0 x22: 0000000000000001 x21: ffffa4dafff6e000
  [   66.568950] x20: ffff000003ba2750 x19: ffff00000a442f40 x18: 0000000000000014
  [   66.569215] x17: 00000000d2794962 x16: ffffa4db51b83400 x15: 00000000cdc0a090
  [   66.569500] x14: 0000000000000344 x13: 0000000000000003 x12: ffff00007fbf2f80
  [   66.569788] x11: 0000000000000006 x10: ffffa4db5345e760 x9 : ffffa4dafff5bd18
  [   66.570091] x8 : ffff000003ba2e08 x7 : 0000000000000000 x6 : 000000000000003f
  [   66.570563] x5 : 0000000000000040 x4 : ffff80000841b938 x3 : ffffa4dafff6e000
  [   66.570954] x2 : 0000000000000001 x1 : ffffa4db515f1d74 x0 : 0000000000000000
  [   66.571460] Kernel panic - not syncing: Unhandled exception
  [   66.571612] CPU: 0 PID: 169 Comm: modprobe Tainted: G           OE     5.15.0-52-generic #58-Ubuntu
  [   66.571820] Hardware name: linux,dummy-virt (DT)
  [   66.572050] Call trace:
  [   66.572115]  dump_backtrace+0x0/0x1ec
  [   66.572443]  show_stack+0x24/0x30
  [   66.572573]  dump_stack_lvl+0x68/0x84
  [   66.572695]  dump_stack+0x18/0x34
  [   66.572836]  panic+0x18c/0x39c
  [   66.572960]  enter_el1_irq_or_nmi+0x0/0x5c
  [   66.573083]  el1h_64_sync_handler+0x64/0xd0
  [   66.573215]  el1h_64_sync+0x7c/0x80
  [   66.573327]  __put_seccomp_filter+0x0/0xc0
  [   66.573441]  p_update_ed_process+0xe8/0x1e0 [lkrg]
  [   66.573657]  p_dump_task_f+0x48/0x130 [lkrg]
  [   66.573888]  p_exploit_detection_init+0x1f0/0x720 [lkrg]
  [   66.574142]  p_lkrg_register+0x244/0x1000 [lkrg]
  [   66.574409]  do_one_initcall+0x50/0x250
  [   66.574526]  do_init_module+0x50/0x260
  [   66.574641]  load_module+0x9dc/0xbc0
  [   66.574754]  __do_sys_finit_module+0xa8/0x11c
  [   66.574866]  __arm64_sys_finit_module+0x2c/0x40
  [   66.575015]  invoke_syscall+0x78/0x100
  [   66.575251]  el0_svc_common.constprop.0+0x54/0x184
  [   66.575399]  do_el0_svc+0x84/0x9c
  [   66.575570]  el0_svc+0x48/0x1b0
  [   66.575684]  el0t_64_sync_handler+0xa4/0x130
  [   66.575796]  el0t_64_sync+0x1a4/0x1a8
  [   66.576497] Kernel Offset: 0x24db49400000 from 0xffff800008000000
  [   66.576600] PHYS_OFFSET: 0x40000000
  [   66.576675] CPU features: 0x00000751,6d333c5f
  [   66.576911] Memory Limit: none
```

So I'm going to create a pull request with `-cpu cortex-a57`.

vt-alt:
> @vt-alt I've just tried re-enabling this block (in a branch in my fork) and changing the option to `-cpu max,bti=off` (no idea if `bti` is even supported in that option, just a guess based on examples I've seen for other AArch64 optional features).

It seems (from [qemu/target/arm/cpu64.c](https://github.com/qemu/qemu/blob/master/target/arm/cpu64.c#L1055)) that `cpu=max` enables `FEAT_BTI` unconditionally.

solardiz:
> It seems (from [qemu/target/arm/cpu64.c](https://github.com/qemu/qemu/blob/master/target/arm/cpu64.c#L1055)) that `cpu=max` enables `FEAT_BTI` unconditionally.

Thank you for checking. However, I am not so sure: from that function, they'd appear to also enable e.g. SVE unconditionally, however it can be post-disabled with `sve=off` (https://github.com/qemu/qemu/blob/master/docs/system/arm/cpu-features.rst), for which I don't immediately see where it is implemented in code. So from that function alone I cannot conclude there isn't a `bti=off` or such (but my attempt to guess and use it failed, as above, so probably there is not).

Anyway, I think for now we're good with the commit I made today. (Perhaps we could use a higher CPU, though, like A72.)

PR #182, opened 2022-04-28
p_ed_enforce_pcfi: validate the p_task stack but not irq_stack

lc85446:
In the irq context, we should not get the p_task stack from p_regs.

### Description
 The security_capable() sometimes may be called in the interrupt context , then the p_ed_enforce_pcfi() would get the irq_stack from pt_regs and warn about the stack pointer mismatch and kill the victim process.   

 So we should use p_get_thread_sp(p_task)  in the interrupt context and use  p_regs_get_sp(p_regs) otherwise. 

 The interrupst_enabled() can be used to check whether running in the interrupt context,  an alternative and maybe better method  is checking  whether on_irq_stack(kernel_stack_pointer(regs) is true,  but unfortunately the on_irq_stack() is not exported by the kernel.

### How Has This Been Tested?
insmod /lib/modules/p_lkrg.ko kint_enforce=1 ,   and this pCFI false positive issue won't occur any more.

Thanks.


solardiz:
Unfortunately, `interrupts_enabled` is defined only on a handful of archs - including ARM and Aarch64, but not including x86 - so this is failing builds.

Per the problem description, I guess the underlying issue is generic, so would need a fix for all archs? Yet somehow we were not seeing it on x86.

When considering this and possible alternatives, we also need to consider how easy or not it would be for exploits' ROP chains to fake the condition... which would let them bypass our checking of their actual current stack?

solardiz:
@lc85446 Should we expect a revision of this PR from you, so that it'd build on all of our supported archs and potentially merge-able? Thank you!

@Adam-pi3 Any comments from you maybe?

lc85446:
@solardiz  Sorry for my late reply,  I've just come back from labor day's holiday in China.  

I modified the PR  with a new wrapper function:  p_lkrg_interrupts_enabled(),  which used interrupts_enabled() for ARM and aarch64 ,  and used native_save_fl() then checked X86_EFLAGS_IF for X86 and X86_64 platform.  

Could you please review  the following patch?   Thanks.

`---
 .../exploit_detection/p_exploit_detection.c        |  8 ++++++-
 src/modules/wrap/p_struct_wrap.h                   | 28 ++++++++++++++++++++++
 2 files changed, 35 insertions(+), 1 deletion(-)

diff --git a/src/modules/exploit_detection/p_exploit_detection.c b/src/modules/exploit_detection/p_exploit_detection.c
index 4b5540d..7ec332f 100644
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -1886,8 +1886,14 @@ int p_ed_enforce_pcfi(struct task_struct *p_task, struct p_ed_process *p_orig, s
    unsigned int p_offset = 1;
    char p_sym1[KSYM_SYMBOL_LEN];
    char p_not_valid = 0;
+   unsigned long p_sp_addr;

-   if (p_ed_pcfi_validate_sp(p_task,p_orig,p_regs_get_sp(p_regs))) {
+   if (p_lkrg_interrupts_enabled(p_regs))
+      p_sp_addr = p_regs_get_sp(p_regs);
+   else
+      p_sp_addr = p_get_thread_sp(p_task);
+
+   if (p_ed_pcfi_validate_sp(p_task,p_orig,p_sp_addr)) {
       p_print_log(P_LKRG_CRIT,
              "<Exploit Detection> Stack pointer corruption (ROP?) - pCFI violation: process[%s | %d] !!!\n",
              p_task->comm,task_pid_nr(p_task));
diff --git a/src/modules/wrap/p_struct_wrap.h b/src/modules/wrap/p_struct_wrap.h
index 96c2625..be199b6 100644
--- a/src/modules/wrap/p_struct_wrap.h
+++ b/src/modules/wrap/p_struct_wrap.h
@@ -25,6 +25,11 @@
 #ifndef P_LKRG_WRAPPER_H
 #define P_LKRG_WRAPPER_H

+#include <asm/ptrace.h>
+#include <asm/irqflags.h>
+#ifdef CONFIG_X86
+#include <asm/processor-flags.h >
+#endif
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 14, 0)

 static inline void p_set_uid(kuid_t *p_arg, unsigned int p_val) {
@@ -474,6 +479,15 @@ static inline void p_lkrg_close_rw(void) {
 //   preempt_enable(); //_no_resched();
 }

+static inline bool p_lkrg_interrupts_enabled(struct pt_regs *p_regs) {
+
+   unsigned long flags = native_save_fl();
+   if ((flags & X86_EFLAGS_IF))
+      return 1;
+   else
+      return 0;
+}
+
 /* ARM */
 #elif defined(CONFIG_ARM)

@@ -606,6 +620,13 @@ static inline void p_lkrg_close_rw(void) {
    preempt_enable(); //_no_resched();
 }

+static inline bool p_lkrg_interrupts_enabled(struct pt_regs *p_regs) {
+
+   if (interrupts_enabled(p_regs))
+      return 1;
+   else
+      return 0;
+}
 /* ARM64 */
 #elif defined(CONFIG_ARM64)

@@ -760,6 +781,13 @@ static inline void p_lkrg_close_rw(void) {
    preempt_enable(); //_no_resched();
 }

+static inline bool p_lkrg_interrupts_enabled(struct pt_regs *p_regs) {
+
+   if (interrupts_enabled(p_regs))
+      return 1;
+   else
+      return 0;
+}
 #endif

 #endif
--`

solardiz:
> I modified the PR

Looks like not yet?

> with a new wrapper function: p_lkrg_interrupts_enabled(), which used interrupts_enabled() for ARM and aarch64 , and used native_save_fl() then checked X86_EFLAGS_IF for X86 and X86_64 platform.

This makes sense to me.

> Could you please review the following patch? Thanks.

I suggest that you turn the `p_lkrg_interrupts_enabled` functions into one-liner macros. So e.g. just:

```c
#define p_lkrg_interrupts_enabled interrupts_enabled
```

on the two ARM platforms, and:

```c
#define p_lkrg_interrupts_enabled(p_regs) !!(native_save_fl() & X86_EFLAGS_IF)
```

on x86.

Also, remove the extra whitespace in:

```c
#include <asm/processor-flags.h >
```

and move the:

```c
#include <asm/irqflags.h>
```

to within `#ifdef CONFIG_X86` (if I understood why/where it's needed correctly).

Other than that, the patch looks good to me - so please make these changes to the PR, and let's see if CI is happy. Thank you!

lc85446:
@solardiz ,  I will modify as your re

solardiz:
I'm sorry, I should have recalled the original issue description when reviewing the patch. It said:

> The security_capable() sometimes may be called in the interrupt context

So we should probably be directly checking for being in interrupt context or not, not for having interrupts disabled or not. And there's `in_interrupt()` for that, so just use it? Something like:

```c
   if (in_interrupt())
      p_sp_addr = p_get_thread_sp(p_task);
   else
      p_sp_addr = p_regs_get_sp(p_regs);
```

solardiz:
> failed to git push the latest patch.

You should amend the commit (probably already did locally?) and then _force_-push (`git push -f`). Without the `-f` option, amended commits wouldn't be accepted (because they replace previous commits, and thus it appears to `git` that "the remote contains work that you do not have locally", as the error message says).

lc85446:
> I'm sorry, I should have recalled the original issue description when reviewing the patch. It said:
> 
> > The security_capable() sometimes may be called in the interrupt context
> 
> So we should probably be directly checking for being in interrupt context or not, not for having interrupts disabled or not. And there's `in_interrupt()` for that, so just use it? Something like:
> 
> ```c
>    if (in_interrupt())
>       p_sp_addr = p_get_thread_sp(p_task);
>    else
>       p_sp_addr = p_regs_get_sp(p_regs);
> ```

Yes, you are right.  I tried it just now  and  it worked as expeced.
 I've thought that  some of the symbols in "in_interrupt()==>irq_count()==>preempt_count()"  were not exported and could not be used by p_lkrg.ko,  but I was wrong.

solardiz:
@lc85446 Do you have the latest patch in your branch `r` locally? If so, just switch to that branch (`git checkout r`) and force-push it without giving any other options (`git push -f`).

lc85446:
> @lc85446 Do you have the latest patch in your branch `r` locally? If so, just switch to that branch (`git checkout r`) and force-push it without giving any other options (`git push -f`).

My local branch is  'lch',   created by `git checkout -b lch remotes/for/r`
I tried  `git push -f`,  but still failed,
```
fatal: The upstream branch of your current branch does not match
the name of your current branch.  To push to the upstream branch
on the remote, use

    git push origin HEAD:r

To push to the branch of the same name on the remote, use

    git push origin lch
```

solardiz:
> Then I tried as it said,

You forgot to add `-f` to there.

lc85446:
> > Then I tried as it said,
> 
> You forgot to add `-f` to there.

Yes ,  `git push -f origin HEAD:r` works  OK.  Thank you!

solardiz:
@lc85446 As an option, please feel free to open a replacement PR from your `lch` branch. We can then merge that one (if it passes CI), and close this one.

solardiz:
No unexpected CI failures, so I'll merge. Thank you, @lc85446!

solardiz:
> The security_capable() sometimes may be called in the interrupt context

@lc85446 Can you share a backtrace showing this? Is this possibly a result of custom kernel changes?

We call our `p_ed_enforce_pcfi` from a few other hooks as well, including `generic_permission` and `mark_inode_dirty`, which are rather frequent. So our change here probably has measurable performance impact, especially given that `in_interrupt()` checks several things.

```c
#define nmi_count()     (preempt_count() & NMI_MASK)
#define hardirq_count() (preempt_count() & HARDIRQ_MASK)
#ifdef CONFIG_PREEMPT_RT
# define softirq_count()        (current->softirq_disable_cnt & SOFTIRQ_MASK)
#else
# define softirq_count()        (preempt_count() & SOFTIRQ_MASK)
#endif
#define irq_count()     (nmi_count() | hardirq_count() | softirq_count())

/*
 * Macros to retrieve the current execution context:
 *
 * in_nmi()             - We're in NMI context
 * in_hardirq()         - We're in hard IRQ context
 * in_serving_softirq() - We're in softirq context
 * in_task()            - We're in task context
 */
#define in_nmi()                (nmi_count())
#define in_hardirq()            (hardirq_count())
#define in_serving_softirq()    (softirq_count() & SOFTIRQ_OFFSET)
#define in_task()               (!(in_nmi() | in_hardirq() | in_serving_softirq()))

/*
 * The following macros are deprecated and should not be used in new code:
 * in_irq()       - Obsolete version of in_hardirq()
 * in_softirq()   - We have BH disabled, or are processing softirqs
 * in_interrupt() - We're in NMI,IRQ,SoftIRQ context or have BH disabled
 */
#define in_irq()                (hardirq_count())
#define in_softirq()            (softirq_count())
#define in_interrupt()          (irq_count())
```

BTW, `in_interrupt` is almost but not exactly the inverse of `in_task` - the latter checks a more specific SoftIRQ flag (maybe excludes the "BH disabled" case?)

We could want to use a more specific condition and/or specialize `p_ed_enforce_pcfi` for calls from different hooks (if the issue can only occur in some).

I don't suggest actually making further changes now, but I wanted to have these thoughts written down.

Issue #131, opened 2021-12-14
"page allocation failure" without CONFIG_KPROBES_ON_FTRACE


udovdh:
Hello,
Please have a look at  https://bugzilla.kernel.org/show_bug.cgi?id=214949 and the conclusion near the bottom.
Why does lkrg do that?
Is this why vm.min_free_kbytes needs to be adjusted so high? 
Please explain.

Udo

solardiz:
@udovdh Yes, I think we use `GFP_ATOMIC` too much and should fix that. Can you please try changing the two instances of `GFP_ATOMIC` below to `GFP_KERNEL`? Does this help?

```
modules/exploit_detection/p_exploit_detection.c:#define p_ed_pcfi_alloc()      kmem_cache_alloc(p_ed_pcfi_cache, GFP_ATOMIC)
modules/exploit_detection/p_rb_ed_trees/p_rb_ed_pids/p_rb_ed_pids_tree.h:#define p_alloc_ed_pids()      kmem_cache_alloc(p_ed_pids_cache, GFP_ATOMIC)
```

udovdh:
I applied the changes, next kernel patch will have the update lkrg.
Due to this vm.min_free_kbytes was increased to 770000. The issue still happens, though less frequently.
What should be sufficient on a 32GB (minus video) system for vm.min_free_kbytes with these changes?

solardiz:
@udovdh For testing these changes, please revert to the defaults (do not override `vm.min_free_kbytes`). We got to make things work with the defaults. Thank you!

udovdh:
I will test 5.15.8 with default vm.min_free_kbytes and the lkrg changes.

udovdh:
Default appears to be:
```
# sysctl -a |grep vm\.min
vm.min_free_kbytes = 67584
```

I.e.: vastly less than what was in there before.

solardiz:
@Adam-pi3 We have two instances of the below comment in `p_ed_enforce_pcfi()`:

```c
   /* Current logic always call this function with IRQ disabled */
```

If this were true, it'd be a reason to use `GFP_ATOMIC` in there. However, to me this looks false. Can you confirm that these comments are wrong or correct me if I am wrong? Also, the comments are used to claim that `local_irq_save(p_flags);` (and its corresponding restore) are not needed. Would they be needed, and why exactly, if the comments' claim is false? Luckily, the memory allocation is just before this point, so we should be able to change to `GFP_KERNEL` even if so.

[...]

PR #190, opened 2022-06-10
Reduce memory overhead generated by kmem_cache_create

Adam-pi3:
This commit brings a few important changes:
 - LKRG has used to leverage SLAB_HWCACHE_ALIGN but memory overhead
   may be too significant for LKRG's use cases
 - Since the kernel 4.5+ we can use SLAB_ACCOUNT to make sure that
   LKRG's caches are standalone
 - Modify the size of pCFI stack buffer cache to be smaller and
   decoupled from the PAGE_SIZE (there is no reason for that)

Additionally, this commit should help addressing #131

### DIFF
--- a/src/modules/exploit_detection/p_exploit_detection.h
+++ b/src/modules/exploit_detection/p_exploit_detection.h
@@ -232,9 +232,9 @@ struct p_seccomp {
 };
 
 #ifdef P_LKRG_TASK_OFF_DEBUG
- #define P_PCFI_STACK_BUF     (PAGE_SIZE >> 2)
+ #define P_PCFI_STACK_BUF     0x150
 #else
- #define P_PCFI_STACK_BUF     (PAGE_SIZE >> 1)
+ #define P_PCFI_STACK_BUF     0x200
 #endif
 
 #ifdef CONFIG_X86_64
--- a/src/p_lkrg_main.h
+++ b/src/p_lkrg_main.h
@@ -104,6 +104,20 @@
 #endif
 
 /*
+ * Define kmem_cache_create() flags:
+ *  - LKRG has used to leverage SLAB_HWCACHE_ALIGN but memory overhead
+ *    may be too significant for LKRG's use cases
+ *  - Since the kernel 4.5+ we can use SLAB_ACCOUNT to make sure
+ *    that LKRG's caches are standalone
+ */
+#if LINUX_VERSION_CODE < KERNEL_VERSION(4,5,0)
+ #define P_LKRG_CACHE_FLAGS 0
+#else
+ #define P_LKRG_CACHE_FLAGS SLAB_ACCOUNT
+#endif
+
+
+/*
  * Some custom compilation of the kernel might aggresively inline
  * critical functions (from LKRG perspective). That's problematic
  * for the project. However, some of the problems *might* be solved
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -367,7 +367,7 @@ static void p_ed_wq_valid_cache_zero(void *p_arg) {
 int p_ed_wq_valid_cache_init(void) {
 
    if ( (p_ed_wq_valid_cache = kmem_cache_create("p_ed_wq_valid_cache", sizeof(struct work_struct),
-                                                 0, SLAB_HWCACHE_ALIGN, p_ed_wq_valid_cache_zero)) == NULL) {
+                                                 0, P_LKRG_CACHE_FLAGS, p_ed_wq_valid_cache_zero)) == NULL) {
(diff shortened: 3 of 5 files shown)


Adam-pi3:
Old cache sizes:
```
p_ed_pcfi_cache - 2122
p_ed_pids - 448
p_ed_wq_valid_cache - 64
p_offload_cache - 64
```
New cache sizes:
```
p_ed_pcfi_cache - 520
p_ed_pids - 408
p_ed_wq_valid_cache - 40
p_offload_cache - 40
```

src/modules/exploit_detection/p_exploit_detection.h
```diff
@@ -232,9 +232,9 @@ struct p_seccomp {
 };
 
 #ifdef P_LKRG_TASK_OFF_DEBUG
- #define P_PCFI_STACK_BUF     (PAGE_SIZE >> 2)
+ #define P_PCFI_STACK_BUF     0x150
 #else
- #define P_PCFI_STACK_BUF     (PAGE_SIZE >> 1)
+ #define P_PCFI_STACK_BUF     0x200
 #endif
```
solardiz, line 238: What exactly happens if these turn out to be insufficient? Perhaps you also tested with smaller values to see that?

src/p_lkrg_main.h
```diff
@@ -103,6 +103,20 @@
...
+ *  - LKRG has used to leverage SLAB_HWCACHE_ALIGN but memory overhead
+ *    may be too significant for LKRG's use cases
+ *  - Since the kernel 4.5+ we can use SLAB_ACCOUNT to make sure
+ *    that LKRG's caches are standalone
+ */
+#if LINUX_VERSION_CODE < KERNEL_VERSION(4,5,0)
+ #define P_LKRG_CACHE_FLAGS 0
+#else
+ #define P_LKRG_CACHE_FLAGS SLAB_ACCOUNT
+#endif
```
solardiz, line 117: I wonder if we can/should use `#ifdef SLAB_ACCOUNT` in place of the kernel version check.

solardiz (approved):

solardiz:
> use SLAB_ACCOUNT to make sure that LKRG's caches are standalone

Do we need them to be standalone, and why?

src/modules/exploit_detection/p_exploit_detection.h
Adam-pi3: We will not validate all the addresses on the stack. On x86-64 we are validating up to 64 entries in stack depth at the time of validation. If stack depth are deeper we would only validate up to 64 entries (or 128 entries on 32bits x86).

Adam-pi3: Now I see that for CONFIG_UNWINDER_ORC we don't have this check (but for non CONFIG_UNWINDER_ORC). I will add this corner case validation for CONFIG_UNWINDER_ORC as well

src/p_lkrg_main.h
Adam-pi3: I would prefer to keep it as is since we might want to use other flags in the future.

Adam-pi3:
> > use SLAB_ACCOUNT to make sure that LKRG's caches are standalone
> 
> Do we need them to be standalone, and why?

LKRG's objects can't be used in exploitation and in case we have bug, it is difficult to exploit it.

src/modules/exploit_detection/p_exploit_detection.h
solardiz: > I will add this corner case validation for CONFIG_UNWINDER_ORC as well

Do you intend to add one more commit to this PR, or create a separate PR?

Also, ideally we'd test triggering the condition in both kinds of builds (by temporarily using deliberately way too low values of `P_PCFI_STACK_BUF`).

Adam-pi3: I've just pushed extra commit to this PR. I've tested this condition for compilation with CONFIG_UNWINDER_ORC and without it. In both cases it works fine - up to N frames from the top are validated (N is related to `P_PCFI_STACK_BUF`).

src/p_lkrg_main.h
solardiz: OK, but even for other flags I think `#ifdef FLAG_NAME` will work better (as long as the kernel defines all flags as macros) - it would enable usage of the flags precisely where available, including in distro backports, with no extra effort on our part.

src/modules/exploit_detection/p_exploit_detection.h
solardiz: The CONFIG_UNWINDER_ORC case fix looks good to me.

src/p_lkrg_main.h
solardiz: Anyway, I think I'll merge this as-is now, and we can switch to checking by flag names in a later code cleanup.

Issue #192, opened 2022-06-21
_STEXT memory block hash is different sometime between kernel 5.15.38 and 5.15.42


0xC0ncord:
It looks like sometime after kernel 5.15.38 I am seeing LKRG alert on mismatched _STEXT memory block hashes. The latest kernel that I have access to where this does not occur is 5.15.38, and the earliest kernel where I started seeing this behavior is 5.15.41.

On this particular system, I see these logs:
```
Jun 21 08:55:05 fuuko kernel: [p_lkrg] Loading LKRG...
Jun 21 08:55:06 fuuko kernel: Freezing user space processes ... (elapsed 0.001 seconds) done.
Jun 21 08:55:06 fuuko kernel: OOM killer disabled.
Jun 21 08:55:06 fuuko kernel: [p_lkrg] LKRG can't enforce SELinux validation (CONFIG_GCC_PLUGIN_RANDSTRUCT detected)
Jun 21 08:55:06 fuuko kernel: [p_lkrg] [kretprobe] register_kretprobe() for <lookup_fast> failed! [err=-2]
Jun 21 08:55:06 fuuko kernel: [p_lkrg] LKRG won't enforce pCFI validation on 'lookup_fast'
Jun 21 08:55:06 fuuko kernel: [p_lkrg] LKRG initialized successfully!
Jun 21 08:55:06 fuuko kernel: OOM killer enabled.
Jun 21 08:55:06 fuuko kernel: Restarting tasks ... done.

[...]

Jun 21 08:55:06 fuuko kernel: [p_lkrg] Enabling "blocking modules" feature.
Jun 21 08:55:06 fuuko kernel: [p_lkrg] Changing "kint_enforce" logic. From Old[2 | PANIC] to new[1 | LOG ONLY (For SELinux and CR0.WP LOG & RESTORE)] one.
Jun 21 08:55:06 fuuko kernel: [p_lkrg] Changing "pint_validate" logic. From Old[1 | CURRENT] to new[2 | CURRENT] one.
Jun 21 08:55:06 fuuko kernel: [p_lkrg] Changing "smap_enforce" logic. From Old[2 | PANIC] to new[1 | LOG & RESTORE] one.

[...]

Jun 21 08:55:06 fuuko kernel: iwlwifi 0000:03:00.0 wlo1: renamed from wlan0
Jun 21 08:55:06 fuuko kernel: r8169 0000:02:00.0 enp2s0: renamed from eth0
[...]

PR #193, opened 2022-06-26
Enable variable length jump label backport for 5.15.40+, cleanup

0xC0ncord:
### Description
Rework the version checking code for the variable length JUMP_LABEL feature and extend this version check for Linux 5.15.40+.

It seems as of right now only the 5.15.x branch received this backport.

Closes: #192

### How Has This Been Tested?
Patch was applied to LKRG 0.9.3 and built against Linux 5.15.50. System was booted in this configuration, "LKRG initialized successfully!" message was observed, and no detected STEXT memory block hash mismatches were observed.


solardiz (approved):
Looks good to me, thanks!

Issue #109, opened 2021-07-25
Don't unload LKRG at shutdown


adrelanos:
https://github.com/openwall/lkrg/blob/main/scripts/bootup/systemd/lkrg.service currently uses:

* `Before=[...] shutdown.target`
* `ExecStop=/sbin/modprobe -v -r p_lkrg`

I don't see what this would be good for? Could even be a security issue to unload LKRG?

System shutdown works well even if LKRG doesn ot get unload. If it's not needed then it's best avoided for simplicity.

Maybe related:
https://github.com/openwall/lkrg/issues/108

morfikov:
I didn't see that either, but some people want to reload the module and need the service in first place. Like they wouldn't just use the modprobe command directly. :)

adrelanos:
Seems like an insufficient rationale. For unload/reload of the kernel module it would be possible to implement this as `sudo systemctl reload lkrg`. `ExecReload=`

Related: the usefulness of the lkrg systemd unit existing at all has been in questioned here:
https://github.com/openwall/lkrg/issues/108#issuecomment-886230431

solardiz:
Our current instructions in the "Installation" and "Uninstalling and upgrading" sections of `README` rely on the current behavior, and `make uninstall` would currently unload LKRG via:

```shell
elif [ "$1" == "uninstall" ]; then
	echo -e "       ${P_GREEN}Stopping ${P_YL}lkrg.service${P_NC}"
	systemctl stop lkrg.service
```

So we should not simply drop these lines from `scripts/bootup/systemd/lkrg.service` without also revising the instructions and perhaps the above script.

Adam-pi3:
@solardiz @adrelanos @morfikov should we close this issue?

morfikov:
Probably.

solardiz:
I think let's close for now. We can revisit later. We're considering additions to LKRG that could affect the decision-making here, so deciding on this issue now is premature.

mrl5:
what was the motivation behind creating dedicated systemd service for lkrg instead of relying on https://www.freedesktop.org/software/systemd/man/modules-load.d.html?

my point here is that it turns out that with dedicated systemd/OpenRC service one must remember about few important things like load order as well as mentioned side effects of service being stopped - maybe it's over-engineering?

solardiz:
[...]

PR #197, opened 2022-07-03
feat(scripts): support OpenRC init system

mrl5:
### Description
introduce OpenRC bootup scripts

### How Has This Been Tested?
I've tested it by running
```console
# make uninstall
# lsmod | grep lkrg
```
```console
# make install &&
  /etc/init.d/lkrg start &&
  lsmod | grep lkrg &&
  /etc/init.d/lkrg status &&
  rc-update add lkrg boot &&
  rc-update show | grep lkrg
```
```console
# make uninstall
# lsmod | grep lkrg
# /etc/init.d/lkrg status
# rc-update show | grep lkrg
```
on OpenRC system ([funtoo linux](https://www.funtoo.org))

**update**
after gathering requirements in [this comment](https://github.com/lkrg-org/lkrg/pull/197#issuecomment-1175588110) I've additionally run
```console
# runlevel
N 3
# /etc/init.d/lkrg stop
# lsmod | grep lkrg;  echo $?
1
# /etc/init.d/lkrg start
# lsmod | grep lkr
p_lkrg                221184  0
```

```console
# init 1
# lsmod | grep lkr
p_lkrg                221184  0
```

and ensured that `p_lkrg` module is not unloaded from the kernel on poweroff or reboot

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,8 @@
+The following major changes have been made since LKRG 0.9.3:
+
+ *) OpenRC init system support
+
+
 The following major changes have been made between LKRG 0.9.2 and 0.9.3:
 
  *) Support new mainline kernels 5.17.x, 5.18-rc*, and hopefully beyond
--- a/README
+++ b/README
@@ -119,48 +119,80 @@ bootup and without the parameter override.
 Installation
 ------------
 
-If your Linux distribution uses systemd, you can install LKRG with:
+If your Linux distribution uses a supported init system (systemd or OpenRC),
+you can install LKRG with:
 
 	sudo make install
 
 while you're still in its top level source code directory.
 
-Run the following command to start the LKRG service just like it would be
-started on next system bootup:
+We don't in any way favor one init system over another, and would gladly add
+support for them as well if there's demand, or especially if we receive such
+contributions.  Meanwhile, on a distribution without a supported init system
+you can let "sudo make install" partially complete (up to the point where it
+finds you're not using a supported init system).
+
+Run the following command to start the LKRG service, for systemd:
 
 	sudo systemctl start lkrg
+
+for OpenRC:
+
+	sudo /etc/init.d/lkrg start
+
+for other:
+
+	sudo modprobe -v p_lkrg
+
+
+Autoload on bootup
+------------------
+
+In order to automatically load LKRG into the Linux kernel on each bootup run
+the following command, for systemd:
+
+	sudo systemctl enable lkrg
+
+for OpenRC:
+
+	sudo rc-update add lkrg boot
+
+for other:
+
+	sudo mkdir -p /etc/modules-load.d/ &&
+		echo p_lkrg | sudo tee /etc/modules-load.d/p_lkrg.conf
+
+Alternatively you can put the "modprobe p_lkrg" command into a system
+startup script.  Please note that ideally this command would run before sysctl
+files (especially /etc/sysctl.d/01-lkrg.conf) are processed, or otherwise the
+LKRG settings specified in those would not take effect.
+
+
+Uninstalling
+------------
 
 You can uninstall LKRG using "make" as well (still in the same directory):
 
 	sudo make uninstall
 
 You can also use the following command to temporarily stop the LKRG service
-without uninstalling it:
+without uninstalling it, for systemd:
 
 	sudo systemctl stop lkrg
 
-We don't in any way favor systemd over other init systems, and would gladly add
-support for those as well if there's demand, or especially if we receive such
-contributions.  Meanwhile, on a system without systemd you can let "sudo make
-install" partially complete (up to the point where it finds you're not using
-systemd) and then use:
+for OpenRC:
 
-	sudo modprobe p_lkrg
+	sudo /etc/init.d/lkrg stop
 
-to load the module.  You can also put the "modprobe p_lkrg" command into a
-system startup script.
+for other:
+
+	sudo modprobe -v -r p_lkrg
 
 
-Uninstalling and upgrading
---------------------------
+Upgrading
+---------
 
-Our suggested way to upgrade LKRG is to start by uninstalling the old version,
-which you can do with:
-
-	sudo make uninstall
-
-from the top level source code directory of that old version (doing so using
-the source code tree of the new version may or may not work correctly).
+Our suggested way to upgrade LKRG is to start by uninstalling the old version.
 
 You can then follow the Testing and Installation steps for the new version.
 
@@ -171,11 +203,12 @@ Recovery
 To account for the hopefully unlikely, but really unfortunate event that some
 incompatibility between the Linux kernel or other components of the system and
 LKRG isn't detected prior to LKRG installation, yet leads to system crash on
-bootup, we've included support for the "nolkrg" kernel parameter in the systemd
-unit file for LKRG.  Thus, if you've followed the above installation procedure
-for LKRG with systemd, you may disable LKRG by specifying "nolkrg" on the
-kernel command-line via your bootloader.  The system should then boot up
-without LKRG, and thus without triggering the problem, letting you fix it.
+bootup, we've included support for the "nolkrg" kernel parameter.  Thus, you
+may disable LKRG by specifying "nolkrg" on the kernel command-line via your
+bootloader.  The system should then boot up without LKRG, and thus without
+triggering the problem, letting you fix it.  You must be aware though, that you
+will not be able to manually load the LKRG module if the kernel was booted with
+this parameter.
 
 
 Module parameters
--- a/scripts/bootup/openrc/lkrg
+++ b/scripts/bootup/openrc/lkrg
@@ -1,0 +1,31 @@
+#!/sbin/openrc-run
+# Distributed under the terms of the GNU General Public License v2
+
+description="Linux Kernel Runtime Guard"
+command="/sbin/modprobe"
+command_args="-v p_lkrg"
+unload_command_args="-v -r p_lkrg"
+runlevel_command="/sbin/runlevel"
+grep_command="/bin/grep"
+
+depend() {
+	before sysctl modules
+}
+
+start() {
+	ebegin "Starting \"lkrg\" service"
+	$command $command_args
+	eend $?
+}
+
+stop() {
+	# for more info about runlevels (e.g. "S", "0", "1", "6") check: man 8 init
+	if $runlevel_command | $grep_command -q ' [S016]$'; then
+		ebegin "Stopping \"lkrg\" service without unloading \"p_lkrg\" module from the kernel"
+		eend $?
+	else
+		ebegin "Unloading \"p_lkrg\" module from the kernel"
+		$command $unload_command_args
+		eend $?
+	fi
+}
--- a/scripts/bootup/openrc/lkrg-openrc.sh
+++ b/scripts/bootup/openrc/lkrg-openrc.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+#
+# OpenRC installation script for LKRG (main branch)
+#
+# Author:
+#  - Adam 'pi3' Zabrocki (http://pi3.com.pl)
+#  - Jakub 'mrl5' Kołodziejczak (https://github.com/mrl5)
+##
+
+P_SYSCTL_DIR="/etc/sysctl.d"
+P_INITD_DIR="/etc/init.d"
+RUNLEVEL="boot"
+
+
+P_RED='\033[0;31m'
+P_GREEN='\033[0;32m'
+P_WHITE='\033[1;37m'
+P_YL='\033[1;33m'
+P_NC='\033[0m' # No Color
+
+echo -e "  ${P_GREEN}[+] ${P_WHITE}OpenRC detected${P_NC}"
+
+if [ "$1" == "install" ]; then
+	if [ -e "$P_INITD_DIR/lkrg" ]; then
+		echo -e "       ${P_RED}ERROR! ${P_YL}lkrg${P_RED} already exists under ${P_YL}$P_INITD_DIR${P_RED} directory${P_NC}"
+		exit 1
+	else
+		echo -e "       ${P_GREEN}Installing ${P_YL}lkrg${P_GREEN} file under ${P_YL}$P_INITD_DIR${P_GREEN} directory${P_NC}"
+		install -pm 755 -o root -g root scripts/bootup/openrc/lkrg "$P_INITD_DIR/lkrg"
+		echo -e "       ${P_GREEN}To start ${P_YL}lkrg${P_GREEN} please use: ${P_YL}/etc/init.d/lkrg start${P_NC}"
+		echo -e "       ${P_GREEN}To enable ${P_YL}lkrg${P_GREEN} on bootup please use: ${P_YL}rc-update add lkrg ${RUNLEVEL}${P_NC}"
+	fi
+	if [ -e "$P_SYSCTL_DIR/01-lkrg.conf" ]; then
+		echo -e "       ${P_YL}01-lkrg.conf${P_GREEN} is already installed, skipping${P_NC}"
+	else
+		echo -e "       ${P_GREEN}Installing ${P_YL}01-lkrg.conf${P_GREEN} file under ${P_YL}$P_SYSCTL_DIR${P_GREEN} directory${P_NC}"
+		install -pm 644 -o root -g root scripts/bootup/lkrg.conf "$P_SYSCTL_DIR/01-lkrg.conf"
+	fi
+elif [ "$1" == "uninstall" ]; then
+	echo -e "       ${P_GREEN}Stopping ${P_YL}lkrg${P_NC}"
+	/etc/init.d/lkrg stop
+	echo -e "       ${P_GREEN}Disabling ${P_YL}lkrg${P_GREEN} on bootup${P_NC}"
+	rc-update del lkrg ${RUNLEVEL}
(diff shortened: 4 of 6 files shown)


README
```diff
@@ -116,39 +116,92 @@ so that it can then be loaded using the same procedure that's used on system
...
+
+	sudo mkdir -p /etc/modules-load.d/ &&
+		echo p_lkrg > /etc/modules-load.d/p_lkrg.conf
+
+Alternatively you can put the "modprobe p_lkrg" command into a system
+startup script.
+
+
+Uninstalling
+------------
```
solardiz, line 185: You'll need to rename and revise the "Uninstalling and upgrading" section then - it should probably become just "Upgrading" and should refer to the section "Uninstalling" instead of partially repeating its content.

README
```diff
@@ -116,39 +116,92 @@ so that it can then be loaded using the same procedure that's used on system
...
+
 You can uninstall LKRG using "make" as well (still in the same directory):
 
 	sudo make uninstall
 
 You can also use the following command to temporarily stop the LKRG service
 without uninstalling it:
 
+for systemd:
```
solardiz, line 194: Style: you don't follow these with colons in your additions to other sections, so either should not do so here or should revise all of them. I'd do the latter, e.g.:

```
You can also use the following command to temporarily stop the LKRG service
without uninstalling it, for systemd:
```

and then it's natural to have the rest just like you wrote them here in `Uninstalling`.

README
```diff
@@ -116,39 +116,92 @@ so that it can then be loaded using the same procedure that's used on system
...
+If your Linux distribution uses supported init system, you can install LKRG
+with:
 
 	sudo make install
 
-while you're still in its top level source code directory.
+while you're still in its top level source code directory. Meanwhile, on a
+system without supported init systems you can let "sudo make install" partially
+complete (up to the point where it finds you're not using supported init
+system).
```
solardiz, line 144: Should be "a supported init system" (add "a ") in both places. Also, two spaces before "Meanwhile" for consistent style with the rest of the file.

README
```diff
@@ -116,39 +116,92 @@ so that it can then be loaded using the same procedure that's used on system
...
+for other
+
+	sudo modprobe -v p_lkrg
+
+
+Autoload on bootup
+------------------
+
+In order to automatically add LKRG module into Linux Kernel on each bootup run
+following command:
```
solardiz, line 165: Please replace this sentence with:

```
In order to automatically load LKRG into the Linux kernel on each bootup run
the following command:
```

scripts/bootup/openrc/lkrg
```diff
@@ -0,0 +1,17 @@
+#!/sbin/openrc-run
+# Distributed under the terms of the GNU General Public License v2
```
solardiz, line 2: Why did you feel it important to specify license in this one tiny file and not in others? We do have top-level `LICENSE` specify GPLv2, so this feels unneeded (would have made sense if you specified a more relaxed license here).

README
```diff
@@ -177,6 +230,10 @@ for LKRG with systemd, you may disable LKRG by specifying "nolkrg" on the
 kernel command-line via your bootloader.  The system should then boot up
 without LKRG, and thus without triggering the problem, letting you fix it.
 
+Alternative way (e.g. with OpenRC) is to use "module_blacklist=p_lkrg" kernel
+parameter. You must be aware though, that you will not be able to manually load
+the LKRG module if kernel was booted with this paramter.
```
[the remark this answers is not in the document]
mrl5: yup, I've checked and confirm that `p_lkrg` doesn't want to load with this option :)

solardiz:
@mrl5 Also, you seem to have a GitHub-provided noreply e-mail address on the commit. You might want to change that to an address of your own, which you'd then add to your GitHub account.

solardiz:
@mrl5 Is there any way for me to contact you about the misunderstandings I see in comments at https://bugs.funtoo.org/browse/FL-10115? That would be off-topic to discuss on this pull request.

mrl5:
> @mrl5 Is there any way for me to contact you about the misunderstandings I see in comments at https://bugs.funtoo.org/browse/FL-10115? That would be off-topic to discuss on this pull request.

I've sent PM to your twitter account: https://twitter.com/solardiz

README
mrl5: > Please replace this sentence with:
> 
> ```
> In order to automatically load LKRG into the Linux kernel on each bootup run
> the following command:
> ```

ok will do. for the record I've taken `add` from `man modprobe` :D
>       modprobe - Add and remove modules from the Linux Kernel

scripts/bootup/openrc/lkrg
mrl5: that's a good question - actually I did it per analogy to most of the scripts that are in `/etc/init.d/*` on my OS:
```console
$ head -n 3 /etc/init.d/* | grep -i license | wc -l
61
$ ls -l /etc/init.d | wc -l
106
```
but I can remove this line if you really want to

CHANGES
```diff
@@ -1,3 +1,7 @@
+The following major changes have been made between LKRG 0.9.3 and 0.9.4:
```
solardiz, line 1: We generally only include the next version number in `CHANGES` shortly before making the release, and until then use something like "since LKRG 0.9.3".

README
```diff
@@ -116,49 +116,100 @@ so that it can then be loaded using the same procedure that's used on system
...
 
 
+Supported init systems
+----------------------
+
+Currently there are dedicated scripts for:
+
+- systemd
+
+- OpenRC
```
solardiz, line 126: I'd shorten this to one line:

```
Currently there are dedicated scripts for systemd and OpenRC.
```

README
```diff
@@ -116,49 +116,100 @@ so that it can then be loaded using the same procedure that's used on system
...
 
-If your Linux distribution uses systemd, you can install LKRG with:
+If your Linux distribution uses a supported init system, you can install LKRG
+with:
 
 	sudo make install
 
-while you're still in its top level source code directory.
+while you're still in its top level source code directory.  Meanwhile, on a
+OS without a supported init system you can let "sudo make install" partially
```
solardiz, line 144: Should be "an" before "OS", or use "distribution" in place of "OS". (I understand you don't want to write "system" referring to two different things in the same sentence.)

README
```diff
@@ -116,49 +116,100 @@ so that it can then be loaded using the same procedure that's used on system
...
+complete (up to the point where it finds you're not using a supported init
+system).
+
+Run the following command to start the LKRG service:
 
-Run the following command to start the LKRG service just like it would be
-started on next system bootup:
+- for systemd:
```
solardiz, line 150: I'm fine with this style, but I'd save two lines by writing:

```
Run the following command to start the LKRG service, for systemd:
```

and then omit the leading "- " before "for OpenRC" and "for other".

Ditto in other places, so making the text significantly shorter.

README
```diff
@@ -116,49 +116,100 @@ so that it can then be loaded using the same procedure that's used on system
...
-Our suggested way to upgrade LKRG is to start by uninstalling the old version,
-which you can do with:
-
-	sudo make uninstall
+Upgrading
+---------
 
+Our suggested way to upgrade LKRG is to start by uninstalling the old version,
 from the top level source code directory of that old version (doing so using
 the source code tree of the new version may or may not work correctly).
```
solardiz, line 214: I think this paragraph about the directory should stay in "Uninstalling", not be in "Upgrading", which should refer to "Uninstalling" for it.

README
```diff
@@ -171,11 +222,12 @@ Recovery
...
-unit file for LKRG.  Thus, if you've followed the above installation procedure
-for LKRG with systemd, you may disable LKRG by specifying "nolkrg" on the
-kernel command-line via your bootloader.  The system should then boot up
-without LKRG, and thus without triggering the problem, letting you fix it.
+bootup, we've included support for the "nolkrg" kernel parameter.  Thus, you
+may disable LKRG by specifying "nolkrg" on the kernel command-line via your
+bootloader.  The system should then boot up without LKRG, and thus without
+triggering the problem, letting you fix it.  You must be aware though, that you
+will not be able to manually load the LKRG module if kernel was booted with
+this paramter.
```
solardiz, line 230: Add "the" before "kernel", fix typo in "paramter".

scripts/bootup/openrc/lkrg
solardiz: OK, up to you.

solardiz (changes requested):
Looks almost ready.

solardiz:
@mrl5 BTW, for the sysctl file to provide (almost-)initial LKRG configuration, LKRG needs to be loaded _before_ that file is processed. However, I don't see anything in the init script that would indicate its priority vs. other scripts. Did you look into that?

It's also desirable for security that LKRG be loaded before services such as sshd, cron, MTA/MDA, and unloaded late on system shutdown (if at all).

On pre-systemd Red Hat'ish systems with `chkconfig`, this would be controlled with lines like `# chkconfig: 2345 55 25` in init files (comments for the shell, but magic for `chkconfig`). These indicated runlevels (`2345` here) and start and stop priorities vs. other init scripts (`55` and `25` here). Does OpenRC have anything similar? Does Funtoo use that?

Quite possibly we'll eventually drop unloading of LKRG on system shutdown somehow. This is currently mostly a side-effect of our uninstall/upgrade instructions.

mrl5:
> @mrl5 BTW, for the sysctl file to provide (almost-)initial LKRG configuration, LKRG needs to be loaded _before_ that file is processed (edit: otherwise those settings would not take effect at all). However, I don't see anything in the init script that would indicate its priority vs. other scripts. Did you look into that?
> 
> It's also desirable for security that LKRG be loaded before services such as sshd, cron, MTA/MDA, and unloaded late on system shutdown (if at all).
> 
> On pre-systemd Red Hat'ish systems with `chkconfig`, this would be controlled with lines like `# chkconfig: 2345 55 25` in init files (comments for the shell, but magic for `chkconfig`). These indicated runlevels (`2345` here) and start and stop priorities vs. other init scripts (`55` and `25` here). Does OpenRC have anything similar? Does Funtoo use that?
> 
> Quite possibly we'll eventually drop unloading of LKRG on system shutdown somehow. This is currently mostly a side-effect of our uninstall/upgrade instructions.

thanks for bringing this up. it actually made me to take a step back and ask this question: https://github.com/lkrg-org/lkrg/issues/109#issuecomment-1175514381

TBH the only reason why I decided to do this pull request was to make OpenRC service per analogy to the existing systemd one. But now I'm really thinking to give up this idea and simply fall back to [modules-load.d](https://www.freedesktop.org/software/systemd/man/modules-load.d.html).

I'm pretty sure OpenRC takes care of correct loading order when [modules-load.d](https://www.freedesktop.org/software/systemd/man/modules-load.d.html) is detected (like mentioned first sysctl file then modules as well as modules before other services). Additionally I don't think that it would unload the modules before shutdown.

If you agree with above statement I'd be happy to modify `README` to include some instructions for [modules-load.d](https://www.freedesktop.org/software/systemd/man/modules-load.d.html)

solardiz:
> I'm pretty sure OpenRC takes care of correct loading order when [modules-load.d](https://www.freedesktop.org/software/systemd/man/modules-load.d.html) is detected

What makes you so sure?

> like mentioned first sysctl file then modules

For LKRG, it should be the other way around! The sysctl's can't be applied before the module is loaded, they simply don't exist yet. For some other module, it could be different - maybe that other module does not have its own sysctl's, but its loading is affected by the kernel's sysctl settings. So there isn't a universal "correct" order that an init system could use.

Anyway, longer term we should probably move away from usage of sysctl's for initial configuration. We intended module parameters to be used for that, but then got contributed this file with all sysctl's based on our `README`, and I understand it's more convenient to edit than a list of module parameters all on one line, but it's also less reliable and is inherently racy.

> I'd be happy to modify README to include some instructions for [modules-load.d](https://www.freedesktop.org/software/systemd/man/modules-load.d.html)

Then we also need instructions for passing parameters to modules loaded that way.

Since you started this PR with an almost ready contribution, I thought we could accept it for now, then _possibly_ drop both systemd and OpenRC specifics at some later time.

mrl5:
> What makes you so sure?
It's an assumption :D sorry - it might sound misleading

> Since you started this PR with an almost ready contribution, I thought we could accept it for now, then possibly drop both systemd and OpenRC specifics at some later time.
very well then - let me go through 2nd iteration of your CR so that I can apply fixes. I'll click on `re-request review` once I think I'm ready

mrl5:
@solardiz for my convenience I decided to gather requirements for proper way of handling `p_lkrg` module by the init system dedicated service (I'm using [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) nomenclature):
1. it MUST load `p_lkrg` module before sysctl config files are processed (esp. `/etc/sysctl.d/01-lkrg.conf`)
2. it MUST load `p_lkrg` module as early after bootup as possible (esp. before services such as sshd, cron, MTA/MDA are started)
3. it MUST unload `p_lkrg` module on explicit user command
4. it SHOULD NOT unload `p_lkrg` module automatically at shutdown
5. it MAY not load `p_lkrg` module if `nolkrg` kernel parameter was provided

pls suggest changes to above if I missed sth or misunderstood sth.

solardiz:
@mrl5 You got those requirements (1, 2, 3) and preferences (4, 5) right. The changes you suggest in "update" are also within consideration. Thanks!

scripts/bootup/openrc/lkrg
```diff
@@ -0,0 +1,31 @@
...
+	ewarn "This command doesn't unload \"p_lkrg\" module from the kernel!"
+	# it SHOULD unload `p_lkrg` module on explicit user command
+	einfo "To unload \"p_lkrg\" module use \"unload\" target"
+}
+
+unload() {
+	ebegin "Unloading \"p_lkrg\" module from the kernel"
+	$command $unload_command_args
+	eend $?
+}
```
solardiz, line 31: I'm OK with this approach, although I thought you could probably check `runlevel` from within `stop`. Is there a working `runlevel` command on systems with OpenRC?

scripts/bootup/openrc/lkrg-openrc.sh
```diff
@@ -0,0 +1,60 @@
...
+	fi
+	if [ -e "$P_SYSCTL_DIR/01-lkrg.conf" ]; then
+		echo -e "       ${P_YL}01-lkrg.conf${P_GREEN} is already installed, skipping${P_NC}"
+	else
+		echo -e "       ${P_GREEN}Installing ${P_YL}01-lkrg.conf${P_GREEN} file under ${P_YL}$P_SYSCTL_DIR${P_GREEN} directory${P_NC}"
+		install -pm 644 -o root -g root scripts/bootup/lkrg.conf "$P_SYSCTL_DIR/01-lkrg.conf"
+	fi
+elif [ "$1" == "uninstall" ]; then
+	echo -e "       ${P_GREEN}Stopping ${P_YL}lkrg${P_NC}"
+	/etc/init.d/lkrg stop
```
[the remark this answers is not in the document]
mrl5: I've managed to move everything to `stop` command keeping [the requirements](https://github.com/lkrg-org/lkrg/pull/197#issuecomment-1175588110)

thanks for the `runlevel` hint in this comment https://github.com/lkrg-org/lkrg/pull/197#discussion_r916613571

scripts/bootup/openrc/lkrg
```diff
@@ -0,0 +1,30 @@
...
+
+start() {
+	ebegin "Loading \"p_lkrg\" module to the kernel"
+	$command $command_args
+	eend $?
+}
+
+stop() {
+	# for more info about runlevels (e.g. "S", "0", "1", "6") check: man 8 init
+	if [[ $($runlevel_command) =~ (S|0|1|6)$ ]]; then
```
solardiz, line 22: Maybe add a space before each runlevel value we check, so that we'd only match on the second token in `runlevel` output? At least for me, `runlevel` outputs two tokens - previous and current runlevels. Here, the previous can potentially be single-user mode, from which the sysadmin might have exited into multi-user. We probably do want to load LKRG in that case.

I mean:

```bash
	if [[ $($runlevel_command) =~ ( S| 0| 1| 6)$ ]]; then
```

Of course, you need to re-test it.

BTW, is it safe to assume the shell is `bash` here?

solardiz (approved):
This looks ready to me. I just raised one minor concern - please feel free to make an adjustment (and re-test) or let me know that you're leaving things as they are (and I'll merge as-is). Thanks!

solardiz:
BTW, note that this PR looks like it'll clash with #200. I hope you don't mind rebasing if we happen to merge that other one first.

mrl5:
> BTW, note that this PR looks like it'll clash with #200. I hope you don't mind rebasing if we happen to merge that other one first.

I don't mind rebasing in case of git conflict :) I hope you don't mind reviewing one more time in case this happens :D

scripts/bootup/openrc/lkrg
mrl5: > Maybe add a space before each runlevel value we check, so that we'd only match on the second token in `runlevel` output? At least for me, `runlevel` outputs two tokens - previous and current runlevels. Here, the previous can potentially be single-user mode, from which the sysadmin might have exited into multi-user. We probably do want to load LKRG in that case.

why not - I'll do it. For the record though, my understanding of `$` regex anchor, `man 8 init` and `man 8 runlevel` is that there shouldn't be any bug with current regex

> BTW, is it safe to assume the shell is bash here?

good point - better safe than sorry - actually I'm not sure how `#!/sbin/openrc-run` works under the hood

solardiz:
Looks good to me. I'll merge now. Thank you!

PR #200, opened 2022-07-06
Added instructions on how to install lkrg using DKMS

Krish-sysadmin:
### Description
Added instructions on how to install lkrg using DKMS

### How Has This Been Tested?

Tested on Fedora 36 switching between kernels

### DIFF
--- a/README
+++ b/README
@@ -168,6 +168,34 @@ files (especially /etc/sysctl.d/01-lkrg.conf) are processed, or otherwise the
 LKRG settings specified in those would not take effect.
 
 
+Installing using DKMS
+---------------------
+
+DKMS enables kernel modules to be dynamically built for each kernel version.  What
+this means in effect is that on kernel upgrades the module is rebuilt.  You can
+install LKRG using DKMS as well.  For instance on Red Hat'ish distributions after
+following the shared download instructions above:
+
+  sudo tar -xzf lkrg-0.9.3.tar.gz -C /usr/src/ 
+  sudo dnf update -y 
+  sudo dnf install kernel-devel dkms openssl
+  sudo dkms add -m lkrg -v 0.9.3
+  sudo dkms build -m lkrg -v 0.9.3 
+  sudo dkms install -m lkrg -v 0.9.3 
+
+The only difference on other distributions should be the installation of the kernel
+headers, the DKMS utility, and OpenSSL.  Install the headers for the target kernels. 
+
+If this succeeds you should get similar output to the following:
+  
+  $ dkms status
+  lkrg/0.9.3, 5.18.9-200.fc36.x86_64, x86_64: installed
+
+Please refer to the previous two sections for how to start the LKRG service or
+have it started on system bootup.  If you wish to use the unit/init file you
+must install it manually.
+
+
 Uninstalling
 ------------
 


Krish-sysadmin:
modprobe p_lkrg and rmmod p_lkrg work as expected. I believe we should let the users know that the systemd service file is not installed automatically when using dkms. I am not aware of a way to do that, maybe there is? 

In any case, it should be noted that the user should manually install the systemd service file so that that works

README
```diff
@@ -151,6 +151,32 @@ to load the module.  You can also put the "modprobe p_lkrg" command into a
 system startup script.
 
 
+Installing using DKMS
+-------------------------
+
+DKMS enables kernel modules to be dynamically built for each kernel version. What
+this means in effect is that on kernel upgrades the module is rebuilt. You can
+install LKRG using DKMS as well. For instance on RHEL-based distributions:
```
solardiz, line 159: This text will need to be formatted more similarly to the rest of the `README` - proper count of dashes, two spaces between sentences. Also, "RHEL-based" isn't exactly right; we use "Red Hat'ish" above, so perhaps let's just repeat that.

solardiz, line 159: This text will need to be formatted more similarly to the rest of the README - proper count of dashes, two spaces between sentences. Also, "RHEL-based" isn't exactly right; we use "Red Hat'ish" above, so perhaps let's just repeat that.

README
```diff
@@ -151,6 +151,32 @@ to load the module.  You can also put the "modprobe p_lkrg" command into a
...
+
+	sudo wget https://lkrg.org/download/lkrg-0.9.3.tar.gz
```
solardiz, line 161: Let's not repeat the download instructions here. Besides, this shouldn't be done as root (no `sudo`) and ideally the signature should be verified. So we should either refer to the shared download instructions above or just assume the file has been downloaded here.

README
```diff
@@ -151,6 +151,32 @@ to load the module.  You can also put the "modprobe p_lkrg" command into a
...
+	tar -xf lkrg-0.9.3.tar.gz -C /usr/src/
```
solardiz, line 162: Now this command probably requires root, to be able to write into `/usr/src`.

I'd use explicit `xzf`, although modern GNU `tar` does detect compression use and type on its own.

README
```diff
@@ -151,6 +151,32 @@ to load the module.  You can also put the "modprobe p_lkrg" command into a
...
+	sudo wget https://lkrg.org/download/lkrg-0.9.3.tar.gz
+	tar -xf lkrg-0.9.3.tar.gz -C /usr/src/ 
+	sudo dnf update -y 
+	sudo dnf install kernel-devel-$(uname -r) dkms openssl
+	sudo dkms add -m lkrg -v 0.9.3
+	sudo dkms build -m lkrg -v 0.9.3 
+	sudo dkms install -m lkrg -v 0.9.3 
+
+The only difference on other distributions should be the installation of the kernel
+headers, the dkms utility and openssl. Install the headers for the target kernels.
```
solardiz, line 170: BTW, we omit the `-$(uname -r)` in similar instructions for Red Hat'ish distributions above, and it seemed to work fine without that? We do include that for Debian/Ubuntu. I don't know/recall if it's due do some difference between those distros, or just us being weird.

Why does this need `openssl`?

Oxford comma before "and openssl", please. ;-)

README
```diff
@@ -151,6 +151,32 @@ to load the module.  You can also put the "modprobe p_lkrg" command into a
...
+headers, the dkms utility and openssl. Install the headers for the target kernels. 
+
+If this succeeds you should get similar output to the following:
+	
+	$ dkms status
+	lkrg/0.9.3, 5.18.9-200.fc36.x86_64, x86_64: installed
+
+
+
+
```
solardiz, line 179: Should have two empty lines here, not four.

solardiz (changes requested):
Please also merge the two commits into one (and force-push it), and set a descriptive commit message - not "Update README", but e.g. "README: Add instructions on installing using DKMS".

solardiz:
@morfikov Please take a look at this PR, and comment on @Krish-sysadmin's comment above. Thank you!

README
Krish-sysadmin: DKMS uses OpenSSL by default but doesn't seem to install it as a dependency

morfikov: For standard kernel module building via DKMS, there's no need for OpenSSL. When a user uses module signing (for instance in oder to satisfy EFI SecureBoot), then the package is needed to make such signatures.

Krish-sysadmin: > When a user uses module signing (for instance in oder to satisfy EFI SecureBoot), then the package is needed to make such signatures

Yup, that's my setup. EFI SecureBoot. But it's not harmful to install openssl too

Krish-sysadmin:
@morfikov @solardiz I made changes to the README. Please have a look. If it is ok, then I'll merge the two commits into one

solardiz (changes requested):
Please correct the word "downloaded" to "download", introduce double spaces between sentences for consistency with the rest of the text, and squash the two commits into one. Other than that, this looks ready unless @morfikov says otherwise. Thanks!

Krish-sysadmin:
>  introduce double spaces between sentences for consistency with the rest of the text

I don't see double spaces in the rest so where is this?

solardiz:
> I don't see double spaces in the rest so where is this?

Everywhere else between sentences within a paragraph in `README`, e.g. in this one before "We" if you need a specific example:

```
LKRG is a kernel module (not a kernel patch), so it can be built for and loaded
on top of a wide range of mainline and distros' kernels, without needing to
patch those.  We currently support kernel versions ranging from as far back as
RHEL7's (and its many clones/revisions) and Ubuntu 16.04's to latest mainline
and distros' kernels.  We've tested this revision of LKRG with Linux kernels
up to and including 5.18-rc*.
```

This style is becoming increasingly archaic, though, so perhaps we'll need to switch away from it - but separately, not in this PR. Maybe along with a switch to Markdown.

morfikov:
> Please correct the word "downloaded" to "download", introduce double spaces between sentences for consistency with the rest of the text, and squash the two commits into one. Other than that, this looks ready unless @morfikov says otherwise. Thanks!

It looks OK to me.

Krish-sysadmin:
Thank you @morfikov @solardiz . I fixed it. By double spaces I thought you meant double-spaces between the lines itself. Thanks for clarifying

Krish-sysadmin:
If it looks OK, I'll merge the commits

solardiz (commented):
Looks good to me. Please squash the commits and this is ready.

Please note that this PR will clash with another concurrent one. So whichever one of these two is to be merged second will need to be rebased first.

Krish-sysadmin:
Done @solardiz

solardiz:
> it should be noted that the user should manually install the systemd service file

We forgot to address this. And it's not only systemd, but configuring LKRG to be loaded on bootup in any way in general. The other PR that's about to be merged introduces OpenRC and also instructions for "other" init systems, so we'll need to have whatever we write here fit in with that.

Krish-sysadmin:
Ok, so we need to wait for that to be completed?

solardiz:
I've just merged that other PR. As expected, this caused a conflict here (sorry!), so you need to rebase and also write something about configuring LKRG to be loaded on bootup.

Krish-sysadmin:
@solardiz Ok, so how do I do this because when I rebase it it gets complicated because there are so many changes to the README in that commit

Krish-sysadmin:
I figured something out. Is it good now?

README
```diff
@@ -168,11 +168,37 @@ files (especially /etc/sysctl.d/01-lkrg.conf) are processed, or otherwise the
 LKRG settings specified in those would not take effect.
 
 
+Installing using DKMS
+----------------------
```
solardiz, line 172: Extra `-` here.

README
```diff
@@ -168,11 +168,37 @@ files (especially /etc/sysctl.d/01-lkrg.conf) are processed, or otherwise the
...
+The only difference on other distributions should be the installation of the kernel
+headers, the DKMS utility, and OpenSSL.  Install the headers for the target kernels. 
+
+If this succeeds you should get similar output to the following:
+  
+  $ dkms status
+  lkrg/0.9.3, 5.18.9-200.fc36.x86_64, x86_64: installed
+
+In order to start the LKRG service please refer to the 'Installation' section and in order
+to autoload it on bootup please refer to the 'Autoload on bootup' section.
```
solardiz, line 195: This doesn't mention how the unit/init file would be installed, or maybe implies that `make install` would be used - is it compatible with these DKMS instructions?

The "In order to ..." line is over-long. We wrap text at 79 characters per line in this file. Maybe replace this paragraph with:

```
Please refer to the previous two sections for how to start the LKRG service or
have it started on system bootup.
```

README
```diff
@@ -168,11 +168,37 @@ files (especially /etc/sysctl.d/01-lkrg.conf) are processed, or otherwise the
...
+
+
 Uninstalling
 ------------
 
 You can uninstall LKRG using "make" as well (still in the same directory):
-
```
solardiz, line 175: Don't remove the empty line here.

README
```diff
@@ -190,7 +216,7 @@ for other:
 
 
 Upgrading
----------
+----------
```
solardiz, line 219: Don't add the extra `-` here.

solardiz (changes requested):
Your rebasing looks OK, but somehow you're making a couple of inadvertent changes elsewhere in `README`.

Krish-sysadmin:
@solardiz Ok, I made the required changes. Can you have a look? If it is OK i'll merge the commits

README
```diff
@@ -168,11 +168,39 @@ files (especially /etc/sysctl.d/01-lkrg.conf) are processed, or otherwise the
...
+headers, the DKMS utility, and OpenSSL.  Install the headers for the target kernels. 
+
+If this succeeds you should get similar output to the following:
+  
+  $ dkms status
+  lkrg/0.9.3, 5.18.9-200.fc36.x86_64, x86_64: installed
+
+Please refer to the previous two sections for how to start the LKRG service or
+have it started on system bootup. If you wish to use the unit/init file you must
+install it manually.
```
solardiz, line 196: Please insert the second space before "If" and wrap "must" onto the next line.

README
```diff
@@ -168,11 +168,39 @@ files (especially /etc/sysctl.d/01-lkrg.conf) are processed, or otherwise the
...
+
+
 Uninstalling
 ------------
 
 You can uninstall LKRG using "make" as well (still in the same directory):
-
+
```
solardiz, line 203: Please remove the whitespace edit here.

solardiz (changes requested):
Please correct the two trivial style issues and squash the commits into one.

BTW, I don't understand why you normally add more commits for review - you can be force-pushing the one work-in-progress commit right away.

Krish-sysadmin:
@solardiz Hope it's perfect now

solardiz (approved):

solardiz:
> Hope it's perfect now

It was good enough, so merged. Thanks!

One thing GitHub UI didn't show is trailing whitespace. I'll plan on fixing that with a separate commit now.

solardiz:
Also, many lines were over 79. Hard to see this in GitHub UI. All of this is now fixed with a subsequent commit.

More importantly, perhaps we should add something on how to uninstall LKRG when using DKMS? And how to upgrade it - is our advice to uninstall first appropriate in that case as well? @morfikov

Krish-sysadmin:
@solardiz Thank you. Yup, uninstall instructions should be added in my opinion. For upgrading, I'm not sure that's the best idea with DKMS. Same as with the normal Upgrade instructions it requires manual work.

Krish-sysadmin:
> And how to upgrade it - is our advice to uninstall first appropriate in that case as well?

I believe so

morfikov:
> More importantly, perhaps we should add something on how to uninstall LKRG when using DKMS? And how to upgrade it - is our advice to uninstall first appropriate in that case as well? @morfikov

Removing is easy as executing the `dkms remove ...` command. Upgrading is to remove and install the module again (newer version).

solardiz:
I wonder if we can/should simplify the instructions from separate use of `tar` and `dkms add` to running `dkms add` on the tarball or using `dkms ldtarball`. Are our release tarballs compatible with that?

> Removing is easy as executing the `dkms remove ...` command.

OK, I'm adding this:

```
If you installed using DKMS, you'd uninstall with:

        sudo dkms remove -m lkrg -v 0.9.3
```

I hope I got this right (untested).

morfikov:
> > Removing is easy as executing the `dkms remove ...` command.
> 
> OK, I'm adding this:
> 
> ```
> If you installed using DKMS, you'd uninstall with:
> 
>         sudo dkms remove -m lkrg -v 0.9.3
> ```
> 
> I hope I got this right (untested).

Yes, you can also use the following to remove the module for all kernels.

`# dkms remove -m lkrg/0.9.3 --all`

solardiz:
Are you saying the command I mentioned isn't correct as-is, or do you merely suggest optional alternatives? Which _one_ command do you recommend for inclusion in the documentation?

solardiz:
I'd rather not include something like `-k 5.7.7-amd64` (kernel version and distro specific), so if anything like this needs to be included, it'd have to be `--all`. But is this mandatory? Will my command not work as-is, I guess for the current kernel?

morfikov:
> Are you saying the command I mentioned isn't correct as-is, or do you merely suggest optional alternatives?

I checked `man dkms` and the command you provided appears to be valid. I was using only the two commands I posted in the previous post.

> Which _one_ command do you recommend for inclusion in the documentation?

All of them, :) 

- If you want to remove lkrg for current kernel, use...
- If you want to remove lkrg for some specific kernel, use...
- if you want to remove lkrg for all kernels, use...

morfikov:
Also be consistent with the command pattern.

`-m lkrg -v 0.9.3` is the same as `-m lkrg/0.9.3` , but I prefer the second one.

PR #218, opened 2022-07-27
CI: Remove EOL'ed Ubuntu releases

vt-alt:
Impish is started to fail, because it's EOL'ed at July 14, 2022 (replaced with `jammy`). Hirsute is EOL'ed at January 20, 2022, but it started to fail only recently (removed).

Tested on GA in my fork.

### DIFF
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -22,7 +22,7 @@ jobs:
             fail-fast: false
             matrix:
                 include:
-                    - image: arm64v8/ubuntu:impish
+                    - image: arm64v8/ubuntu:jammy
                       packages: libelf-dev linux-headers-generic
                     - image: arm32v7/alt:latest
                       packages: elfutils kernel-headers-modules-std-def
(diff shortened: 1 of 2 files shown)


vt-alt:
JFYI it's possible to add `Kinetic` to the (mkosi) tests, but their default kernel is same as `Impish` (5.15) yet. It's possible to configure it to use `kinetic-proposed` sub-repo with 5.19-rc7 (AFAIK), but this would require not trivial change of scripts, so I postponed this.

solardiz (approved):
Thank you, @vt-alt! I was worried we'd lose arm64 tests, but I see you're moving them to `jammy`. Great.

vt-alt:
> JFYI it's possible to add `Kinetic` to the (mkosi) tests, but their default kernel is same as `Jammy` (5.15) yet. It's possible to configure it to use `kinetic-proposed` sub-repo with 5.19-rc7 (AFAIK), but this would require not trivial change of scripts, so I postponed this.

Since we already have mkosi-mainline test, I decided that we don't need a hassle of testing in `kinetic-proposed` (it would make config setup more complicated (to `apt-add-repository -p proposed` pocket somewhere) and don't provide additional benefit to existing mainline test).

PR #228, opened 2022-09-02
CI: mkosi: Do not assume grub is installed

vt-alt:
mkosi' "--boot-protocols is deprecated and has no effect anymore", thus do not assume we will be booting using grub/bios.

Fixes: https://github.com/lkrg-org/lkrg/issues/227

[Not fully tested locally — will see GA report here.]

PR #229, opened 2022-09-03
CI: mkosi: Restore loading LKRG module from initrd

vt-alt:
For (non-grub) BLS boot - install generated by Dracut initrd into systemd-boot loader entry.

Man pages for curious details: systemd-boot(7), bootctl(1), kernel-install(8).

Link: https://github.com/lkrg-org/lkrg/pull/228#issuecomment-1235403740
Fixes: db42541 ("CI: mkosi: Do not assume grub is installed")

Only tested locally before push.

Using `kernel-install` is somewhat low-level'ish but it seems there's no more generic way.

ps. Curiously, this took 3 hours to investigate, produce best solution, and test.

### DIFF
--- a/mkosi.postinst
+++ b/mkosi.postinst
@@ -11,8 +11,24 @@ if test -x /usr/bin/dracut; then
 	banner postinst >&2
 	# Register our module in kmod database.
 	depmod -a $KERNELRELEASE
-	# Install module into (and force load early in) initrd.
-	dracut --force --force-drivers lkrg /boot/initrd.img-$KERNELRELEASE $KERNELRELEASE
+
+	# Install module into (and force its load early in) initrd.
+
+	# Add lkrg by default for any dracut invocation w/o requirement of
+	# passing '--force-drivers lkrg' command line option.
+	echo 'force_drivers+=" lkrg "' > /etc/dracut.conf.d/31-lkrg.conf
+	dracut --force /boot/initrd.img-$KERNELRELEASE $KERNELRELEASE
+
+	if [ -d /boot/loader/entries ]; then
+		# Install into systemd-boot loader.
+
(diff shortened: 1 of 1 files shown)


vt-alt:
OK this is only worked for mainline kernel test, but isn't for Ubuntu releases. Please do not merge yet.

vt-alt:
I suspect calling `bootctl ` is not reliable in `mkosi.postinst` stage on Ubuntu older than _kinetic_, so I replaced it with other check. And there is no `-v` in `kernel-install` on _bionic_. Can be merged if all turns green.

solardiz:
Thank you, @vt-alt!

The boot logs look good to me now.

One thing I don't get from reading your comments here and in the code: was there any issue with `--force-drivers lkrg`, what issue, or why else are we avoiding the need for this option now?

Anyway, I'll merge this now. Thanks again.

vt-alt:
There is no issue with the option in the last version, so this could be removed.

Intermediate versions used initrd rebuild by other methods which required adding `force_drivers+=` to dracut.conf.

But I thought it will be OK to keep dracut.conf method as more exemplary — in case someone take it as example of how to load LKRG in intrd (for their normal systems). (Perhaps, there is much overthinking on my part.) I did not find proper method of how to regenerate initrd now (for uefi). Before you just run  1 command — `dracut -f` (or even dracut with particular kernel version like we run) — but now this is not enough as initrd needs to additionally installed into BLS boot entry, so `kernel-install` needs to be run somehow too. Mkosi authors worked this around by running `dpkg-reconfigure dracut`, which (via hooks run `dracut` and `kernel-install` but) does not let any options pass into dracut.

vt-alt:
(Additionally) Another 'generic' method is to run `kernel-install` without passing last argument with initrd filename. This would trigger dracut run to [re]generate it.

Issue #183, opened 2022-05-06
Make LKRG's calls to kernel functions AArch64 BTI compatible


solardiz:
See https://github.com/lkrg-org/lkrg/pull/181#issuecomment-1117476393

Perhaps we could check the pointed-to instruction at runtime and, if it's a `bti`, then `bl` or `b` to it accordingly.

solardiz:
One way to workaround this issue could be to have a second set of function pointers in our RO page, and have assembly stubs with `b` instructions going via those pointers. Then if the target function needs that, we'd be copying the looked up symbol address to that second pointer, and replacing our main pointer with address of the stub. The stubs could be hard-coded or generated at runtime (in the latter case we'd need a writable page that we'd turn into executable after initialization - not ideal).

redplait:
according to https://lwn.net/Articles/804982/
> if BTI is enabled, the first instruction encountered after an indirect jump must be a special BTI instruction

(interesting note: not all exported functions conform with this requirement, see http://redplait.blogspot.com/2022/10/bti-incompatible-exported-functions-in.html)
so we can just check if first instruction in function is BTI c
and if not we could make 8byte thunk like
BTI c
B jimm26 ; to kernel function - this will be direct branch and BTI not violated
bcs offset is limited this page with thunks must be located in +-128mb from kernel
one of possible solution is to use memory from eBPF JIT

solardiz:
Here's my attempt at fixing this issue on top of #415

```diff
+++ b/src/p_lkrg_main.h
@@ -330,7 +330,8 @@ extern p_ro_page p_ro;
       goto p_sym_error; \
    }
 
-#ifdef CONFIG_X86_KERNEL_IBT
+//#ifdef CONFIG_X86_KERNEL_IBT
+#if defined(CONFIG_X86_64) || defined(CONFIG_ARM64)
 
 /*
  * When Intel CET IBT is in use, we can't call non-exported kernel functions
@@ -361,11 +362,20 @@ extern p_ro_page p_ro;
  * could improve upon this by having these functions in their own translation
  * unit, ideally coming from an assembly rather than C source file.
  */
+#ifdef CONFIG_X86_64
 #define GENERATE_CALL_FUNC(type, name, ...) \
    notrace __attribute__ ((noipa)) type call_##name(__VA_ARGS__) { \
       __asm__ __volatile__ ("pushq %0" :: "m" (P_SYM(name))); \
    } \
[...]

PR #237, opened 2022-10-29
CI: Re-enable AArch64 test, but without BTI

solardiz:
### Description
1. Re-enable AArch64 test in our GitHub Actions. The test had been introduced in #181, but promptly disabled since it was failing all the time because of our lack of support of BTI in our hackish calls to non-exported kernel functions, tracked as #183.
2. Now we bypass that issue by specifying to QEMU `-cpu cortex-a57` instead of `-cpu max`, so that we do test on AArch64 but without BTI. I got `-cpu cortex-a57` from https://wiki.ubuntu.com/ARM64/QEMU
3. We also switch to testing on Ubuntu Jammy instead of Ubuntu Impish, as the latter had been EOL'ed and is no longer downloadable (at least by this test as previously configured).

### How Has This Been Tested?
The testing is described in some detail in comments I added today to #181. In short, this test now passes in my fork, whereas a different test with only re-enabling and switching to Jammy, but keeping `-cpu max`, fails with the BTI error like before. So all 3 of the changes here are required.

### DIFF
--- a/.github/workflows/docker-boot.yml
+++ b/.github/workflows/docker-boot.yml
@@ -11,13 +11,13 @@ jobs:
 #                    - args: image=i386/ubuntu:bionic
 #                            qemu=qemu-system-i386
 #                      install: qemu-system-x86
-#
-#                    - args: image=arm64v8/ubuntu:impish
-#                            opts="-M virt,gic-version=3 -cpu max"
-#                            console=ttyAMA0
-#                            root=/dev/vda
-#                            qemu=qemu-system-aarch64
-#                      install: qemu-system-arm
+
+                    - args: image=arm64v8/ubuntu:jammy
+                            opts="-M virt,gic-version=3 -cpu cortex-a57"
+                            console=ttyAMA0
+                            root=/dev/vda
+                            qemu=qemu-system-aarch64
+                      install: qemu-system-arm
 
                     # It's possible to use: opts="-M raspi2b" dtb=bcm2709-rpi-2-b.dtb root=/dev/mmcblk0
                     # but since power-off is unreliable anyway (until QEMU 6.2) use simpler virt machine.


solardiz:
@vt-alt Do you have any advice on whether/how we can possibly have two variations of this test at once - with and without BTI?

In a private fork of this repo, @redplait and I have just tried adding an extra `args` block:

```
                    # test with BTI
                    - args: image=arm64v8/ubuntu:jammy
                            opts="-M virt,gic-version=3 -cpu max"
                            console=ttyAMA0
                            root=/dev/vda
                            qemu=qemu-system-aarch64
                      install: qemu-system-arm
```

However, this didn't result in an extra job running, perhaps because both have `image=arm64v8/ubuntu:jammy` and that becomes the job name. So how can we have different names yet use the same image?

solardiz:
> However, this didn't result in an extra job running, perhaps because both have `image=arm64v8/ubuntu:jammy` and that becomes the job name. So how can we have different names yet use the same image?

No, I'm wrong - this _did_ result in both jobs running, confusingly under the same name in GitHub UI, but clearly separated in e-mail that GitHub sent me:

```
  * build (image=arm64v8/ubuntu:jammy opts="-M virt,gic-version=3 -cpu cortex-a57" console=ttyAMA0 ro... succeeded (2 annotations)
  * build (image=arm64v8/ubuntu:jammy opts="-M virt,gic-version=3 -cpu max" console=ttyAMA0 root=/dev... failed (3 annotations)
```

So the question is - can we assign them different names, to make it easier to distinguish them in GitHub web UI? @vt-alt

vt-alt:
Each (`-`) element in `include` array is a hash table, so we can add any key, like `name` before `args` and it should be displayed first.

vt-alt:
I don't see any change anywhere, perhaps it's helped.

solardiz:
@vt-alt Your answer was helpful, thank you! We did not make use of it yet, simply keeping two tests as above in a private repo and that was sufficient for now. We might add the `name` thing later, for BTI or something else.

BTW, it could be useful for us to be able to enable BTI and/or PAC separately - right now, `-cpu max` appears to enable both at once. I did not yet look into whether/how QEMU can enable them separately, but we'll need separate changes to support them, so testing once change at a time could be cleaner.

vt-alt:
It seems the info should be in [ID_AA64ISAR1_EL1](https://developer.arm.com/documentation/ddi0595/2021-03/AArch64-Registers/ID-AA64ISAR1-EL1--AArch64-Instruction-Set-Attribute-Register-1) register.

In 2017 https://lore.kernel.org/lkml/1500480092-28480-5-git-send-email-mark.rutland@arm.com/
```
From ARMv8.3 onwards, ID_AA64ISAR1 is no longer entirely RES0, and now
has four fields describing the presence of pointer authentication
functionality:

* APA - address authentication present, using an architected algorithm
* API - address authentication present, using an IMP DEF algorithm
* GPA - generic authentication present, using an architected algorithm
* GPI - generic authentication present, using an IMP DEF algoithm

This patch adds the requisite definitions so that we can identify the
presence of this functionality. For the timebeing, the features are
hidden from userspace.
```

I [tried to read](https://docs.kernel.org/arm64/cpu-feature-registers.html) ID_AA64ISAR1_EL1 from userspace when `-cpu pauth=on` and `off`:
```
ON:  ID_AA64ISAR1_EL1    : 0x0000000000011001
OFF: ID_AA64ISAR1_EL1    : 0x0000000000011001
```
So there is no difference.

vt-alt:
On test VM:
```
root@aarch64:/.in# zcat /proc/config.gz |grep CONFIG_ARM64_PTR_AUTH
CONFIG_ARM64_PTR_AUTH=y
CONFIG_ARM64_PTR_AUTH_KERNEL=y
```
So `FTR_VISIBLE_IF_IS_ENABLED(CONFIG_ARM64_PTR_AUTH),` should be true.
Also it seems there should be `CPU features: detected` line in dmesg `Address authentication` but it's not there. Boot with `-cpu pauth=on`:
```
root@aarch64:/.in# dmesg|grep -i CPU.feature
[    0.000000] CPU features: detected: GIC system register CPU interface
[    0.000000] CPU features: detected: Hardware dirty bit management
[    0.000000] CPU features: detected: Spectre-v4
[    0.000000] CPU features: kernel page table isolation forced OFF by mitigations=off
[    0.176424] CPU features: detected: Common not Private translations
[    0.176425] CPU features: detected: CRC32 instructions
[    0.176426] CPU features: detected: Data cache clean to Point of Persistence
[    0.176428] CPU features: detected: LSE atomic instructions
[    0.176429] CPU features: detected: Privileged Access Never
[    0.176429] CPU features: detected: RAS Extension Support
```

vt-alt:
Ah, qemu command lines so long, and there was additional `-cpu max` option. When I delete it qemu fails with:
```
qemu-system-aarch64: 'pauth' feature not supported by KVM on this host
```
Then when I disable KVM, auth is there:
```
root@aarch64:/.in# dmesg|grep -i CPU.feature|grep -i auth
[    0.000000] CPU features: detected: Address authentication (architected QARMA5 algorithm)
[    0.538645] CPU features: detected: Generic authentication (architected QARMA5 algorithm)
```
When I run with `-cpu max,pauth=off` these lines disappear. When I run `-cpu max` they appear. (Note that at first I run `-cpu max` in KVM, now in TCG.)

So how we know method of selectively disabling/enabling pauth.

vt-alt:
Also we can disable bti/pauth support from kernel command line (since v5.12) by passing `arm64.nobti` and `arm64.nopauth`. I don't know if this is better than disabling from qemu for our purposes.

solardiz:
> > `-cpu max` appears to enable both at once.
> 
> How do you see that PAC is enabled?

Indirectly: when @redplait worked around BTI in a fork/branch of LKRG, the next hurdle was failing pCFI, meaning that addresses on the stack were not what our code was expecting, which we assumed was a result of PAC (it's as expected).

> Then when I disable KVM, auth is there:
> 
> ```
> root@aarch64:/.in# dmesg|grep -i CPU.feature|grep -i auth
> [    0.000000] CPU features: detected: Address authentication (architected QARMA5 algorithm)
> [    0.538645] CPU features: detected: Generic authentication (architected QARMA5 algorithm)
> ```
> 
> When I run with `-cpu max,pauth=off` these lines disappear. When I run `-cpu max` they appear.

Cool. BTW, how could you even have KVM before if your host isn't AArch64, or is it?

> Also we can disable bti/pauth support from kernel command line (since v5.12) by passing `arm64.nobti` and `arm64.nopauth`. This maybe sufficient for our purposes!

Great. So we have two ways to control these features separately. Thank you!

vt-alt:
> BTW, how could you even have KVM before if your host isn't AArch64, or is it?

I experimented with this locally on Kunpeng-920 didn't realizing it doesn't have these features.

PR #239, opened 2022-11-03
Add support for Kernel 5.19+ for aarch64

wladmis:
Since torvalds/linux@e9d75a0ba87851187fe52493f1527229a7e101b3 struct stackframe has been renamed to struct unwind_state for arm64.

### DIFF
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -1648,7 +1648,11 @@ int p_ed_enforce_pcfi(struct task_struct *p_task, struct p_ed_process *p_orig, s
    struct stack_frame p_frame;
 #endif
 #elif defined(CONFIG_ARM64)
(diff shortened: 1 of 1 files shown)


solardiz:
Thanks!

@vt-alt This probably means we need CI tests using latest mainline kernels on all supported archs, not only x86_64. Easy to add?

vt-alt:
> @vt-alt This probably means we need CI tests using latest mainline kernels on all supported archs, not only x86_64. Easy to add?

Ubuntu have built mainline packages for required arches, but perhaps we will not be able to install it easily on x86-64 GA CI system (then to cross-compile lkrg). Need to double verify this though, if `dpkg` have something like `rpm`s `--nodeps`.

Other options — (cross-)compile "manually" Linux from source tree (this will take traffic and time (remember there is just 2 cpu); use qemu-user'd docker (like we have for docker-boot) to then install these mainline packages normally and compile;

Issue #204, opened 2022-07-17
Harden LKRG's sysctl handlers


solardiz:
Right now, LKRG's sysctl handlers open/close r/w access at the outer layer, regardless of whether the access is a write or a read. We should either skip that for reads or use temporary (not on r/o page) variables in the outer layer so that the open/close can be moved to inside the existing `if` block.

Also, the min/max variables used in the sysctl table should be made `const`, and/or the values checked against the corresponding constants before being put into the r/o page.

solardiz:
Also, should fix handling of races between 2+ concurrent sysctl updates, where with current code one of these could set the page back to r/o and another then crash on writing into the page that it had opened r/w access for.

solardiz:
> use temporary (not on r/o page) variables in the outer layer so that the open/close can be moved to inside the existing if block.

I think we'll need to do that, because as seen in backtrace in https://github.com/lkrg-org/lkrg/issues/191#issuecomment-1159913630 the call to `proc_dointvec_minmax` can sleep (perhaps that was a debugging build deliberately tagging all potentially-sleeping memory allocations, but anyway at least theoretically the issue is presumably there).

I happen to have implemented this today for `log_level` for an unrelated reason. We could do similar for them all.

solardiz:
While at it, also look into and clean up usage of signed vs. unsigned integers vs. printf formats. Right now, we use a weird mix.

solardiz:
I think before proceeding to fix these issues, we should change most of the sysctl handlers to use a shared function in place of their bodies. So that we don't introduce even more copy-pasted-edited code.

solardiz:
> as seen in backtrace in [#191 (comment)](https://github.com/lkrg-org/lkrg/issues/191#issuecomment-1159913630) the call to `proc_dointvec_minmax` can sleep

Added the `bug` label for this reason.

PR #249, opened 2022-11-19
Add synchronization logic around sysctl interface

Adam-pi3:
This commit fixes the race problem between 2+ concurrent sysctl updates. It is part of the work related to #204

### Description
The way how the fix is implemented is to move synchronization logic to the `p_lkrg_open/close_rw` interface. The main reason for that is because we use these API outside of sysctl interface as well. We need to synchronize to these usage as well. It makes more sense to have such synchronization logic there.

It is fine to wrap `proc_dointvec_minmax` around mutex and kernel does that as well:
https://elixir.bootlin.com/linux/latest/source/kernel/kprobes.c#L953

TODO: We should rework sysctl interface to reduce the window between the `p_lkrg_open/close_rw` calls. Currently, we do all the work between these calls, but we don't need to. We can call `p_lkrg_close_rw` as soon as we update the control structures, and then do the rest of the work. However, it should be part of different PR.

### How Has This Been Tested?
I run various concurrent processes doing sysctl updates and successfully repro the problem. With these changes, problem didn't repro anymore.

### DIFF
--- a/src/modules/wrap/p_struct_wrap.h
+++ b/src/modules/wrap/p_struct_wrap.h
@@ -25,6 +25,8 @@
 #ifndef P_LKRG_WRAPPER_H
 #define P_LKRG_WRAPPER_H
 
+extern struct mutex p_ro_page_mutex;
+
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 14, 0)
 
 static inline void p_set_uid(kuid_t *p_arg, unsigned int p_val) {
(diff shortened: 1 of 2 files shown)


solardiz:
This approach makes sense to me. I intend to also take a closer look at the code in context, to ensure no instance is missed.

solardiz (approved):
Looks good to me.

solardiz:
> we use these API outside of sysctl interface as well. We need to synchronize to these usage as well.

Yes, we need to, but is it safe to potentially sleep on the mutex in all of those other places? It's in the module notifier that we set up with `register_module_notifier`. (I hope this is safe. I just don't know, hence the question.)

solardiz:
> is it safe to potentially sleep on the mutex in all of those other places? It's in the module notifier that we set up with `register_module_notifier`

Adam says this is fine, and reminds me we also take another mutex in the notifier anyway.

So I'll merge.

Issue #253, opened 2022-12-22
mkosi: error: unrecognized arguments: --with-unified-kernel-images=no --root-size=4G


solardiz:
Today `mkosi-mainline` has failed with `mkosi: error: unrecognized arguments: --with-unified-kernel-images=no --root-size=4G`. I guess something changed in `mkosi` and we need to adapt. @vt-alt

vt-alt:
```
## v15
    We also remove the WithoutUnifiedKernelImages= switch as building unified
    kernel images is trivial and fast these days.
```
IC. "Trivial" for someone, but needs investigating how to do it properly.

vt-alt:
Update. If I just remove `WithUnifiedKernelImages=no` and `RootSize=4G` to see what will happen, there is another error:
```
‣   Installing boot loader…
bootctl: unrecognized option '--root'
```
Meaning there is, perhaps, too old `bootctl` without `--root` option support. Even on _kinetic_ which is the latest Ubuntu. I will try to open issue at mkosi.

vt-alt:
My prediction is this will not be fixed (because the change systemd/mkosi@8bbbd836078a2b6f625ad298ebc13d06601971e4 is so fundamental in mkosi new workflow) making mkosi tests on Ubuntu not usable for us anymore.
```
Previously, we first provisioned the disk image with all its
partitions, then mounted the disk image and finally populated it with
contents. Now, we provision to a regular directory on the host filesystem
first, followed by creating the disk image with systemd-repart.
```
By this I conclude that `--root` option usage in `bootctl` is important to the new provisioning method. But we cannot just upgrade `bootctl` on these Ubuntu systems, especially on older Ubuntu.

Also, `systemd-repart` is only available since Ubuntu _jellyfish_.

PR #255, opened 2022-12-24
CI: Use latest usable mkosi v14

vt-alt:
Main (and only) branch of mkosi is switching to a bleeding edge systemd tools which are not available on Ubuntu we are testing on. Instead of migrating off mkosi to something else, we can still use some previously wording tag such as `v14'.

Link: https://github.com/lkrg-org/lkrg/issues/253

### DIFF
--- a/.github/workflows/mkosi-boot.yml
+++ b/.github/workflows/mkosi-boot.yml
@@ -20,7 +20,7 @@ jobs:
               # Native focal package seems to be too old (v5) and not even
               # able to build images properly.
               run: |
-                  sudo python3 -m pip install git+https://github.com/systemd/mkosi.git
+                  sudo python3 -m pip install git+https://github.com/systemd/mkosi.git@v14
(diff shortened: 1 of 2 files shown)


solardiz:
Thank you, @vt-alt! This is also the approach I'd have suggested based on your comments on the issue.

We can merge this as-is, or maybe you want to fix minor issues in the commit description first? Maybe edit it to:

```
CI: Use latest usable mkosi v14

Main (and only) branch of mkosi is switching to bleeding edge systemd
tools, which are not available on Ubuntu we are testing on. Instead of
migrating off mkosi to something else, we can still use some previously
working tag such as v14.

Fixes: https://github.com/lkrg-org/lkrg/issues/253
```

(fix typo in "wording" / "working", drop "a", use "Fixes" in place of "Link")

vt-alt:
Thanks for suggestions, force-pushed fixed commit.

Issue #258, opened 2023-02-01
Built-in LKRG on 6.1 Prevents Automatic Loading of RootFS Module at Boot


sempervictus:
When building LKRG into the kernel to permit trimming unused ksyms, booting Arch Linux fails at the FS mount with:
```sh
:: running early hook [udev]
Starting version 251.7-1-arch
:: running hook [udev]
:: Triggering uevents...
:: performing fsck on '/dev/vda1'
/dev/vda1: clean, 80060/2621440 files, 761980/10485499 blocks
:: mounting '/dev/vda1' on real root
[    3.030935][  T257] LKRG: ALERT: BLOCK: UMH: Executing program name /usr/bin/modprobe
mount: /new_root: unknown filesystem type 'ext4'.
       dmesg(1) may have more information after failed mount system call.
You are now being dropped into an emergency shell.
sh: can't access tty; job control turned off
[rootfs ]# 
```
despite use of `lkrg.profile_enforce=0` and `lkrg.profile_validate=0` at the kernel commandline.
Issuing a `mount` for the root FS results in:
```sh
[rootfs ]# mount /dev/vda1 /root/
[  135.459467][  T264] LKRG: ALERT: BLOCK: UMH: Executing program name /usr/bin/modprobe
mount: /root: unknown filesystem type 'ext4'.
       dmesg(1) may have more information after failed mount system call.
```
Manually running `modprobe ext4` works from the init shell, which then permits me to mount `/dev/vda1` on `/root`

LKRG pulled in from upstream yesterday evening, so current revision. 6.1 is built w/ `linux-hardened` patchset on GCC 12.2

Adam-pi3:
Looks like `/usr/bin/modprobe` is not in the UMH allow list. Can you try setting `lkrg.umh_enforce=0` and `lkrg.umh_validate=0` and verify if it boots?
If yes, we can add `/usr/bin/modprobe` to the allow list.

sempervictus:
Roger, wilco - the test environment w/ that built-in kernel is toast, but i'll get another built-in one set up once i wrap up some other tasks (including LLVM 16's kCFI+LTO validation).

sempervictus:
@Adam-pi3 - in #259 i'm seeing `squashfs` not being loaded while `loop` seems to be fine:
![image](https://user-images.githubusercontent.com/1331084/216214710-29a308fe-9e2a-49df-b64f-5c22d5511159.png)
[...]

PR #260, opened 2023-02-02
Add UMH modprobe path for Arch Linux init context

sempervictus:
Address #258 by adding /usr/bin/modprobe to the UMH allow-list.

### Description
Fixes booting on Arch Linux when LKRG is built into the kernel binary

### How Has This Been Tested?
Booted in an Arch Linux VM:
```sh
[root@svl-arch00 ~]# uname -r ; modinfo lkrg |head;sysctl -a|grep umh_
6.1.8
name:           lkrg
filename:       (builtin)
license:        GPL v2
file:           security/lkrg/lkrg
description:    pi3's Linux kernel Runtime Guard
author:         Adam 'pi3' Zabrocki (http://pi3.com.pl)
parm:           log_level:log_level [3 (issue) is default] (uint)
parm:           heartbeat:heartbeat [0 (don't print) is default] (uint)
parm:           block_modules:block_modules [0 (don't block) is default] (uint)
parm:           interval:interval [15 seconds is default] (uint)
lkrg.umh_enforce = 1
lkrg.umh_validate = 1
```

### DIFF
--- a/src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
+++ b/src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
@@ -60,6 +60,7 @@ static const char * const p_umh_global[] = {
    "/sbin/tomoyo-init",
    "/sbin/v86d",
    "/system/bin/start",
+   "/usr/bin/modprobe",
    "/usr/lib/systemd/systemd-cgroups-agent",
    "/usr/lib/systemd/systemd-coredump",
    "/usr/libexec/abrt-hook-ccpp",


src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
```diff
@@ -50,6 +50,7 @@ static const char * const p_umh_global[] = {
    "/sbin/drbdadm",
    "/sbin/hotplug",
    "/sbin/modprobe",
+   "/usr/bin/modprobe",
```
Adam-pi3, line 53: We are trying to keep this list in alphabetic order. Would you be able to move it accordingly?

Adam-pi3 (changes requested):

src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
sempervictus: All set

Adam-pi3:
Thanks @sempervictus !

PR #270, opened 2023-05-18
Simplify the use of P_SYM_INIT()

fluidog:
### Description

I simplified the usage of P_SYM_INIT by removing the second parameter - type declaration. I think this is unnecessary because type declaration is done during variable definition, not during variable initialization.

### How Has This Been Tested?

I compiled it successfully on Ubuntu 22.04 and did not encounter any errors when running it.

### DIFF
--- a/src/modules/database/FTRACE/p_ftrace_modify_all_code/p_ftrace_modify_all_code.c
+++ b/src/modules/database/FTRACE/p_ftrace_modify_all_code/p_ftrace_modify_all_code.c
@@ -207,10 +207,10 @@ int p_install_ftrace_modify_all_code_hook(void) {
 
    int p_tmp;
 
-   P_SYM_INIT(ftrace_lock, struct mutex *)
-   P_SYM_INIT(ftrace_rec_iter_start, struct ftrace_rec_iter *(*)(void))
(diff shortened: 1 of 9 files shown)


solardiz:
Thank you for the contribution, @fluidog! This has passed our CI tests (except for one, which failed for unrelated reason) and my brief test on outdated CentOS 7 (older than our CI tests). So it looks good in terms of backwards compatibility.

What about people who build kernel+LKRG with clang, though? Would they get warnings? https://stackoverflow.com/questions/74990777/how-to-use-typeof-in-clang-compiler-without-warning

I notice Linux kernel itself uses `typeof` starting with a version older than the oldest we support. So I guess its build environment maybe already suppresses that warning when building with clang?

Anyway, I think we can merge this. What do you think, @Adam-pi3?

solardiz (approved):

Adam-pi3:
LGTM. In case of `clang` maybe we can document that warning and point to the stackoverflow link which you pasted

solardiz:
> In case of `clang` maybe we can document that warning and point to the stackoverflow link which you pasted

I think that's premature - we don't even know whether the warning appears specifically when building LKRG. We also don't fully support clang - IIRC, we expect pCFI to be broken when LKRG is loaded into kernels built with clang, right?

PR #274, opened 2023-06-12
Update copyright

adrelanos:
If you're wondering why I am doing it, it's because it fixes a Debian lintian warning.

### DIFF
--- a/debian/copyright
+++ b/debian/copyright
@@ -4,13 +4,13 @@ Upstream-Contact: Adam 'pi3' Zabrocki <pi3@pi3.com.pl>
 Source: https://lkrg.org
 
 Files: *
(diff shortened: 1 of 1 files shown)


solardiz:
Weird stuff. Did _all_ of these people really make more contributions in 2023? I think not. So are we introducing errors to please a lint? That's counter-productive, so I wouldn't.

adrelanos:
Is making a contribution a necessary condition to update the copyright year? Bumping the year for me is only interpreted as bumping the expiration date of the copyright.

solardiz:
> Is making a contribution a necessary condition to update the copyright year?

In my understanding, yes.

> Bumping the year for me is only interpreted as bumping the expiration date of the copyright.

You have no grounds to bump the expiration date of the copyright if the person/entity made no changes to the copyrighted work. Essentially, you end up making a false statement.

PR #278, opened 2023-07-10
make bootup scripts independent of cwd, add usage and root checks

vobst:
### Description
For my use case it happened that I ran the install scripts not from the root of the project (and not through make). I noticed that they fail "silently" in that case. I thought it would not hurt to update the scripts to be more self-sufficient, i.e., independent of the cwd and with some usage and root checks.

### How Has This Been Tested?
Tested that it works as expected for systemd, cannot say anything about the openrc part.

### DIFF
--- a/scripts/bootup/lkrg-bootup.sh
+++ b/scripts/bootup/lkrg-bootup.sh
@@ -6,13 +6,26 @@
 #  - Adam 'pi3' Zabrocki (http://pi3.com.pl)
 ##
 
-P_LKRG_SYSTEMD="scripts/bootup/systemd/lkrg-systemd.sh"
-P_LKRG_OPENRC="scripts/bootup/openrc/lkrg-openrc.sh"
+set -eu
+
+P_SCRIPT_DIR="$(dirname "$0")"
+P_LKRG_SYSTEMD="${P_SCRIPT_DIR}/systemd/lkrg-systemd.sh"
+P_LKRG_OPENRC="${P_SCRIPT_DIR}/openrc/lkrg-openrc.sh"
 
 P_RED='\033[0;31m'
 P_GREEN='\033[0;32m'
 P_WHITE='\033[1;37m'
 P_NC='\033[0m' # No Color
+
+if [ $# -ne 1 ]; then
+	echo "Usage: $0 (install|uninstall)" >&2
+	exit 1
+fi
+
+if [ "$(id -u)" -ne 0 ]; then
+	echo "Please run as root." >&2
+	exit 1
+fi
 
 echo -e " ${P_GREEN}[*] ${P_WHITE}Executing LKRG's bootup installation script${P_NC}"
 
--- a/scripts/bootup/openrc/lkrg-openrc.sh
+++ b/scripts/bootup/openrc/lkrg-openrc.sh
@@ -7,8 +7,11 @@
 #  - Jakub 'mrl5' Kołodziejczak (https://github.com/mrl5)
 ##
 
+set -eu
+
 P_SYSCTL_DIR="/etc/sysctl.d"
 P_INITD_DIR="/etc/init.d"
+P_SCRIPT_DIR="$(dirname "$0")"
 RUNLEVEL="boot"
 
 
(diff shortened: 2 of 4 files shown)


scripts/bootup/lkrg-bootup.sh
```diff
@@ -6,24 +6,37 @@
 #  - Adam 'pi3' Zabrocki (http://pi3.com.pl)
 ##
 
-P_LKRG_SYSTEMD="scripts/bootup/systemd/lkrg-systemd.sh"
-P_LKRG_OPENRC="scripts/bootup/openrc/lkrg-openrc.sh"
+set -eu
+
+P_SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
```
solardiz, line 11: I think the `readlink` part here (and in other scripts you patch similarly) is not necessary. It puts the full absolute path into the variable, but we can also use a relative path (like we did before) since we're not changing directories.

In fact, we use `dirname` alone in `scripts/copy-builtin.sh`, so it's inconsistent to do things differently here.

scripts/bootup/lkrg-bootup.sh
```diff
@@ -6,24 +6,37 @@
...
 
+if [ "$#" -ne 1 ]; then
+  echo "Usage: $0 (install|uninstall)" >&2
+  exit 1
+fi
+
+if [ "$(id -u)" -ne 0 ]; then
+  echo "Please run as root." >&2;
+  exit 1;
+fi
```
solardiz, line 28: Wrong indentation. We indent with tabs further in the script.

The semicolons within the second `then` ... `fi` are unneeded, probably are typos.

Putting `$#` into double-quotes is weird since it's a numeric shell builtin variable.

I'm not sure we want the `>&2` redirects here, but I don't mind.

scripts/bootup/openrc/lkrg-openrc.sh
```diff
@@ -7,8 +7,11 @@
 #  - Jakub 'mrl5' Kołodziejczak (https://github.com/mrl5)
 ##
 
+set -ue
```
solardiz, line 10: Let's standardize on `set -eu` (order of arguments) like in the first script patched here.

solardiz (changes requested):
I made a few comments. These also apply to other instances of the same within the scripts patched here.

solardiz:
Regarding requiring root, I guess we could at some point want to make the opposite move - allow for running the scripts as non-root during package build (installing into a build user writable "buildroot" tree), skipping the actions that require root. However, for now I don't mind introducing the root checks as this is consistent with the current inflexible scripts.

solardiz:
@mrl5 Please review the OpenRC script changes here. Thank you!

scripts/bootup/lkrg-bootup.sh
```diff
@@ -6,24 +6,37 @@
...
 
 case "`readlink -e /proc/1/exe`" in
 	/usr/lib/systemd/systemd | \
 	/lib/systemd/systemd)
-		exec "$P_LKRG_SYSTEMD" "$@"
+		exec "$P_LKRG_SYSTEMD" "$1"
 		;;
 	/sbin/openrc-init | \
 	/sbin/init)
-		exec "$P_LKRG_OPENRC" "$@"
```
mrl5, line 26: although it takes only one arg, I'd leave `"$@"` as is

this is just personal preference backed up with a usecase where you introduce another arg and don't need to remember to pass `"$1" "$2"` in bunch of other places

mrl5 (commented):
looks legit! thanks @vobst for fixes and improvements 🚀  
Please consider applying suggestions from comments.

@solardiz if you prefer, I can test it a bit later on an OpenRC system and then leave my approval

solardiz (changes requested):
Thanks, the overall changes look good to me now. However, the separation into commits does not. I didn't say it - maybe I should have - but we generally expect that feedback is addressed by making changes to the existing commits and force-pushing, not by adding more commits. What do you use to manage this git branch - do you use git on the command line, or GitHub UI, or something else?

Are you able to easily revise existing commits and force-push? If you are, then please replace the added commit with the corresponding changes included into your first commit instead.

If you are not, then please revert the commit making changes to `README` (add a revert commit) and we'll have to merge this PR via GitHub's squash-and-merge (turning it into just one commit) and get these other changes via your other PR.

solardiz:
Also, we prefer to mention the changed subsystem or file name in the commit message (near the beginning). Your "make bootup scripts independent of cwd, add usage and root checks" does mention "bootup scripts", but "Add remark about lkrg-bootup.sh to DKMS section" leaves it unclear where the changes are being made (where that "DKMS section" is) without looking at the changes themselves. I suggest replacing the latter commit message with "README: Add remark about lkrg-bootup.sh to DKMS section".

solardiz (approved):
Looks good to me now. Thank you!

solardiz:
@mrl5 I think I'll merge this now. You can then test on an OpenRC system and let us know of your results in a comment here, and if there are any issues then send us another PR to fix those (but I don't expect issues). Thank you!

PR #279, opened 2023-07-10
Add remark about lkrg-bootup.sh to DKMS section

vobst:
### Description
Thought it might be nice to tell the people using DKMS how to easily install the bootup files.

### How Has This Been Tested?
Tested with systemd and DKMS.


solardiz:
Thanks. This commit is also currently included in #278. Depending on how you revise that other PR, we might end up merging these changes via that PR or this one. I'll leave the possibility (and this PR) open for now.

solardiz:
The same changes merged via #278 instead.

Issue #283, opened 2023-08-02
mkosi-mainline tests failing


solardiz:
This has been happening for a while, @vt-alt:

```
Processing triggers for linux-image-unsigned-6.5.0-060500rc4daily20230731-generic (6.5.0-060500rc4daily20230731.202307302200) ...
/etc/kernel/postinst.d/dracut:
dracut: Generating /boot/initrd.img-6.5.0-060500rc4daily20230731-generic
/etc/kernel/postinst.d/zz-systemd-boot:
Installing kernel version 6.5.0-060500rc4daily20230731-generic in systemd-boot...
debconf: unable to initialize frontend: Dialog
debconf: (Dialog frontend will not work on a dumb terminal, an emacs shell buffer, or without a controlling terminal.)
debconf: falling back to frontend: Readline
debconf: unable to initialize frontend: Readline
debconf: (Can't locate Term/ReadLine.pm in @INC (you may need to install the Term::ReadLine module) (@INC contains: /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.36.0 /usr/local/share/perl/5.36.0 /usr/lib/x86_64-linux-gnu/perl5/5.36 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl-base /usr/lib/x86_64-linux-gnu/perl/5.36 /usr/share/perl/5.36 /usr/local/lib/site_perl) at /usr/share/perl5/Debconf/FrontEnd/Readline.pm line 7.)
debconf: falling back to frontend: Teletype
update-initramfs: deferring update (trigger activated)
dpkg: error: dpkg frontend lock was locked by another process with pid 5
Note: removing the lock file is always wrong, can damage the locked area
and the entire system. See <https://wiki.debian.org/Teams/Dpkg/FAQ#db-lock>.
run-parts: /etc/kernel/postinst.d/zz-systemd-boot exited with return code 2
dpkg: error processing package linux-image-unsigned-6.5.0-060500rc4daily20230731-generic (--install):
 installed linux-image-unsigned-6.5.0-060500rc4daily20230731-generic package post-installation script subprocess returned error exit status 1
Processing triggers for dracut (059-4) ...
dracut: Generating /boot/initrd.img-6.3.0-7-generic
dracut: Generating /boot/initrd.img-6.5.0-060500rc4daily20230731-generic
Errors were encountered while processing:
 linux-image-unsigned-6.5.0-060500rc4daily20230731-generic
[...]

PR #284, opened 2023-08-03
CI: mkosi-mainline: Workaround 'dpkg: error: dpkg frontend lock'

vt-alt:
Mkosi's own dracut hook, which is triggered from `dpkg -i`, calls `dpkg-reconfigure` causing dpkg frontend lock conflict. But `initrd.img` is generated correctly. Workaround by ignoring `dpkg-reconfigure` errors in the hook.

  dpkg: error: dpkg frontend lock was locked by another process with pid 5
  Note: removing the lock file is always wrong, can damage the locked area
  and the entire system. See <https://wiki.debian.org/Teams/Dpkg/FAQ#db-lock>.
  run-parts: /etc/kernel/postinst.d/zz-systemd-boot exited with return code 2

Link: https://github.com/lkrg-org/lkrg/issues/283

### Description
I did not investigate thoroughly all internal workings of `dpkg -i`/`dpkg-reconfigure` and what is changed since last week in Ubuntu (old action logs are not enough to clarify this.)

### How Has This Been Tested?
Locally.

### DIFF
--- a/.github/workflows/mkosi-mainline.yml
+++ b/.github/workflows/mkosi-mainline.yml
@@ -20,6 +20,7 @@ jobs:
               run: |
                   sudo python3 -m pip install git+https://github.com/systemd/mkosi.git@v14
                   sudo sed -i 's/linux-generic/linux-virtual/' /usr/local/lib/python*/dist-packages/mkosi/__init__.py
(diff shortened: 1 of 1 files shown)


solardiz:
Thank you, @vt-alt! Looks like CI is happy here. I'll merge.

PR #285, opened 2023-08-21
CI: Update workflows to use latest versions of the actions

vt-alt:
This should quiet the warning:

> The following actions uses node12 which is deprecated and will be  forced to run on node16: actions/checkout@v2. For more info:  https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -25,16 +25,16 @@ jobs:
                     - image: registry.opensuse.org/opensuse/tumbleweed
                       run: zypper -n install -y gcc make kernel-default-devel awk
                     - image: registry.opensuse.org/opensuse/leap
-                      run: zypper -n install -y gcc make kernel-default-devel
+                      run: zypper -n install -y gcc make kernel-default-devel tar gzip
 
         container:
             image: ${{ matrix.image }}
             options: --security-opt seccomp=unconfined
 
         steps:
-            - uses: actions/checkout@v1
             - run: cat /etc/os-release
             - run: ${{ matrix.run }}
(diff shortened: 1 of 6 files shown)


vt-alt:
Hmm, openSUSE builds also got broken (missing `awk` and `git` in their runs). perhaps this is unrelated to actions update and related to some docker image update since previous successful `docker bold` was long `3 weeks ago`. I will try add commit with a fix later to this PR.

vt-alt:
I filed issue into GH checkout/action but they did not reply. This mean we will be not able to retain `opensuse/leap` for tests. If you are OK with it, I will add commit deleting it.

`opensuse/tumbleweed` fixes if we add `gawk` to `zypper install`.

ps. I said before 'it misses git' but I was incorrectly identified problem att.

solardiz:
> we will be not able to retain `opensuse/leap` for tests. If you are OK with it, I will add commit deleting it.

I would be OK with it if there were a better reason to do it, but removing a currently working test just to silence a warning does not feel like a good tradeoff. Do I understand this correctly? What do you think?

vt-alt:
Them we shall not merge it and postpone this until tests really fail with `v1`.

vt-alt:
Perhaps they started to use `tar` from inside of contained in `@v3`, and `leap` image does not have it (but `tumbleweed` have it). So if I install `tar` and `gzip` (and reorder `steps`) build starting to work again. I reopen, and will push fixed version later today.

vt-alt:
Can be merged now I think.

Issue #287, opened 2023-09-13
mkosi-mainline "Failed to start initrd-switch-root.service - Switch Root."


solardiz:
@vt-alt `mkosi-mainline` is failing for 2 days with:

```
 [  OK  ] Reached target initrd-switch-root.target - Switch Root.
         Starting initrd-switch-root.service - Switch Root...
[FAILED] Failed to start initrd-switch-root.service - Switch Root.
See 'systemctl status initrd-switch-root.service' for details.

Generating "/run/initramfs/rdsosreport.txt"
```

vt-alt:
Looks like boot failure. Ubuntu seems diverging more and more from that old mkosi we use.

vt-alt:
It seems that GitHub does not even plan to add Ubuntu newer than 22.04 on runners in foreseeable future, so we still cannot start to use latest `mkosi` again.

solardiz:
@vt-alt So we need to use mkosi v14 tag, right? Will you please make this change? Thanks!

vt-alt:
We already [using v14](https://github.com/lkrg-org/lkrg/pull/255).

I thought - maybe we finally can switch to a newer mkosi to fix the issue, but it's not time yet.

solardiz:
@vt-alt Ah, so since we have to stay with v14, we should also stay with older Ubuntu? Can we?

vt-alt:
Cannot, because (the) mainline kernels are for bleeding edge Ubuntu-s by design.

vt-alt:
I debugged the boot process and found that the immediate cause of the boot failure is Dracut not mounting the root filesystem even though all devices are present.

Dracut does not say why but with debug  enabled (`rd.debug rd.debugverbose`) it says too much and it's still unclear why.

Instead of delving deeper and becoming Dracut expert, it occurs to me that we can simply pass `root=LABEL=root` into cmdline, as mkosi creates root fs with the label `root`. And the experiment shows that this works.

PR #288, opened 2023-09-20
CI: mkosi: Fix mainline boot on Ubuntu mantic

vt-alt:
Ubuntu removed 'systemd-gpt-auto-generator' from systemd_253.5-1ubuntu5, causing dracut to be unable to find the root partition without assistance. It appears that specifying the root label could be a good solution.

Fixes: https://github.com/lkrg-org/lkrg/issues/287

### How Has This Been Tested?
Will be tested in this PR.


solardiz:
Thank you, @vt-alt! All green here and the change looks good to me. I'll merge now.

PR #291, opened 2023-10-12
Some small fixes and possible valuable information

alexander-pick:
### Description

I was hunting down a global oobr triggered in kasan. Tuned out the modprobe has to be run from `sudo -s` instead just using `sudo` only to make it work. The only noticeable trace is the error dumping to dmesg if it fails. I assume this happens if `kptr_restrict` is set and the user who runs `modprobe` has no `CP_SYSLOG` permissions. 

During the process I made a few changes which might be valuable:
* reflected the above situation to `README` 
* added some lines to remove static keywords `add-exports.sh`, otherwise my 6.x kernel refused to compile (using clang)
* added some checks to avoid hashes of zero sized sections so it's visible in debug log if this happens by setting the hash to 0xFFFFFFFF
* fixed `CONFIG_SECURITY_SELINUX` vs. `CONFIG_SECURITY_SELINUX_DEVELOP` which would enable `sel_write_enforce`
* fixes a typo in `P_LKRG_EXPLOIT_DETECTION_SECURTIY_PTRACE_ACCESS_H`
* fixed some types for clarification
* removed a sizeof as we can safely assume 8 for `uint64_t` and the next line shares this assumption by using `& 7`
* added a case 0 to the hash algorithm to satisfy clang and avoid warnings
* clarified a warning which I found to be misleading (I build in-tree, but as a module)

### How Has This Been Tested?
Tested the build a lot of times on multiple systems with different configuration

### DIFF
--- a/README
+++ b/README
@@ -142,8 +142,7 @@ for OpenRC:
 
 for other:
 
-  sudo -s 
-  modprobe -v lkrg
+  sudo modprobe -v lkrg
 
 
 Autoload on bootup
--- a/README
+++ b/README
@@ -142,7 +142,8 @@ for OpenRC:
 
 for other:
 
-	sudo modprobe -v lkrg
+  sudo -s 
+  modprobe -v lkrg
 
 
 Autoload on bootup
--- a/scripts/add-exports.sh
+++ b/scripts/add-exports.sh
@@ -8,3 +8,9 @@ echo "EXPORT_SYMBOL(lookup_fast);" >> fs/namei.c
 # keep symbol export inside the ifdef block defining the symbol
 sed -i '/static void __seccomp_filter_release.*/iEXPORT_SYMBOL(__put_seccomp_filter);\n' kernel/seccomp.c
 echo "EXPORT_SYMBOL(do_seccomp);" >> kernel/seccomp.c
+# fixes modpost errors
+sed -i 's/static int change_page_attr_set_clr/int change_page_attr_set_clr/g'  arch/x86/mm/pat/set_memory.c
+sed -i 's/static struct dentry \*lookup_fast/struct dentry \*lookup_fast/g' fs/namei.c
+sed -i 's/static long do_seccomp/long do_seccomp/g' kernel/seccomp.c
+sed -i 's/static void __seccomp_filter_release(/void __seccomp_filter_release(/g' kernel/seccomp.c
+sed -i 's/static void __put_seccomp_filter/void __put_seccomp_filter/g' kernel/seccomp.c
--- a/src/modules/database/p_database.c
+++ b/src/modules/database/p_database.c
@@ -32,13 +32,19 @@ int hash_from_ex_table(void) {
 
    p_db.kernel_ex_table.p_size = (unsigned long)(p_tmp - (unsigned long)p_db.kernel_ex_table.p_addr);
 
-   p_db.kernel_ex_table.p_hash = p_lkrg_fast_hash((unsigned char *)p_db.kernel_ex_table.p_addr,
-                                                  (unsigned int)p_db.kernel_ex_table.p_size);
+   if(p_db.kernel_ex_table.p_size != 0) {
+      p_db.kernel_ex_table.p_hash = p_lkrg_fast_hash( (unsigned char *)p_db.kernel_ex_table.p_addr,
+                                                      (unsigned int)p_db.kernel_ex_table.p_size);
+   } else {
+      p_debug_log(P_LOG_DEBUG,
+          "hashing of _ex_table failed, is kptr_restrict set or CAP_SYSLOG missing?");
+      p_db.kernel_ex_table.p_hash = 0xFFFFFFFF;
+   }
 
    p_debug_log(P_LOG_DEBUG,
-          "hash [0x%llx] ___ex_table start [0x%lx] size [0x%lx]",p_db.kernel_ex_table.p_hash,
-                                                                   (long)p_db.kernel_ex_table.p_addr,
-                                                                   (long)p_db.kernel_ex_table.p_size);
+      "hash [0x%llx] ___ex_table start [0x%lx] size [0x%lx]",p_db.kernel_ex_table.p_hash,
+                                                               (long)p_db.kernel_ex_table.p_addr,
+                                                               (long)p_db.kernel_ex_table.p_size);
 
    return P_LKRG_SUCCESS;
 }
@@ -95,10 +101,12 @@ int hash_from_kernel_rodata(void) {
 
 #if !defined(CONFIG_GRKERNSEC)
 
-   if(p_db.kernel_rodata.p_size != 0x0) {
+   if(p_db.kernel_rodata.p_size != 0) {
       p_db.kernel_rodata.p_hash = p_lkrg_fast_hash((unsigned char *)p_db.kernel_rodata.p_addr,
                                                    (unsigned int)p_db.kernel_rodata.p_size);
    } else {
+      p_debug_log(P_LOG_DEBUG,
+          "hashing of _rodata failed, is kptr_restrict set or CAP_SYSLOG missing?"
       p_db.kernel_rodata.p_hash = 0xFFFFFFFF;
    }
 
@@ -139,10 +147,16 @@ int hash_from_iommu_table(void) {
    p_db.kernel_iommu_table.p_hash = 0xFFFFFFFF;
 #endif
 
-   p_debug_log(P_LOG_DEBUG,
-          "hash [0x%llx] __iommu_table start [0x%lx] size [0x%lx]",p_db.kernel_iommu_table.p_hash,
-                                                                     (long)p_db.kernel_iommu_table.p_addr,
-                                                                     (long)p_db.kernel_iommu_table.p_size);
+   if(p_db.kernel_iommu_table.p_size != 0) {
+      p_debug_log(P_LOG_DEBUG,
+            "hash [0x%llx] __iommu_table start [0x%lx] size [0x%lx]",p_db.kernel_iommu_table.p_hash,
+                                                                        (long)p_db.kernel_iommu_table.p_addr,
+                                                                        (long)p_db.kernel_iommu_table.p_size);
+   } else {
+      p_debug_log(P_LOG_DEBUG,
+          "hashing of iommu failed, is kptr_restrict set or CAP_SYSLOG missing?"
+      p_db.kernel_iommu_table.p_hash = 0xFFFFFFFF;
+   }
 
 #else
 
--- a/src/modules/database/p_database.c
+++ b/src/modules/database/p_database.c
@@ -95,8 +95,12 @@ int hash_from_kernel_rodata(void) {
 
 #if !defined(CONFIG_GRKERNSEC)
 
-   p_db.kernel_rodata.p_hash = p_lkrg_fast_hash((unsigned char *)p_db.kernel_rodata.p_addr,
-                                                (unsigned int)p_db.kernel_rodata.p_size);
+   if(p_db.kernel_rodata.p_size != 0x0) {
+      p_db.kernel_rodata.p_hash = p_lkrg_fast_hash((unsigned char *)p_db.kernel_rodata.p_addr,
+                                                   (unsigned int)p_db.kernel_rodata.p_size);
+   } else {
+      p_db.kernel_rodata.p_hash = 0xFFFFFFFF;
+   }
 
 #else
 
--- a/src/modules/database/p_database.c
+++ b/src/modules/database/p_database.c
@@ -61,9 +61,15 @@ int hash_from_kernel_stext(void) {
    }
 
    p_db.kernel_stext.p_size = (unsigned long)(p_tmp - (unsigned long)p_db.kernel_stext.p_addr);
-   p_db.kernel_stext.p_hash = p_lkrg_fast_hash((unsigned char *)p_db.kernel_stext.p_addr,
-                                               (unsigned int)p_db.kernel_stext.p_size);
 
+   if(p_db.kernel_stext.p_size != 0) {
+      p_db.kernel_stext.p_hash = p_lkrg_fast_hash((unsigned char *)p_db.kernel_stext.p_addr,
+                                                (unsigned int)p_db.kernel_stext.p_size);
+   } else {
+      p_debug_log(P_LOG_DEBUG,
+          "hashing of _stext failed, is kptr_restrict set or CAP_SYSLOG missing?");
+      p_db.kernel_stext.p_hash = 0xFFFFFFFF;
+   }
 
 #if defined(P_LKRG_JUMP_LABEL_STEXT_DEBUG)
    if (!p_db.kernel_stext_copy) {
@@ -106,7 +112,7 @@ int hash_from_kernel_rodata(void) {
                                                    (unsigned int)p_db.kernel_rodata.p_size);
    } else {
       p_debug_log(P_LOG_DEBUG,
-          "hashing of _rodata failed, is kptr_restrict set or CAP_SYSLOG missing?"
+          "hashing of _rodata failed, is kptr_restrict set or CAP_SYSLOG missing?");
       p_db.kernel_rodata.p_hash = 0xFFFFFFFF;
    }
 
@@ -140,23 +146,25 @@ int hash_from_iommu_table(void) {
 
 
 #ifdef P_LKRG_IOMMU_HASH_ENABLED
-   p_db.kernel_iommu_table.p_hash = p_lkrg_fast_hash((unsigned char *)p_db.kernel_iommu_table.p_addr,
+
+   if(p_db.kernel_iommu_table.p_size != 0) {
+      p_db.kernel_iommu_table.p_hash = p_lkrg_fast_hash((unsigned char *)p_db.kernel_iommu_table.p_addr,
                                                      (unsigned int)p_db.kernel_iommu_table.p_size);
+   } else {
+      p_debug_log(P_LOG_DEBUG,
+          "hashing of iommu failed, is kptr_restrict set or CAP_SYSLOG missing?");
+      p_db.kernel_iommu_table.p_hash = 0xFFFFFFFF;
+   }
+
+   p_debug_log(P_LOG_DEBUG,
+      "hash [0x%llx] __iommu_table start [0x%lx] size [0x%lx]",p_db.kernel_iommu_table.p_hash,
+                                                               (long)p_db.kernel_iommu_table.p_addr,
+                                                               (long)p_db.kernel_iommu_table.p_size);
+
 #else
 // Static value - might change in normal system...
    p_db.kernel_iommu_table.p_hash = 0xFFFFFFFF;
 #endif
-
-   if(p_db.kernel_iommu_table.p_size != 0) {
-      p_debug_log(P_LOG_DEBUG,
-            "hash [0x%llx] __iommu_table start [0x%lx] size [0x%lx]",p_db.kernel_iommu_table.p_hash,
-                                                                        (long)p_db.kernel_iommu_table.p_addr,
-                                                                        (long)p_db.kernel_iommu_table.p_size);
-   } else {
-      p_debug_log(P_LOG_DEBUG,
-          "hashing of iommu failed, is kptr_restrict set or CAP_SYSLOG missing?"
-      p_db.kernel_iommu_table.p_hash = 0xFFFFFFFF;
-   }
 
 #else
 
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -379,7 +379,11 @@ int p_ed_wq_valid_cache_init(void) {
 
 static void p_ed_wq_valid_cache_delete(void) {
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 6, 0)
+   __flush_workqueue(system_unbound_wq);
+#else 
    flush_workqueue(system_unbound_wq);
+#endif
    if (p_ed_wq_valid_cache) {
       kmem_cache_destroy(p_ed_wq_valid_cache);
       p_ed_wq_valid_cache = NULL;
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -162,7 +162,7 @@ static const struct p_functions_hooks {
      "Won't enforce validation on 'generic_permission'",
      1
    },
-#ifdef CONFIG_SECURITY_SELINUX
+#ifdef CONFIG_SECURITY_SELINUX_DEVELOP
    { "sel_write_enforce",
      p_install_sel_write_enforce_hook,
      p_uninstall_sel_write_enforce_hook,
--- a/src/modules/exploit_detection/syscalls/p_security_ptrace_access/p_security_ptrace_access.h
+++ b/src/modules/exploit_detection/syscalls/p_security_ptrace_access/p_security_ptrace_access.h
@@ -19,7 +19,7 @@
  */
 
 #ifndef P_LKRG_EXPLOIT_DETECTION_SECURITY_PTRACE_ACCESS_H
-#define P_LKRG_EXPLOIT_DETECTION_SECURTIY_PTRACE_ACCESS_H
+#define P_LKRG_EXPLOIT_DETECTION_SECURITY_PTRACE_ACCESS_H
 
 struct p_security_ptrace_access_data {
     ktime_t entry_stamp;
--- a/src/modules/hashing/p_lkrg_fast_hash.c
+++ b/src/modules/hashing/p_lkrg_fast_hash.c
@@ -30,7 +30,7 @@ notrace uint64_t p_lkrg_fast_hash(const char *p_data, unsigned int p_len) {
 
    uint64_t p_tmp = 0;
 
-   p_lkrg_siphash(p_data, p_len, (uint8_t *)&p_global_siphash_key, (uint8_t *)&p_tmp, sizeof(p_tmp));
+   p_lkrg_siphash((uint8_t *)p_data, p_len, (uint8_t *)&p_global_siphash_key, (uint8_t *)&p_tmp, sizeof(p_tmp));
    return p_tmp;
 }
 
--- a/src/modules/hashing/p_lkrg_fast_hash.c
+++ b/src/modules/hashing/p_lkrg_fast_hash.c
@@ -26,11 +26,11 @@ uint128_t p_global_siphash_key;
 inline void p_lkrg_siphash(const uint8_t *in, const size_t inlen, const uint8_t *k,
                            uint8_t *out, const size_t outlen);
 
-notrace uint64_t p_lkrg_fast_hash(const char *p_data, unsigned int p_len) {
+notrace uint64_t p_lkrg_fast_hash(const unsigned char *p_data, unsigned int p_len) {
 
    uint64_t p_tmp = 0;
 
-   p_lkrg_siphash((uint8_t *)p_data, p_len, (uint8_t *)&p_global_siphash_key, (uint8_t *)&p_tmp, sizeof(p_tmp));
+   p_lkrg_siphash((uint8_t *)p_data, (size_t)p_len, (uint8_t *)&p_global_siphash_key, (uint8_t *)&p_tmp, sizeof(p_tmp));
    return p_tmp;
 }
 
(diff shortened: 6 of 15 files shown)


README
```diff
@@ -142,7 +142,8 @@ for OpenRC:
 
 for other:
 
-	sudo modprobe -v lkrg
+  sudo -s
```
Adam-pi3, line 145: @alexander-pick Why would you want to start a shell? @solardiz FYI

scripts/add-exports.sh
```diff
@@ -8,3 +8,9 @@ echo "EXPORT_SYMBOL(lookup_fast);" >> fs/namei.c
 # keep symbol export inside the ifdef block defining the symbol
 sed -i '/static void __seccomp_filter_release.*/iEXPORT_SYMBOL(__put_seccomp_filter);\n' kernel/seccomp.c
 echo "EXPORT_SYMBOL(do_seccomp);" >> kernel/seccomp.c
+# fixes modpost errors
+sed -i 's/static int change_page_attr_set_clr/int change_page_attr_set_clr/g'  arch/x86/mm/pat/set_memory.c
+sed -i 's/static struct dentry \*lookup_fast/struct dentry \*lookup_fast/g' fs/namei.c
+sed -i 's/static long do_seccomp/long do_seccomp/g' kernel/seccomp.c
+sed -i 's/static void __seccomp_filter_release(/void __seccomp_filter_release(/g' kernel/seccomp.c
```
Adam-pi3, line 15: It's already set at line 9 of this script.

scripts/add-exports.sh
```diff
@@ -8,3 +8,9 @@ echo "EXPORT_SYMBOL(lookup_fast);" >> fs/namei.c
 # keep symbol export inside the ifdef block defining the symbol
 sed -i '/static void __seccomp_filter_release.*/iEXPORT_SYMBOL(__put_seccomp_filter);\n' kernel/seccomp.c
 echo "EXPORT_SYMBOL(do_seccomp);" >> kernel/seccomp.c
+# fixes modpost errors
+sed -i 's/static int change_page_attr_set_clr/int change_page_attr_set_clr/g'  arch/x86/mm/pat/set_memory.c
```
Adam-pi3, line 12: We already export that symbol, why do you need it?

scripts/add-exports.sh
```diff
@@ -8,3 +8,9 @@ echo "EXPORT_SYMBOL(lookup_fast);" >> fs/namei.c
...
+sed -i 's/static struct dentry \*lookup_fast/struct dentry \*lookup_fast/g' fs/namei.c
```
Adam-pi3, line 13: It's also exported

scripts/add-exports.sh
```diff
@@ -8,3 +8,9 @@ echo "EXPORT_SYMBOL(lookup_fast);" >> fs/namei.c
...
+sed -i 's/static long do_seccomp/long do_seccomp/g' kernel/seccomp.c
```
Adam-pi3, line 14: and this

src/modules/database/p_database.c
```diff
@@ -32,13 +32,19 @@ int hash_from_ex_table(void) {
 
    p_db.kernel_ex_table.p_size = (unsigned long)(p_tmp - (unsigned long)p_db.kernel_ex_table.p_addr);
 
-   p_db.kernel_ex_table.p_hash = p_lkrg_fast_hash((unsigned char *)p_db.kernel_ex_table.p_addr,
-                                                  (unsigned int)p_db.kernel_ex_table.p_size);
+   if(p_db.kernel_ex_table.p_size != 0) {
```
Adam-pi3, line 35: Can the exception table be 0 in size? I'm not sure. @solardiz do you know?

src/modules/database/p_database.c
```diff
@@ -55,9 +61,15 @@ int hash_from_kernel_stext(void) {
    }
 
    p_db.kernel_stext.p_size = (unsigned long)(p_tmp - (unsigned long)p_db.kernel_stext.p_addr);
-   p_db.kernel_stext.p_hash = p_lkrg_fast_hash((unsigned char *)p_db.kernel_stext.p_addr,
-                                               (unsigned int)p_db.kernel_stext.p_size);
 
+   if(p_db.kernel_stext.p_size != 0) {
```
Adam-pi3, line 65: kernel _stext certainly cannot be 0 in size so that's not needed

src/modules/database/p_database.c
```diff
@@ -95,8 +107,14 @@ int hash_from_kernel_rodata(void) {
 
 #if !defined(CONFIG_GRKERNSEC)
 
-   p_db.kernel_rodata.p_hash = p_lkrg_fast_hash((unsigned char *)p_db.kernel_rodata.p_addr,
-                                                (unsigned int)p_db.kernel_rodata.p_size);
+   if(p_db.kernel_rodata.p_size != 0) {
```
Adam-pi3, line 110: Can the Read-Only table be 0 in size? I'm not sure too. I think it's not possible. @solardiz do you know?

src/modules/exploit_detection/syscalls/p_security_ptrace_access/p_security_ptrace_access.h
```diff
@@ -19,7 +19,7 @@
  */
 
 #ifndef P_LKRG_EXPLOIT_DETECTION_SECURITY_PTRACE_ACCESS_H
-#define P_LKRG_EXPLOIT_DETECTION_SECURTIY_PTRACE_ACCESS_H
+#define P_LKRG_EXPLOIT_DETECTION_SECURITY_PTRACE_ACCESS_H
```
Adam-pi3, line 22: Good catch ;-)

src/modules/hashing/p_lkrg_fast_hash.c
```diff
@@ -26,11 +26,11 @@ uint128_t p_global_siphash_key;
 inline void p_lkrg_siphash(const uint8_t *in, const size_t inlen, const uint8_t *k,
                            uint8_t *out, const size_t outlen);
 
-notrace uint64_t p_lkrg_fast_hash(const char *p_data, unsigned int p_len) {
+notrace uint64_t p_lkrg_fast_hash(const unsigned char *p_data, unsigned int p_len) {
```
Adam-pi3, line 29: Why?

src/modules/hashing/p_lkrg_fast_hash.c
```diff
@@ -45,8 +45,8 @@ notrace inline void p_lkrg_siphash(const uint8_t *in, const size_t inlen, const
    uint64_t k1 = U8TO64_LE(k + 8);
    uint64_t m;
    int i;
-   const uint8_t *end = in + inlen - (inlen % sizeof(uint64_t));
-   const int left = inlen & 7;
+   const uint8_t *end = in + inlen - (inlen % 8);
```
Adam-pi3, line 48: this is not needed. Sizeof is static value generated by compiler and more portable

src/modules/hashing/p_lkrg_fast_hash.c
```diff
@@ -65,25 +65,25 @@ notrace inline void p_lkrg_siphash(const uint8_t *in, const size_t inlen, const
 
    switch (left) {
       case 7:
-         b |= ((uint64_t)in[6]) << 48;
+         b |= ((uint64_t)end[6]) << 48;
```
Adam-pi3, line 68: Why? looks wrong, CC: @solardiz

src/modules/hashing/p_lkrg_fast_hash.h
```diff
@@ -78,6 +78,6 @@ typedef struct uint128_t {
 
 extern uint128_t p_global_siphash_key;
 
-uint64_t p_lkrg_fast_hash(const char *data, unsigned int len);
+uint64_t p_lkrg_fast_hash(const unsigned char *data, unsigned int len);
```
Adam-pi3, line 81: why?

src/modules/print_log/p_lkrg_print_log.h
```diff
@@ -137,6 +137,8 @@
    case P_LOG_FLOOD:                                                                       \
       p_print_ret = printk(KERN_DEBUG   P_LKRG_SIGNATURE "FLOOD: " p_fmt "\n", ## p_args); \
       break;                                                                               \
+   default:                                                                                \
```
Adam-pi3, line 140: I don't think it's needed since we manually inject specific type of log. If we decide to 'default' anyway, it should be an invasive break not silent break. E.g., it should include `#error` macro breaking compilation. @solardiz what do you think?

Adam-pi3 (changes requested):

README
wladmis: I'd prefer not o mention `sudo` at all. Just a text like 'run as superuser'

alexander-pick: I changed it to `sudo -s` as it will trigger a global oob read with kasan if you do it from `sudo` directly as outlned above. It was the reason why I even started looking into things at all.

scripts/add-exports.sh
alexander-pick: My additions remove the static keywords, line 9 adds the export. If static remains the module will fail to compile properly:

```
ERROR: modpost: vmlinux: local symbol 'change_page_attr_set_clr' was exported
ERROR: modpost: vmlinux: local symbol '__put_seccomp_filter' was exported
ERROR: modpost: vmlinux: local symbol 'do_seccomp' was exported
ERROR: modpost: vmlinux: local symbol 'lookup_fast' was exported
```

Essentially my changes stop this from happening.

scripts/add-exports.sh
alexander-pick: See above, it fixes modpost errors due to local syms.

scripts/add-exports.sh
alexander-pick: See above, it removes the static keyword, not adds an export, it's just to solve modpost errors.

scripts/add-exports.sh
alexander-pick: See above.

src/modules/hashing/p_lkrg_fast_hash.c
alexander-pick: All code paths I found pass an unsigned char* (even casted to one) not a char* so corrected this to match the inputs. In the end it won't matter as the pointer is a pointer but I cam across it a few times looking through the code.

src/modules/hashing/p_lkrg_fast_hash.c
alexander-pick: I know, but you are using & 7 in the line below, so this indicated for me that you safely assume that uint64_t is always 8 (which should be true as it is not arch dependent such as size_t).

src/modules/hashing/p_lkrg_fast_hash.c
alexander-pick: All implementations I know use end not in, but once in reached = end in the for loop it should be the same. That was just changed for readability and to avoid a moving pointer.

src/modules/hashing/p_lkrg_fast_hash.h
alexander-pick: Same as the comment for the .c:

All code paths I found pass an unsigned char* (even casted to one) not a char* so corrected this to match the inputs. In the end it won't matter as the pointer is a pointer but I cam across it a few times looking through the code.

src/modules/print_log/p_lkrg_print_log.h
alexander-pick: clang will shelf out a warning if you have no default in there. As I compile in-tree with CONFIG_WERROR the build will fail and I get a lot of warnings from clang about it. Other than that it doesn\t do much at all.

src/modules/hashing/p_lkrg_fast_hash.c
alexander-pick: Just noticed that `in` will never equal `end` as the for loop terminates on !=, so if using `in` for the left over bytes would read again the last block which was previously processed in the loop for calculation, instead of the real left over bytes. 

Reference: https://android.googlesource.com/kernel/common/+/3af7a2f61023/lib/siphash.c

alexander-pick: Just reevaluated it a few times. `in` and `end` are equal after the loop, as `in` get bumped again but not processed by the `for`. So writing either end or in is pretty much the same. My reference for using end was mainly the Android version here:

https://android.googlesource.com/kernel/common/+/3af7a2f61023/lib/siphash.c

And if working through things it avoids confusion with the `!=` .

alexander-pick:
Is there anything else I could help with in regards of this PR? Were the comments helpful so far?

solardiz:
Hi. Thank you @alexander-pick for contributing. I'm sorry I don't yet comment on the individual issues. I think that this PR as a whole is not to be merged(*), but there's valuable information here and many of the individual changes should get in, some in revised form.

(*) too many assorted changes, some are questionable, the separation into commits is historical rather than logical, commit author names are placeholders, there are merge commits instead of rebases

README
Adam-pi3: `sudo -s` doesn't look right in that context. I'm wondering witch OOB read you see, related to LKRG or no. In general, LKRG does "intentional" OOB read because we do calculate the hash over entire sections of the kerle (e.g., entire RO section). By design, we read entire section size using pointer to the beginning of it.

scripts/add-exports.sh
Adam-pi3: Ah, thanks. Can you group the symbols together? e.g., when operating on `__seccomp_filter_release` we likely want to do all the changes at once and not in 2-3 different places. I assume, your changes can me merged into the original lines manipulating on that specific symbol. For other symbols just to work on them at the same place in file.

src/modules/hashing/p_lkrg_fast_hash.c
Adam-pi3: It's difficult to tell what is 'safe assumption'. I would rather leave it as it was original, especially that LKRG likely will get other arch support

src/modules/hashing/p_lkrg_fast_hash.c
Adam-pi3: I would leave it to @solardiz as he is more competent in that field

src/modules/print_log/p_lkrg_print_log.h
Adam-pi3: Can you add more context for that 'default' statement per my original comment?

Adam-pi3:
I just replied to some of the comments. Sorry for delay - had busy week.

solardiz:
> Tuned out the modprobe has to be run from `sudo -s` instead just using `sudo` only to make it work. The only noticeable trace is the error dumping to dmesg if it fails. I assume this happens if `kptr_restrict` is set and the user who runs `modprobe` has no `CP_SYSLOG` permissions.

This makes no sense to me, but perhaps I am missing something. First, why would these different ways to use `sudo` affect availability of `CAP_SYSLOG` to `modprobe`? Then, why would `CAP_SYSLOG` matter here? Finally, what's the presumably relevant change here - is it not passing the command on `sudo` command-line, or is it using the `-s` option?

This looks like a maybe-workaround for some issue (or non-issue) that probably remains. Is this your understanding, too?

If so, perhaps this isn't a reliable workaround either. I really doubt we want to have it, and certainly we don't want to have the weird references to these guessed details in LKRG messages.

src/modules/exploit_detection/p_exploit_detection.c
```diff
@@ -379,7 +379,11 @@ int p_ed_wq_valid_cache_init(void) {
 
 static void p_ed_wq_valid_cache_delete(void) {
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 6, 0)
+   __flush_workqueue(system_unbound_wq);
+#else 
    flush_workqueue(system_unbound_wq);
+#endif
```
solardiz, line 386: Why is the change from `flush_workqueue` to `__flush_workqueue` needed, and why only on 6.6+? Is this possibly to workaround the warning? If so, that's not 6.6+ specific, is it?

src/modules/integrity_timer/verify_kprobes/p_verify_kprobes.c
```diff
@@ -28,7 +28,7 @@ static struct kretprobe p_lkrg_dummy_kretprobe = {
     .entry_handler = p_lkrg_dummy_entry,
 };
 
-__attribute__((optimize(0)))
+__attribute__ ((optnone))
```
solardiz, line 31: Why is the change from `attribute__((optimize(0)))` to `__attribute__ ((optnone))`?

src/modules/print_log/p_lkrg_print_log.h
solardiz: Adam, I think we can add a `default:` to pacify compiler warnings. We just need to think where exactly to put it best. We can't do a `#error` since this would only be detected after the preprocessor stage (even if still at compile-time). We could maybe do a `static_assert`, but I think this is too minor a use case to start using it in this codebase.

solardiz (commented):
I've made some comments. Overall, I think instead of this PR with a lot in it, we need separate PRs (initially just one, then more) for the different kinds of changes - e.g., start with fixing the typos and changing wording of messages and using proper committer name there - we'll merge that, then proceed to the next PR with slightly less trivial changes, etc. Eventually, we'd have decided on these all (accepted some, revised others, rejected the rest). Thanks!

solardiz:
For supporting zero-sized sections, I think we need to revise the hash function itself instead of its call sites. This also side-steps the question of which sections can or cannot be zero-size.

solardiz:
> added some checks to avoid hashes of zero sized sections so it's visible in debug log if this happens by setting the hash to 0xFFFFFFFF

What's the problem and goal of change here? Did anything misbehave at all? Or is it simply an enhancement to be able to see which sections are zero-sized? If so, why is that desirable?

src/modules/exploit_detection/p_exploit_detection.c
```diff
@@ -162,7 +162,7 @@ static const struct p_functions_hooks {
      "Won't enforce validation on 'generic_permission'",
      1
    },
-#ifdef CONFIG_SECURITY_SELINUX
+#ifdef CONFIG_SECURITY_SELINUX_DEVELOP
    { "sel_write_enforce",
```
solardiz, line 166: Assuming that this change from `CONFIG_SECURITY_SELINUX` to `CONFIG_SECURITY_SELINUX_DEVELOP` is correct, we also need to make it in `src/modules/exploit_detection/syscalls/p_sel_write_enforce/p_sel_write_enforce.c`. Otherwise we'll have dead code in some builds.

src/modules/database/p_database.c
```diff
@@ -32,13 +32,19 @@ int hash_from_ex_table(void) {
...
    p_db.kernel_ex_table.p_size = (unsigned long)(p_tmp - (unsigned long)p_db.kernel_ex_table.p_addr);
 
-   p_db.kernel_ex_table.p_hash = p_lkrg_fast_hash((unsigned char *)p_db.kernel_ex_table.p_addr,
-                                                  (unsigned int)p_db.kernel_ex_table.p_size);
+   if(p_db.kernel_ex_table.p_size != 0) {
+      p_db.kernel_ex_table.p_hash = p_lkrg_fast_hash( (unsigned char *)p_db.kernel_ex_table.p_addr,
+                                                      (unsigned int)p_db.kernel_ex_table.p_size);
+   } else {
+      p_debug_log(P_LOG_DEBUG,
+          "hashing of _ex_table failed, is kptr_restrict set or CAP_SYSLOG missing?");
```
solardiz, line 40: Why would `kptr_restrict set or CAP_SYSLOG missing` result in sections appearing to be zero-sized? This really makes no sense to me, in any of these places.

alexander-pick: Well this is what I see with `sudo` only, works with `sudo -s`:
```
cat /proc/kallsyms | grep rodata
0000000000000000 T mark_rodata_ro
0000000000000000 D __start_rodata
0000000000000000 D rodata_enabled
0000000000000000 R __end_rodata
0000000000000000 R __end_rodata_aligned
0000000000000000 R __end_rodata_hpage_align
0000000000000000 d rodata_resource
0000000000000000 t set_debug_rodata
0000000000000000 d __setup_str_set_debug_rodata
0000000000000000 d __setup_set_debug_rodata
0000000000000000 t hash_from_kernel_rodata	[lkrg]
```
I assumed `kptr_restrict` is the issue here as the module also had some issues. I didn't debug it in detail, another possible issue might be some issue with the kprobe hack used to get `get_kallsyms_address`. So I basically added some input validation to avoid a null pointer. The message is just info I added to see if something went bad.

src/modules/integrity_timer/verify_kprobes/p_verify_kprobes.c
alexander-pick: The flag will be cause an unknown attribute warning with clang, `optnone` works. Possibly it needs to be changed to for full compatibility:

```
#ifdef __clang__
__attribute__ ((optnone))
#endif
#ifdef __GNUC__
__attribute__((optimize(0)))
#endif
```

src/modules/exploit_detection/p_exploit_detection.c
alexander-pick: Flushing the global workqueue will trigger a warning now. The discussion about it is from 2022 and it was introduced with 6.6rc1.

Reference:
https://github.com/torvalds/linux/commit/20bdedafd2f63e0ba70991127f9b5c0826ebdb32
https://github.com/torvalds/linux/blob/d88520ad73b79e71e3ddf08de335b8520ae41c5c/include/linux/workqueue.h#L619

src/modules/hashing/p_lkrg_fast_hash.c
alexander-pick: It's a standardized fixed width integer type and to my knowledge it it always 8 bytes, it's not like size_t which is arch dependent. 
If you want to be extra sure that some strange compiler is not messing it up, you would need to change `inlen & 7` to `inlen & (sizeof(uint64_t) - 1)` too. Otherwise you have a bug in one version or the other. 
Found it quite interesting that this is already the same in the reference implementation, the `siphash()` implementation in the linux kernel does it with sizeof in both places.

README
alexander-pick: You are right abotu the oob part. That was what I was investigating in first place. It turned out to be exactly that as lkrg runs into the first kasan global redzone (0xf9) it finds during hash creation by reading the entire section. Sadly the whole issue was a bit diffuse as some other hardening options I run created other issues as well. The `sudo -s` just avoided some of them.

src/modules/print_log/p_lkrg_print_log.h
alexander-pick: Here is a snippet with some warning clang will give you for each `p_print_log` as reference. As clang being a pretty wide spread compiler these days I think it is important to deal with it.

```
In file included from /work/kernel/lkrg/src/modules/database/arch/arm/p_arm_metadata.c:29:
In file included from /work/kernel/lkrg/src/modules/database/arch/arm/../../../../p_lkrg_main.h:435:
In file included from /work/kernel/lkrg/src/modules/database/arch/arm/../../../../modules/exploit_detection/p_exploit_detection.h:583:
/work/kernel/lkrg/src/modules/database/arch/arm/../../../../modules/exploit_detection/arch/x86/p_ed_x86_arch.h:186:25: warning: no case matching constant switch condition '0'
            p_print_log(P_LOG_ALERT, "BLOCK: CPU: SMAP got disabled, re-enabling");
                        ^~~~~~~~~~~
/work/kernel/lkrg/src/modules/database/arch/arm/../../../../modules/print_log/p_lkrg_print_log.h:92:21: note: expanded from macro 'P_LOG_ALERT'
#define P_LOG_ALERT 0
                    ^
/work/kernel/lkrg/src/modules/database/arch/arm/../../../../modules/print_log/p_lkrg_print_log.h:112:12: note: expanded from macro 'p_print_log'
   switch (p_level) {                                                                      \
           ^~~~~~~
In file included from /work/kernel/lkrg/src/modules/database/arch/arm/p_arm_metadata.c:29:
In file included from /work/kernel/lkrg/src/modules/database/arch/arm/../../../../p_lkrg_main.h:435:
In file included from /work/kernel/lkrg/src/modules/database/arch/arm/../../../../modules/exploit_detection/p_exploit_detection.h:583:
/work/kernel/lkrg/src/modules/database/arch/arm/../../../../modules/exploit_detection/arch/x86/p_ed_x86_arch.h:199:25: warning: no case matching constant switch condition '0'
            p_print_log(P_LOG_ALERT, "ALLOW: CPU: SMAP got disabled, accepting");
                        ^~~~~~~~~~~
/work/kernel/lkrg/src/modules/database/arch/arm/../../../../modules/print_log/p_lkrg_print_log.h:92:21: note: expanded from macro 'P_LOG_ALERT'
#define P_LOG_ALERT 0
                    ^
/work/kernel/lkrg/src/modules/database/arch/arm/../../../../modules/print_log/p_lkrg_print_log.h:112:12: note: expanded from macro 'p_print_log'
   switch (p_level) {                                                                      \
           ^~~~~~~
15 warnings generated.
```

src/modules/hashing/p_lkrg_fast_hash.c
alexander-pick: I pushed a change for both lines now.

solardiz:
@alexander-pick First of all, thank you for your effort. As I already mentioned, this PR as a whole is not merge-able. Do you want to start creating separate PRs for the different kinds of changes here, or should I or Adam be creating our own PRs based on your suggested changes here and crediting you indirectly (e.g., as Suggested-by or Co-authored-by, perhaps with your review)? Thanks again.

alexander-pick:
No problem, you can do it as you want and credit me in some way as you feel it's best done. The PR originated in work I did for myself and some review I did for curiosity, that is why it sadly got a bit mixed up. But I felt like some of it might be valuable and hope it helps.

src/modules/exploit_detection/p_exploit_detection.c
solardiz: Confirmed that `CONFIG_SECURITY_SELINUX_DEVELOP` is the correct flag to check, but the source code link in the comment above isn't for `sel_write_enforce`. Here's the correct snippet, from `git blame security/selinux/selinuxfs.c`:

```
^1da177e4c3f4 (Linus Torvalds          2005-04-16 15:20:36 -0700  134) #ifdef CONFIG_SECURITY_SELINUX_DEVELOP
1872981b51dac (Eric Paris              2008-04-17 14:15:45 -0400  135) static ssize_t sel_write_enforce(struct file *file, const char __user *buf,
^1da177e4c3f4 (Linus Torvalds          2005-04-16 15:20:36 -0700  136)                           size_t count, loff_t *ppos)
```

I used `git blame` to ensure there was no recent change - in other words, that this check would be valid across our range of supported kernels.

solardiz:
> No problem, you can do it as you want and credit me in some way as you feel it's best done.

I've proceeded to take pieces from this PR into separate commits and am testing them in my fork of the repo. I use `Co-authored-by: Alexander Pick <contact@alexander-pick.com>`.

solardiz:
> I've proceeded to take pieces from this PR into separate commits and am testing them in my fork of the repo.

My commits so far passed tests, so pushed to this repo as well. There are still many more things to take from this PR.

solardiz:
For SipHash, I see several kinds of changes here:
1. Use Linux's own implementation where available.
2. Avoid possible clash on `uint128_t`.
3. Avoid compiler warnings.
4. Make the code more similar to other implementations.

Regarding 1, I am not sure. On one hand, it's good to make LKRG smaller (in terms of compiled code) and use what's maintained upstream. On the other, we'll need to maintain and test our own implementation anyway (for older kernels) and using the kernel's may go against our security model. We compare those hashes in case the kernel or a module are modified - to hopefully detect that and respond to it. If we use the kernel's SipHash, we trust it, including with the per-boot random key passed to it as a function argument. So we'd possibly make it a bit easier for a kernel rootkit or such to be undetected by LKRG, as it could easily intercept SipHash and compute/return updated hashes that would be valid for the key we use. They could probably intercept LKRG's SipHash just as well, but at least that's something we should defend against in future versions of LKRG (make it harder to locate). For a kernel's exported symbol, we would have no reasonable recourse other than stopping to use it.

For 2, I'm merging this change (simplified).

For 3, I'm adding an `#if` to allow building this code in userspace, which enabled me to easily expose and then avoid some compiler warnings with gcc and clang.

For 4, I am not sure. I'm not doing anything yet. If we're to clean up the code, we could as well make even more clean-ups. For example, if/when we switch to Linux kernel's coding style, we could maybe have the code almost match what's used there, including indentation.

solardiz:
> For supporting zero-sized sections, I think we need to revise the hash function itself instead of its call sites.

Actually, I see nothing to revise - it appears to work just fine on empty input as-is:

```c
#include <stdio.h>
#include <stdint.h>
#include "p_lkrg_fast_hash.h"

int main(void)
{
	printf("%llx\n", (unsigned long long)p_lkrg_fast_hash(NULL, 0));
	return 0;
}
```

prints `1e924b9d737700d7`, so that's apparently the hash value for empty input with all-zero key (not initialized here).

@alexander-pick Can you please explain why you wanted to handle zero-sized sections specially? I see currently no reason, but maybe I'm missing something.

Issue #269, opened 2023-05-14
kernel.stext integrity check failed due to an error in handling the p_arch_jump_label_transform_entry address on ARM64


root-hardenedvault:
Reproduction steps:
1. Build LKRG & load the LKM
2. Reboot and then the kernel will panic:

Hardware: Rasperry Pi 4
OS: Ubuntu 22.04

```
[  302.646797] VED: ALERT: DETECT: Kernel: _stext hash changed unexpectedly
[  302.660354] VED: ALERT: DETECT: Kernel: 1 checksums changed unexpectedly
[  302.667154] VED: ALERT: BLOCK: Kernel: 1 checksums changed unexpectedly
[  302.673863] Kernel panic - not syncing: Kernel: 1 checksums changed unexpectedly
[  302.681366] CPU: 2 PID: 484 Comm: kworker/u8:6 Tainted: G         C OE     5.15.0-1027-raspi #29-Ubuntu
[  302.690899] Hardware name: Raspberry Pi 4 Model B Rev 1.4 (DT)
[  302.696816] Workqueue: events_unbound p_check_integrity [ved]
[  302.702698] Call trace:
[  302.705173]  dump_backtrace+0x0/0x200
[  302.708890]  show_stack+0x20/0x30
[  302.712250]  dump_stack_lvl+0x8c/0xb8
[  302.715965]  dump_stack+0x18/0x34
[  302.719324]  panic+0x1e4/0x3e4
[  302.722420]  p_check_integrity+0x1370/0x18d4 [ved]
[  302.727316]  process_one_work+0x204/0x4e0
[  302.731385]  worker_thread+0x144/0x490
[  302.735187]  kthread+0x128/0x134
[  302.738457]  ret_from_fork+0x10/0x20
[  302.742086] SMP: stopping secondary CPUs
[  302.746067] Kernel Offset: 0x5a489ca00000 from 0xffff800008000000
[  302.752245] PHYS_OFFSET: 0xffffadd980000000
[  302.756482] CPU features: 0x800804f1,00000846
[  302.760899] Memory Limit: none
[  302.763996] ---[ end Kernel panic - not syncing: Kernel: 1 checksums changed unexpectedly ]---
```

[...]

PR #294, opened 2023-10-25
kINT: Fix synchronization problem

Adam-pi3:
The reported problem with integrity verification on ARM64 (#269) is a result of a very tight race condition with tracepoints. Changes which simplify synchronization with JUMP_LABEL engine: f98da1b17c01c168ec163b0349327661adc9e38d
affected differently ARM64 platform which made such race possible. However, potentially the same race problem may exist on x86 and this commit fixes it and should address #269

### How Has This Been Tested?
I've run various stress manual tests on:
- ARM64: using Raspberry Pi 4 (under the kernel 6.5.0-1005-raspi)
- x86: using various kernels from Ubuntu 23.10 and 22.04.3, custom kernel 6.3.4 and CentOS 8 (kernel 4.18.0-240.15.1.el8_3.x86_64)

### DIFF
--- a/src/modules/database/p_database.c
+++ b/src/modules/database/p_database.c
@@ -191,6 +191,7 @@ int p_create_database(void) {
 
    P_SYM_INIT(jump_label_mutex)
    P_SYM_INIT(text_mutex)
+   P_SYM_INIT(tracepoints_mutex)
 
    /*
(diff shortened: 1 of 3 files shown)


solardiz (approved):
I was a bit concerned this could introduce dead-locks, but grepping the kernel tree for uses of `tracepoints_mutex` this looks safe.

Issue #295, opened 2023-10-25
"_stext hash changed unexpectedly" when loading at boot on RHEL8


solardiz:
For a package of LKRG 0.9.7 built with our RPM spec file here on a RHEL 8 rebuild (package built for Rocky 8, loaded on Alma 8), after `systemctl enable lkrg` and `reboot`, with an edit to `/etc/sysctl.d/01-lkrg.conf` to prevent the panic:

```
[    4.076077] LKRG: ALIVE: Loading LKRG
[    4.588367] LKRG: ISSUE: [kretprobe] register_kretprobe() for <ovl_dentry_is_whiteout> failed! [err=-22]
[    4.623828] LKRG: ISSUE: [kretprobe] register_kretprobe() for ovl_dentry_is_whiteout failed and ISRA / CONSTPROP version not found!
[    4.623834] LKRG: ISSUE: Can't hook 'ovl_dentry_is_whiteout'. This is expected when OverlayFS is not used.
[    4.832509] LKRG: ISSUE: IOMMU table can't be found (skipping)
[    4.849556] LKRG: ALIVE: LKRG initialized successfully
[    5.000069] LKRG: STATE: Changing 'kint_enforce' from 2 (PANIC) to 1 (LOG ONLY (For SELinux and CR0.WP LOG & RESTORE))
[    7.571959] LKRG: ALERT: DETECT: Kernel: _stext hash changed unexpectedly
[    7.577733] LKRG: ALERT: DETECT: Kernel: 1 checksums changed unexpectedly
[    9.027341] LKRG: ALERT: DETECT: Kernel: _stext hash changed unexpectedly
[    9.032777] LKRG: ALERT: DETECT: Kernel: Module hash changed unexpectedly, name nf_conntrack
[    9.032814] LKRG: ALERT: DETECT: Kernel: Module list hash changed unexpectedly
[    9.032844] LKRG: ALERT: DETECT: Kernel: Module KOBJ list hash changed unexpectedly
[    9.032874] LKRG: ALERT: DETECT: Kernel: Module KOBJ hash changed unexpectedly, name nf_conntrack
[    9.032901] LKRG: ALERT: DETECT: Kernel: 5 checksums changed unexpectedly
[   19.947718] LKRG: ALERT: DETECT: Kernel: _stext hash changed unexpectedly
[   19.952714] LKRG: ALERT: DETECT: Kernel: Module hash changed unexpectedly, name nf_nat
[...]

PR #296, opened 2023-11-07
Add support for RHEL 8.8

Adam-pi3:
A new 'text_poke_loc' was backported to RHEL 8.8+. This commit brings the support for such kernels as well as addresses the #295 problem.

### How Has This Been Tested?
Tested on RHEL 9.1 (Alma), RHEL 9.0 (Alma), RHEL 8.7 (Alma), RHEL 8.8 (Alma) and RHEL 8.3 (CentOS)

### DIFF
--- a/src/modules/database/JUMP_LABEL/p_arch_jump_label_transform_apply/p_arch_jump_label_transform_apply.h
+++ b/src/modules/database/JUMP_LABEL/p_arch_jump_label_transform_apply/p_arch_jump_label_transform_apply.h
@@ -62,7 +62,9 @@
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(5,6,0) || \
     defined(P_LKRG_KERNEL_RHEL_VAR_LEN_JUMP_LABEL)
  #if !P_LKRG_KERNEL_HAS_VAR_LEN_JUMP_LABEL || \
-     (defined(P_LKRG_KERNEL_RHEL_VAR_LEN_JUMP_LABEL) && RHEL_RELEASE_CODE < RHEL_RELEASE_VERSION(9,1))
+     (defined(P_LKRG_KERNEL_RHEL_VAR_LEN_JUMP_LABEL) && \
+      RHEL_RELEASE_CODE < RHEL_RELEASE_VERSION(9,1) && \
+      RHEL_RELEASE_CODE < RHEL_RELEASE_VERSION(8,8))
 typedef struct _p_text_poke_loc {
     s32 rel_addr; /* addr := _stext + rel_addr */
     s32 rel32;


solardiz:
Thanks @Adam-pi3. Are you sure this backport is 8.8+ only, and wasn't introduced e.g. in 8.7? Is this based on your testing and what I wrote in #295 only, or also on your own RH kernel code reviews?

Are you sure this one change is sufficient? I ask because there may be more relevant backports introduced in RHEL8 recently (e.g., not only the struct layout, but also the kinds of jump instructions used), which I'm not sure our other build-time checks detect. This uncertainty is why I hadn't made a pull request like this yet, I wanted to review the code changes some more, but haven't completed that yet.

Adam-pi3:
I did verify it on RHEL 9.1 (Alma), RHEL 9.0 (Alma), RHEL 8.7 (Alma), RHEL 8.8 (Alma) and RHEL 8.3 (CentOS) and haven't seen the problem visible in #295 but I did not verify all the changes between the kernels.
Yes this backport is 8.8+ only from what I can see in my test-VMs

solardiz:
Confirmed that the backport affecting struct layout here is 8.8+, and we appear to have other logic for the previous 8.5 to 8.6 change. So the change here looks correct, and I'll merge this.

It is possible that besides the struct layout change, there may be more relevant backports for us to adapt to, but we can review and, if needed, make those changes separately.

solardiz:
Oh, actually looking at the change with 9.x in mind, it should break 9.0. Adam, you write you tested on it, but I see no way that would be right. The condition is now "below 9.1 and below 8.8", which is equivalent to "below 8.8". We probably need "9.0 or below 8.8". I'll fix that.

Adam-pi3:
For my understanding it is related to the order of OR which we have. Because first OR is < 9.1 which 9.0 can fall it will stop there and not proceed 2nd OR. If it is 8.8/8.7 first OR is failed and then moves to 2nd OR which will handle it. I've tested it on RHEL 9.1 (Alma), RHEL 9.0 (Alma), RHEL 8.7 (Alma), RHEL 8.8 (Alma) and RHEL 8.3 (CentOS) and haven't seen the problem

solardiz:
Where are you seeing OR? It's AND in there. And it would also be wrong with OR, as '< 9.1` would also match 8.8 - in fact, that was the problem before this PR.

As to your testing, have you tried enabling LKRG to start on system bootup and rebooting?

solardiz:
Anyway, the problem seen on Alma 8.8 reboots is gone for me with the latest changes (this PR + my fixup commit for 9.0). Did 3 reboots now. So things look good now.

Adam-pi3:
Hm... I retested it again with only mine change and with your changes and in both(?) cases it works well on RHEL 9.1 (Alma), RHEL 9.0 (Alma), RHEL 8.7 (Alma), RHEL 8.8 (Alma) and RHEL 8.3 (CentOS). Well... it fixed so let's keep it as is :)

solardiz:
Adam, were you able to reproduce the issue on 8.8 before these changes?

> in both(?) cases it works well

To me, this suggests you don't have a proper testcase at least for 9.0.

Adam-pi3:
> Adam, were you able to reproduce the issue on 8.8 before these changes?

Yes

> To me, this suggests you don't have a proper testcase at least for 9.0.

In theory, but I used the same test cases to repro https://github.com/lkrg-org/lkrg/commit/0e46fc41a5b385ab89e8f444e26a8d5cf49d03e8 and I did it successfully. I could root-cause it by generating preprocessor macro to see the output structure. Although, I don't want to loose more cycles on it since the problem is solved.

Issue #338, opened 2024-06-01
Inconsistent use of get_seccomp_filter vs. __put_seccomp_filter


solardiz:
On 5.9+, we use `__put_seccomp_filter` instead of the (no longer existent) `put_seccomp_filter`. However, we still pair it with `get_seccomp_filter`. This looks wrong at least for current mainline because `get_seccomp_filter` increments two refcounts:

```c
static void __get_seccomp_filter(struct seccomp_filter *filter)
{
        refcount_inc(&filter->refs);
}

/* get_seccomp_filter - increments the reference count of the filter on @tsk */
void get_seccomp_filter(struct task_struct *tsk)
{
        struct seccomp_filter *orig = tsk->seccomp.filter;
        if (!orig)
                return;
        __get_seccomp_filter(orig);
        refcount_inc(&orig->users);
}
```

whereas `__put_seccomp_filter` decrements just one. Do we possibly trigger a resource leak there? Refcount overflow or hopefully saturation risk, too?

```c
static void __seccomp_filter_orphan(struct seccomp_filter *orig)
{
        while (orig && refcount_dec_and_test(&orig->users)) {
                if (waitqueue_active(&orig->wqh))
                        wake_up_poll(&orig->wqh, EPOLLHUP);
                orig = orig->prev;
        }
}

static void __put_seccomp_filter(struct seccomp_filter *orig)
{
        /* Clean up single-reference branches iteratively. */
        while (orig && refcount_dec_and_test(&orig->refs)) {
                struct seccomp_filter *freeme = orig;
                orig = orig->prev;
                seccomp_filter_free(freeme);
        }
}

static void __seccomp_filter_release(struct seccomp_filter *orig)
{
        /* Notify about any unused filters in the task's former filter tree. */
        __seccomp_filter_orphan(orig);
        /* Finally drop all references to the task's former tree. */
        __put_seccomp_filter(orig);
}
```

Looks like we'd need to use `__seccomp_filter_release` instead, or even the non-static `seccomp_filter_release` (for which we'd need a dummy task struct). Those `wake_up_poll` things don't look like something we normally want to do, but perhaps if it just happened that the seccomp filter was otherwise freed while we held it, then it becomes our responsibility to do this, too?

solardiz:
> Looks like we'd need to use `__seccomp_filter_release` instead, or even the non-static `seccomp_filter_release` (for which we'd need a dummy task struct).

[...]

PR #346, opened 2024-07-06
[Seccomp] Switch to refcount logic for kernels >= 5.9

Adam-pi3:
Starting from kernel 5.9+ function 'put_seccomp_filter' has been inlined and unavailable for hooking. However, internal not-inline function was used to mitigate the problem. Unfortunately, there is no equivalent counter-part function for new hook and the old one looks incompatible which we overlooked. This patch is switching the logic for kernels 5.9+ to custom implementation of refcount logic and it should address the issue reported as #338

### How Has This Been Tested?
Verified on AlmaLinux 8.8 (4.18.0-477.27.2.el8_8.x86_64) and Ubuntu 22.04.4 LTS (6.3.0-060300-generic). Likely more comprehensive tests are needed.

### DIFF
--- a/src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c
+++ b/src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c
@@ -34,6 +34,43 @@ static struct kretprobe p_seccomp_kretprobe = {
     .data_size = sizeof(struct p_seccomp_data),
 };
 
+int p_lkrg_seccomp_init(void) {
+
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5,9,0)
+   P_SYM_INIT(get_seccomp_filter)
+   P_SYM_INIT(put_seccomp_filter)
+#endif
+
+   return P_LKRG_SUCCESS;
+
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5,9,0)
+p_sym_error:
+   return P_LKRG_GENERAL_ERROR;
+#endif
+}
+
+void p_lkrg_seccomp_filter_get(struct task_struct *p_task) {
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5,9,0)
+   P_SYM(p_get_seccomp_filter)(p_task);
+#else
+   struct p_fake_seccomp_filter *p_filter = (struct p_fake_seccomp_filter *)p_task->seccomp.filter;
+
+   if (p_filter)
+      refcount_inc(&p_filter->refs);
+#endif
+}
+
+void p_lkrg_seccomp_filter_put(struct task_struct *p_task) {
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5,9,0)
+   P_SYM(p_put_seccomp_filter)(p_task);
+#else
+   struct p_fake_seccomp_filter *p_filter = (struct p_fake_seccomp_filter *)p_task->seccomp.filter;
+
+   if (p_filter)
+      refcount_dec(&p_filter->refs);
+#endif
+}
+
 /*
  * x86-64 syscall ABI:
  *  *rax - syscall_number
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -424,7 +424,7 @@ static notrace void p_dump_creds(struct p_cred *p_where, const struct cred *p_fr
 #if defined(CONFIG_SECCOMP)
 static notrace void p_dump_seccomp(struct p_seccomp *p_sec, struct task_struct *p_task, char p_force) {
 
-   P_SYM(p_get_seccomp_filter)(p_task);
+   p_lkrg_seccomp_filter_get(p_task);
    p_sec->sec.mode    = p_task->seccomp.mode;   // Mode
    p_sec->sec.filter  = p_task->seccomp.filter; // Filter
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(5,11,0)
@@ -437,12 +437,7 @@ static notrace void p_dump_seccomp(struct p_seccomp *p_sec, struct task_struct *
       p_sec->flag = 0;
    if (p_force)
       p_sec->flag_sync_thread = 0;
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,9,0)
-   P_SYM(p_put_seccomp_filter)(p_task->seccomp.filter);
-#else
-   P_SYM(p_put_seccomp_filter)(p_task);
-#endif
-
+   p_lkrg_seccomp_filter_put(p_task);
 }
 #endif
 
@@ -1373,8 +1368,12 @@ static int p_cmp_tasks(struct p_ed_process *p_orig, struct task_struct *p_curren
 
 #if defined(CONFIG_SECCOMP)
    /* Seccomp */
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5,9,0)
    if (p_orig->p_ed_task.p_sec.flag) { // SECCOMP was enabled so it make sense to compare...
-      P_SYM(p_get_seccomp_filter)(p_current);
+#else
+   if (p_orig->p_ed_task.p_sec.flag && current == p_current) { // SECCOMP was enabled so it make sense to compare...
+#endif
+      p_lkrg_seccomp_filter_get(p_current);
 
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(5,11,0)
       if (test_task_syscall_work(p_current,SECCOMP) != p_orig->p_ed_task.p_sec.flag) {
@@ -1408,11 +1407,7 @@ static int p_cmp_tasks(struct p_ed_process *p_orig, struct task_struct *p_curren
 
       P_CMP_PTR(p_orig->p_ed_task.p_sec.sec.filter, p_current->seccomp.filter, "seccomp filter")
 
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,9,0)
-      P_SYM(p_put_seccomp_filter)(p_current->seccomp.filter);
-#else
-      P_SYM(p_put_seccomp_filter)(p_current);
-#endif
+      p_lkrg_seccomp_filter_put(p_current);
    }
 #endif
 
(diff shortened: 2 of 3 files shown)


solardiz:
As seen from our CI builds, somehow this is trying (and failing) to use `refcount_t` on CentOS 7, I don't immediately see why.

Adam-pi3:
> We should wrap this in compile-time checks like we do for the usage

What type of wrapping are you thinking? We are using refcount API and expecting that `refcount_t` is available. If that's not the case, should we use `atomic`?

Adam-pi3:
The reason I'm asking since `atomic` instead of `refcount` are in a very old kernels. If something is < 5.9 it should use normally exposed API, only new kernel (which guarantees to have `refcount`) is using new logic. It doesn't make sense that compilation is failing for kernels 3.10.xxx

Adam-pi3:
Oh now I see, wrapping the definition around the kernel version :)

src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c
```diff
@@ -34,6 +34,44 @@ static struct kretprobe p_seccomp_kretprobe = {
...
+}
+
+void p_lkrg_seccomp_filter_put(struct task_struct *p_task) {
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5,9,0)
+   P_SYM(p_put_seccomp_filter)(p_task);
+#else
+   struct p_fake_seccomp_filter *p_filter = (struct p_fake_seccomp_filter *)p_task->seccomp.filter;
+
+   if (p_filter)
+      refcount_dec(&p_filter->refs);
```
solardiz, line 71: I think it is unsafe to just `dec`. We have to implement the logic of `__put_seccomp_filter` here. If another part of the kernel releases the only other remaining reference while we were holding ours, then it becomes our responsibility to release the associated resources (that other part of the kernel couldn't do that because of our extra reference). Please re-read my comments on the issue.

src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c
```diff
@@ -34,6 +34,44 @@ static struct kretprobe p_seccomp_kretprobe = {
...
 
+int p_lkrg_seccomp_init(void) {
+
+   P_SYM_INIT(get_seccomp_filter)
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,9,0)
+#define p___put_seccomp_filter p_put_seccomp_filter
+   P_SYM_INIT(__put_seccomp_filter)
+#else
+   P_SYM_INIT(put_seccomp_filter)
+#endif
```
solardiz, line 45: I think we shouldn't lookup these symbols on 5.9+ at all. We should even exclude our pointer variables that would hold those. Since we don't actually call those functions on 5.9+, then why still have a dependency on being able to look them up at all. Besides, not having them on 5.9+ would avoid us inadvertently calling them (with a bug elsewhere in the code).

solardiz (changes requested):
I made two comments. One is about a major bug to fix. The other is about code cleanup.

solardiz:
Oh, also this PR should be just one commit, not several. So next time you update, please amend the first commit and force-push. You can easily squash the second commit into the first one by marking it a `fixup` in `git rebase -i`.

src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c
solardiz: Testing this special case (race with release by another part of the kernel) is tricky, though. And it's not good for us to have untested code in LKRG. Can you come up with a good test case?

src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c
Adam-pi3: Yes, I overlooked that. Thx

src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c
Adam-pi3: To be able to implement seccomp_free we would need to define full seccomp structure because of the potential BPF:
https://elixir.bootlin.com/linux/latest/source/kernel/seccomp.c#L523
```
static inline void [seccomp_filter_free](https://elixir.bootlin.com/linux/latest/C/ident/seccomp_filter_free)(struct [seccomp_filter](https://elixir.bootlin.com/linux/latest/C/ident/seccomp_filter) *[filter](https://elixir.bootlin.com/linux/latest/C/ident/filter))
{
	if ([filter](https://elixir.bootlin.com/linux/latest/C/ident/filter)) {
		[bpf_prog_destroy](https://elixir.bootlin.com/linux/latest/C/ident/bpf_prog_destroy)([filter](https://elixir.bootlin.com/linux/latest/C/ident/filter)->[prog](https://elixir.bootlin.com/linux/latest/C/ident/prog));
		[kfree](https://elixir.bootlin.com/linux/latest/C/ident/kfree)([filter](https://elixir.bootlin.com/linux/latest/C/ident/filter));
	}
}```

Maybe that's the only way? In that case, I guess a bit more research is needed before merging the PR...

Comment (solardiz) [reply src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c:71]: Oh, right. This may make the whole approach infeasible.

Maybe we should keep the `dec` and document the potential resource leak in a comment? Not that I like it...

Comment (solardiz) [reply src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c:71]: Another option is for us to fix only the get/put inconsistency bug, and not the symbol lookup failures our users are occasionally seeing - if so, we could pair `inc` with `__put_seccomp_filter`. Or we could have that as the primary approach, and have a fallback for when symbol lookup fails (either allow for resource leak then, or not monitor seccomp).

Comment (solardiz) [reply src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c:71]: BTW, I don't understand why `__put_seccomp_filter` has a loop over a list of refs when `__get_seccomp_filter` increments only one. It'd be good for us to figure this out.

**Comment by solardiz** [comment]
Adam, as I wrote in a review comment here:

Another option is for us to fix only the get/put inconsistency bug, and not the symbol lookup failures our users are occasionally seeing - if so, we could pair `inc` with `__put_seccomp_filter`.

I think we should do just that in this PR, which would fix #338 and only it.

Let's not try to remove the dependency on this symbol at this same time. We can approach that as a separate task/PR.

File: src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c (hunk 1 above)
Comment (Adam-pi3) [reply src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c:71]: I'm not sure there will be a leak if we keep only the `inc/dec`. If filter is attached to the task_struct, it cannot be removed:
```here is no mechanism for removing filters once they have been applied to a process```
When process dies, that's the only way to remove filters. However, new filters can be added to the existing one, but to be able to do that, you need to go through the `do_seccomp` function which we hook and control.
When new process is created (or thread) and seccomp is already set, then fork() will increment seccomp filter refcount to 1:
https://elixir.bootlin.com/linux/latest/source/kernel/fork.c#L1915
That  means that if the filter is there, it must exist (can't be removed while we operate on it). In theory we do not need any sync logic since we have guarantee that as long as process is not dead, seccomp filters cannot be gone. However, filters can be modified (if current filters allows it, or priv entity does that). Based on that I think it's secure to keep just simple `inc/dec`

Comment (solardiz) [reply src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c:71]: > no mechanism for removing filters once they have been applied to a process

I was wondering where this phrase came from. Found it only in https://lwn.net/Articles/599931/ from 2014. I wonder if this still holds.

> When process dies, that's the only way to remove filters.

If so, what happens if the task dies while we held the seccomp filter reference?

> In theory we do not need any sync logic since we have guarantee that as long as process is not dead, seccomp filters cannot be gone. However, filters can be modified

Does the refcount increase help us in any way in case the filter is modified? This is not locking against concurrent access anyway, and if you say the filter can't be gone at this time anyway, then the refcount increase is not needed.

Let's be consistent. By your logic, it appears that either we don't need even the inc/dec or we need the full refcount decrease logic (including resource freeing on a possible decrease to zero). The middle ground of inc/dec only does not make sense under assumption that the filter can't be gone at this time.

However, there's the case of task dying. For this case, the inc/dec could reduce the impact from accessing freed memory to a resource leak.

Unless you convince me that this can't happen even without refcount inc (task dying and filter freed from under us), I suggest we split this PR in two - at first pair `inc` with `__put_seccomp_filter`. At least this would be consistent with what the kernel itself does, where it merely has `inc` in `__get_seccomp_filter`. So this is also a code state we might end up reverting to if we later find that whatever logic we come up with in the second PR is flawed.

Comment (Adam-pi3) [reply src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c:71]: Sorry for delay in this thread. I did some homework and per my understanding, the process can't be killed while we execute the logic if seccomp verification in the context of the process which we execute on in the kernel. If we are interrupting / hooking kernel call from the current process, and we verify their seccomp rules, the process can't die while is in the kernel. SIGKILL will be added to the list of signals, but process won't die (if other thread decides to die). It means we are safe here. The corner case is if we decide to verify not-currently-running-in-the-kernel process (which can happen in paranoid mode).
Filtered can be only freed when process is dying and this is addressed by the comment above.

Comment (solardiz) [reply src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c:71]: > The corner case is if we decide to verify not-currently-running-in-the-kernel process (which can happen in paranoid mode).

Well, even if you're correct about everything, we do have this mode and thus this corner case, and it is wrong for us to be doing especially unsafe things when in paranoid mode.

I'm really tempted to just exclude our seccomp support on 5.9+ for now.

Comment (Adam-pi3) [reply src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c:71]: I would rather add the check for paranoid mode that if operates on non-current context to not perform seccomp verification but still do it otherwise. What do you think?

Comment (solardiz) [reply src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c:71]: > I would rather add the check for paranoid mode that if operates on non-current context to not perform seccomp verification but still do it otherwise. What do you think?

I'm not entirely happy about that, but I'm OK with it if the code ends up looking sane.

**Comment by Adam-pi3** [comment]
> Oh, also this PR should be just one commit, not several. So next time you update, please amend the first commit and force-push. You can easily squash the second commit into the first one by marking it a `fixup` in `git rebase -i`.

I'm not sure if I did it correctly, but I think it works

**Review by solardiz — APPROVED** [review]
Looks OK'ish to me now, except that the inc/dec should be unneeded if your understanding described in comments here is correct and is insufficient otherwise. However, I suppose it's fine to keep in case the understanding is incorrect, so that impact is likely reduced to a resource leak... unless the struct layout ever changes and we inc/dec something wrong.

**Comment by solardiz** [comment]
I think we should condition this part of `p_lkrg_main.h`:

```c
#if defined(CONFIG_SECCOMP)
   void (*p_get_seccomp_filter)(struct task_struct *p_task);
#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,9,0)
   void (*p_put_seccomp_filter)(struct seccomp_filter *p_filter);
#else
   void (*p_put_seccomp_filter)(struct task_struct *p_task);
#endif
#endif
```

on `#if LINUX_VERSION_CODE < KERNEL_VERSION(5,9,0)` (and indeed remove the `>=` part from inside of this piece).

solardiz:
> I think we should condition this part of `p_lkrg_main.h`:

Done via de5bdc2729e69e681684a96ef2d921e7bbbf7c99

PR #348, opened 2024-08-05
CI: mkosi: Fix the lack of gcc-13 for mainline builds

vt-alt:
### Description
Installing dkms should be generic enough to bring everything needed to build kernel modules.

This is not 100% solution since exact kernel version may still differ:

  $ make ...
  The kernel was built by: x86_64-linux-gnu-gcc-13 (Ubuntu 13.3.0-3ubuntu1) 13.3.0
  You are using:           gcc-13 (Ubuntu 13.3.0-4ubuntu1) 13.3.0

But this is best effort.

### How Has This Been Tested?
Tested on my GA account.

### DIFF
--- a/mkosi.default
+++ b/mkosi.default
@@ -22,8 +22,7 @@ RootSize=4G
 [Packages]
 BuildPackages=
 	diffutils
-	gcc
-	make
+	dkms
 	linux-virtual
 Packages=
 	sysvbanner


solardiz:
Thanks @vt-alt!

Now this is failing with:

```
 /root/src/src/modules/comm_channel/p_comm_channel.c:147:25: error: initialization of 'int (*)(const struct ctl_table *, int,  void *, size_t *, loff_t *)' {aka 'int (*)(const struct ctl_table *, int,  void *, long unsigned int *, long long int *)'} from incompatible pointer type 'int (*)(struct ctl_table *, int,  void *, size_t *, loff_t *)' {aka 'int (*)(struct ctl_table *, int,  void *, long unsigned int *, long long int *)'} [-Werror=incompatible-pointer-types]
  147 |       .proc_handler   = p_sysctl_kint_validate,
      |                         ^~~~~~~~~~~~~~~~~~~~~~
```

which I guess is related to the recent sysctl changes, so I'll need to look into that and fix it.

solardiz:
> Installing dkms should be generic enough to bring everything needed to build kernel modules.

I wonder if we should also edit our documentation in some way. Right now, `README` says:

```
For example, under Debian and Ubuntu you can install all of these with:

        sudo apt-get install make gcc libelf-dev linux-headers-$(uname -r)

and under Red Hat'ish distributions (e.g. RHEL, CentOS, Fedora) with:

        sudo yum install make gcc elfutils-libelf-devel kernel-devel
```

vt-alt:
The example isn't incorrect. Additionally, `dkms` brings some unnecessary packages and is not a requirement. It may be acceptable for CI but is excessive for "Build requirements". However, it may be noted to simplify the installation process, though it might render the "Installing using DKMS" section somewhat redundant.

Issue #351, opened 2024-08-11
aarch64: LKRG: FATAL: Can't hook 'arch_jump_label_transform'


vt-alt:
`lkrg` suddenly become broken on 6.6.45, 6.10.4, 6.1.104 (all released today stable kernels).
```
[00:00:19] ++ insmod lkrg.ko log_level=3
[00:00:19] [    1.032612] lkrg: loading out-of-tree module taints kernel.
[00:00:19] [    1.033327] lkrg: module verification failed: signature and/or required key missing - tainting kernel
[00:00:19] [    1.044763] LKRG: ALIVE: Loading LKRG
[00:00:19] [    1.045508] Freezing user space processes
[00:00:19] [    1.047050] Freezing user space processes completed (elapsed 0.001 seconds)
[00:00:19] [    1.047809] OOM killer disabled.
[00:00:19] [    1.051342] LKRG: ISSUE: [kretprobe] register_kretprobe() for <ovl_dentry_is_whiteout> failed! [err=-2]
[00:00:19] [    1.052404] LKRG: ISSUE: Can't hook 'ovl_dentry_is_whiteout'. This is expected when OverlayFS is not used.
[00:00:19] [    1.054388] LKRG: FATAL: [kretprobe] register_kretprobe() for <arch_jump_label_transform> failed! [err=-2]
[00:00:19] [    1.055477] LKRG: FATAL: Can't hook 'arch_jump_label_transform'
[00:00:19] [    1.056140] LKRG: FATAL: Can't register CPU architecture specific metadata
[00:00:19] [    1.056873] LKRG: FATAL: Can't create database
[00:00:19] [    1.057340] LKRG: DYING: Not loading LKRG (initialization failed)
[00:00:19] [    1.295735] WARNING: Flushing system-wide workqueues will be prohibited in near future.
[00:00:19] [    1.296641] CPU: 7 PID: 141 Comm: insmod Tainted: G           OE      6.6.45-un-def-alt1 #1
[00:00:19] [    1.297570] Hardware name: linux,dummy-virt (DT)
[00:00:19] [    1.298095] Call trace:
[00:00:19] [    1.298378]  dump_backtrace+0xa0/0x148
[00:00:19] [    1.298812]  show_stack+0x20/0x48
[00:00:19] [    1.299188]  dump_stack_lvl+0x48/0x78
[00:00:19] [    1.299603]  dump_stack+0x18/0x30
[...]

PR #352, opened 2024-08-24
Add support for JUMP_LABEL batch-mode feature for ARM64

Adam-pi3:
ARM64 added support for JUMP_LABEL batch-mode feature. This commit adds support for it and addresses #351

### How Has This Been Tested?
Run it under Raspberry Pi 4 (under ALT Linux - thanks to @vt-alt ). I run it for couple of hours under the following condition:

screen 1: run the command `while true; do echo 1 > /sys/kernel/debug/tracing/events/enable && echo 0 > /sys/kernel/debug/tracing/events/enable; done` which enforces JUMP_LABEL batch mode
screen 2: run the command `while true; do sysctl lkrg.trigger=1; sleep 1; done` to enforce integrity validation.

More comprehensive tests are probably needed. However, it seems like that current logic works (at least for now).


solardiz (approved):

solardiz:
I'm OK merging this, but I suggest we discuss a bit more on #351 next.

PR #353, opened 2024-09-20
fix: debian/control: Add build-depends dh-dkms

Avenger-285714:
### Description
According to Debian packaging guidelines, 'dh-dkms' should be added to the build dependencies.

### How Has This Been Tested?
Link: https://github.com/linuxdeepin/developer-center/issues/10661

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -8,7 +8,7 @@ Vcs-Git: https://github.com/lkrg-org/lkrg.git
 Standards-Version: 4.5.1.1
 Rules-Requires-Root: no
 Build-Depends: debhelper-compat (= 13),
- dkms
+ dh-dkms
 
 Package: lkrg
 Architecture: all


solardiz:
Thanks @Avenger-285714!

@morfikov @adrelanos Can you please review this PR (even though it looks trivial)?

Should this change be listed in `debian/changelog`?

morfikov:
Actually, yes. But it should be just `dh-dkms` instead of `dkms, dh-dkms`.

solardiz:
Thank you @morfikov.

@Avenger-285714 Would you like to amend the commit here (and force-push) to use just `dh-dkms` instead of `dkms, dh-dkms` and update the changelog?

adrelanos:
dh-dkms only indeed.

Avenger-285714:
> Thank you @morfikov.
> 
> @Avenger-285714 Would you like to amend the commit here (and force-push) to use just `dh-dkms` instead of `dkms, dh-dkms` and update the changelog?

Done.

solardiz (approved):
Looks good now. This doesn't update the changelog yet, but this can be done separately.

Issue #356, opened 2024-10-19
CI: Docker cross-builds using Ubuntu Jammy for ARM failing


solardiz:
@vt-alt 3 of our builds started failing:

docker arch boot / build (image=arm64v8/ubuntu:jammy opts="-M virt,gic-version=3 -cpu cortex-a57" console=ttyAMA0 ro... (push) Failing after 43s
[Details](https://github.com/solardiz/lkrg/actions/runs/11413995119/job/31762370214)
docker cross build / cross (arm64v8/ubuntu:jammy, libelf-dev linux-headers-generic) (push) Failing after 12s
[Details](https://github.com/solardiz/lkrg/actions/runs/11413995122/job/31762324395)
docker arch boot / build (image=arm32v7/ubuntu:jammy opts="-M virt" console=ttyAMA0 root=/dev/vda kernel="linux-rasp... (push) Failing after 44s
[Details](https://github.com/solardiz/lkrg/actions/runs/11413995119/job/31762370284)

The failures look like this:

```
  #2 [internal] load metadata for docker.io/arm64v8/ubuntu:jammy
  #2 ERROR: no match for platform in manifest: not found
  ------
   > [internal] load metadata for docker.io/arm64v8/ubuntu:jammy:
  ------
  Dockerfile:2
  --------------------
     1 |     # bionic is the latest Ubuntu with i386 support.
     2 | >>> FROM arm64v8/ubuntu:jammy
     3 |     ENV DEBIAN_FRONTEND="noninteractive"
     4 |     RUN apt-get update -y &&     apt-get install -y             gcc             git             make             libelf-dev linux-headers-generic linux-virtual
  --------------------
  ERROR: failed to solve: arm64v8/ubuntu:jammy: failed to resolve source metadata for docker.io/arm64v8/ubuntu:jammy: no match for platform in manifest: not found
```

PR #357, opened 2024-10-19
CI: Fix Docker multi-platform builds

vt-alt:
### Description

Docker now seems to require specifying the platform for ARM. But add it for 386 too to make configuration generic.

Fixes: https://github.com/lkrg-org/lkrg/issues/356

### How Has This Been Tested?
GA


vt-alt:
Ah docker arch boot needs the fix too.

vt-alt:
The grass is green.

solardiz:
Thank you @vt-alt, this was quick! Indeed all green now.

PR #368, opened 2025-03-11
Some exploit detection cleanup

kerneltoast:
Most significantly, the small cleanups from this PR help eliminate further unnecessary {get|put}_task_struct() usage, which can be confusing and make it seem like there's a race on task_struct's lifetime when there isn't.


solardiz:
As I can see, this PR has the exact same commits as the first 4 commits in #369. So we'll probably close it if/once we merge #369. But let's keep it open until that happens.

Issue #371, opened 2025-04-07
Support new kernels without hook-able {override,revert}_creds()


solardiz:
Today someone has posted the following anonymous Quick Comment via the Openwall website:

> I know, you are likely aware of it, but lkrg is broken on latest kernels because of https://github.com/torvalds/linux/commit/a51a1d6bcaa345cc88e738cad468083c4e13aa3b . While I have traced the LKRG change affecting those symbols to 343153fe5ba6af909d61b5490379f0d27e23e15a , I'm unsure if the hooks there can be removed ... since it feels that those hooks cannot be safely redirected to rcu_assign_pointer.

The referenced mainline commit is:

```
commit a51a1d6bcaa345cc88e738cad468083c4e13aa3b
Author:     Christian Brauner <brauner@kernel.org>
AuthorDate: Mon Nov 25 15:10:00 2024 +0100
Commit:     Christian Brauner <brauner@kernel.org>
CommitDate: Mon Dec 2 11:25:09 2024 +0100

    cred: remove old {override,revert}_creds() helpers
    
    They are now unused.
    
    Link: https://lore.kernel.org/r/20241125-work-cred-v2-4-68b9d38bb5b2@kernel.org
    Reviewed-by: Jeff Layton <jlayton@kernel.org>
    Reviewed-by: Jens Axboe <axboe@kernel.dk>
    Signed-off-by: Christian Brauner <brauner@kernel.org>
```

which is roughly when our CI job testing on mainline kernels has failed.

vt-alt:
Is this build error or boot error? "Helpers removal" looked like it should produce a build error. Building against latest kernel are still performed in `builds.yml`. Ah, quick grep showed you seems intercepting `override_creds`, not linking to it, so simple build may not trigger an error.

solardiz:
> intercepting `override_creds`, not linking to it, so simple build may not trigger an error.

Exactly. We're hooking those functions, not calling them. This probably results in a FATAL message on loading LKRG, which our previous CI job should have detected (if it continued working for newer kernels).

solardiz:
The start-of-thread message from the referenced kernel commit is:

```
From: Christian Brauner <brauner@kernel.org>
To: Linus Torvalds <torvalds@linux-foundation.org>
Cc: Amir Goldstein <amir73il@gmail.com>,
	Miklos Szeredi <miklos@szeredi.hu>,
[...]

PR #381, opened 2025-04-26
Support new kernels without hook-able {override,revert}_creds()

solardiz:
### Description
This works by not validating the effective credentials when the pointer differs, which may indicate that override_creds() is in effect or that the pointer was overwritten in an attack.  So with this workaround we'll be blind to cred pointer overwrite attacks, but would still detect overwrites of the creds themselves if the pointer is kept intact.  We should also still detect exploits' use of commit_creds() as that function also replaces real_cred, which we continue to validate in all cases.

Works around #371

### How Has This Been Tested?
On Arch Linux with 6.14.3-arch1-1, running SUID programs such as `su` and `passwd` on the Arch system itself, as well as in a Docker container with Alpine Linux. Without the relaxed validation, all of these triggered alerts and the commands were killed. With this PR's changes, they just work correctly and nothing appears in the logs.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,7 @@
 The following major changes have been made since 0.9.9:
 
+ *) Support Linux 6.14+ where we can't hook {override,revert}_creds() anymore,
+    by limiting detection of cred pointer overwrite attacks on those kernels
  *) Overhaul locking of per-task shadow data, using finer-grain locks
  *) Improve performance of per-task shadow data lookups by making them lockless
  *) Fix several lethal race conditions involving SECCOMP_FILTER_FLAG_TSYNC
@@ -10,7 +12,6 @@ The following major changes have been made since 0.9.9:
     between profile_validate=4 and others
  *) Make kprobes testing via LKRG's own dummy function hooking optional (works
     around issue seen on recent Gentoo)
- *) Support building against (but not yet loading into) Linux 6.15-rc1+
  *) Build and link the userspace logger tools with hardening flags, and pass
     distributions' RPM packaging hardening flags to the compiler and linker
  *) lkrg-logctl: Support and report continuation lines (an extra one-character
--- a/CHANGES
+++ b/CHANGES
@@ -1,7 +1,7 @@
 The following major changes have been made since 0.9.9:
 
- *) Support Linux 6.14+ where we can't hook {override,revert}_creds() anymore,
-    by limiting detection of cred pointer overwrite attacks on those kernels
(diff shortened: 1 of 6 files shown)


solardiz:
I've just additionally tested that on an older kernel (for which I have a working exploit) a credentials overwrite (keeping the pointer intact) is still detected despite of these changes, separately (two tests) with the changes as-is and with the checks for 6.14+ hacked to check for 4.14+ (arbitrary) - the latter simulating what would happen on a 6.14+ kernel now.

solardiz:
I realized that we can continue to check `cred` when `real_cred` differs, so added that. Also added a lengthy comment.

solardiz:
I've also just added this commit:

```
    Detect some cred pointer overwrites even on 6.13+
    
    by checking the rest of the cred struct fields that we track (currently
    just user_ns) even when we can't insist on the UIDs/GIDs being correct.
```

and it passes my tests so far, but I think I'll remove it because of concern that some overrides may include `user_ns`. For example, in `fs/nfs/nfs4idmap.c` I see that `override_creds` is used on:

```c
        cred = prepare_kernel_cred(&init_task);
```

so it could be switching from the current task's ns to init's.

solardiz:
> ```
>     Detect some cred pointer overwrites even on 6.13+
>     
>     by checking the rest of the cred struct fields that we track (currently
>     just user_ns) even when we can't insist on the UIDs/GIDs being correct.
> ```
> 
> and it passes my tests so far, but I think I'll remove it

Removed. We could maybe later add some kind of sanity-checking of the cred struct, e.g. that it contains valid kernel pointers where expected. But I'm not sure there's much value in that as an exploit that already runs a ROP chain could just use `prepare_kernel_cred` to make the struct look just as valid as legitimate override creds do. Maybe if we check pCFI on `prepare_kernel_cred` return, but for that to make sense we need to review the kernel for other similar gadgets.

solardiz:
Rebased on top of two other just-merged PRs. There was a conflict in `CHANGES`.

solardiz:
> I've just additionally tested that on an older kernel (for which I have a working exploit) a credentials overwrite (keeping the pointer intact) is still detected despite of these changes, separately (two tests) with the changes as-is and with the checks for 6.14+ hacked to check for 4.14+ (arbitrary) - the latter simulating what would happen on a 6.14+ kernel now.

I've just repeated the equivalent of these tests for the latest code in this PR. Also ran them with profile_validate=4, in addition to with defaults. All of these 4 tests passed - exploit detected and killed, nothing unexpected happened.

PR #369, opened 2025-03-28
Introduce a faster and safer ed task tree

kerneltoast:
~~NOTE: Doesn't support kernels <4.2 right now (see comment at the top of ed_task_tree.c). I'd really like not to carve up ed_task_tree.c with a rwlock_t fallback for ancient kernels, but will do so if needed once everything else looks good to merge.~~

EDIT: Kernels <4.2 are supported now.

PR #383, opened 2025-04-28
seccomp: Invoke p_set_ed_process_on() with non-zero flag_sync_thread

kerneltoast:


kerneltoast:
To fully fix #329, @solardiz's [diff from this comment is also needed.](https://github.com/lkrg-org/lkrg/issues/329#issuecomment-2836299095)

solardiz:
My current understanding is that even these two fixes together don't fully fix #329 because other places in the code exist that may attempt to set `off` without override and would stumble upon it already being set by seccomp and would get upset about that.

kerneltoast:
> My current understanding is that even these two fixes together don't fully fix #329 because other places in the code exist that may attempt to set `off` without override and would stumble upon it already being set by seccomp and would get upset about that.

With both fixes together, the override is always engaged when seccomp has the off flag set. So I'm not seeing how "other places in the code exist that may attempt to set `off` without override" is possible, could you elaborate?

solardiz:
> With both fixes together, the override is always engaged when seccomp has the off flag set. So I'm not seeing how "other places in the code exist that may attempt to set `off` without override" is possible, could you elaborate?

Oh, you seem to be right! Somehow I was thinking the override is just for the call from seccomp, but it's actually via a field that's also used by other calls.

solardiz:
Here's the commit you asked for. I think you can `git am` it and then rebase yours on top.

```
From 09f5557fd379f3b4316f5e7430cb2510d11bf8f0 Mon Sep 17 00:00:00 2001
From: Solar Designer <solar@openwall.com>
Date: Tue, 29 Apr 2025 04:22:28 +0200
Subject: [PATCH] seccomp: Invoke p_set_ed_process_on() with non-zero
 flag_sync_thread

Have p_set_ed_process_on() invoked with pre-decrement flag_sync_thread,
because it checks that value to choose the "override" path or not.  It
could report off flag corruption when it chooses the non-override path.

diff --git a/src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c b/src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c
index f70ba94..fa453b3 100644
--- a/src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c
+++ b/src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c
@@ -69,10 +69,11 @@ static __always_inline void p_seccomp_filter_tsync(bool entry) {
                p_set_ed_process_off(p_tmp);
          } else {
             /* Clear the off flag once the tsync depth hits zero */
-            if (!--p_tmp->p_ed_task.p_sec.flag_sync_thread) {
+            if (p_tmp->p_ed_task.p_sec.flag_sync_thread == 1) {
                p_update_ed_process(p_tmp, thread, 0);
                p_set_ed_process_on(p_tmp);
             }
+            p_tmp->p_ed_task.p_sec.flag_sync_thread--;
          }
          ed_task_unlock(p_tmp);
       }
```

kerneltoast:
> Here's the commit you asked for. I think you can `git am` it and then rebase yours on top.
> 
> ```
> From 09f5557fd379f3b4316f5e7430cb2510d11bf8f0 Mon Sep 17 00:00:00 2001
> From: Solar Designer <solar@openwall.com>
> Date: Tue, 29 Apr 2025 04:22:28 +0200
> Subject: [PATCH] seccomp: Invoke p_set_ed_process_on() with non-zero
>  flag_sync_thread
> 
> Have p_set_ed_process_on() invoked with pre-decrement flag_sync_thread,
> because it checks that value to choose the "override" path or not.  It
> could report off flag corruption when it chooses the non-override path.
> 
> diff --git a/src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c b/src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c
> index f70ba94..fa453b3 100644
> --- a/src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c
> +++ b/src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c
> @@ -69,10 +69,11 @@ static __always_inline void p_seccomp_filter_tsync(bool entry) {
>                 p_set_ed_process_off(p_tmp);
>           } else {
>              /* Clear the off flag once the tsync depth hits zero */
> -            if (!--p_tmp->p_ed_task.p_sec.flag_sync_thread) {
> +            if (p_tmp->p_ed_task.p_sec.flag_sync_thread == 1) {
>                 p_update_ed_process(p_tmp, thread, 0);
>                 p_set_ed_process_on(p_tmp);
>              }
> +            p_tmp->p_ed_task.p_sec.flag_sync_thread--;
>           }
>           ed_task_unlock(p_tmp);
>        }
> ```

Just did and updated the PR, thanks.

solardiz:
I'm now testing this in two VMs, one is Rocky Linux 9.5, the other is Arch Linux with `6.14.3-arch1-1` and a Docker container with Alpine. No issues so far. But these systems didn't exhibit any issues prior to this PR as well.

@kerneltoast It'd be good if you could come up with a test case for seccomp specifically. Either our own program or some kind of usage of something pre-existing that would be expected to expose the issues we're trying to fix here - ideally, something that's convenient to run from the command-line and can be put into a shell loop and left running in screen/tmux (so web browsers are not ideal).

solardiz:
Oh, actually there is `tools/testing/selftests/seccomp/` in the Linux kernel tree. I'm running these on the Rocky system now.

solardiz:
> Oh, actually there is `tools/testing/selftests/seccomp/` in the Linux kernel tree. I'm running these on the Rocky system now.

No differences in tests' behavior observed so far for 3 cases: LKRG prior to this PR, LKRG with this PR, and no LKRG. Now running 4 concurrent shell loops with them (2 of each program) with LKRG from just prior to this PR, hoping to trigger any issue.

kerneltoast:
> > Oh, actually there is `tools/testing/selftests/seccomp/` in the Linux kernel tree. I'm running these on the Rocky system now.
> 
> No differences in tests' behavior observed so far for 3 cases: LKRG prior to this PR, LKRG with this PR, and no LKRG. Now running 4 concurrent shell loops with them (2 of each program) with LKRG from just prior to this PR, hoping to trigger any issue.

I think the most strenuous test would involve a single group of 2-3 threads that constantly call seccomp() with _and_ without SECCOMP_FILTER_FLAG_TSYNC, with each thread randomly alternating between using SECCOMP_FILTER_FLAG_TSYNC and not using it.

solardiz:
> Now running 4 concurrent shell loops with them (2 of each program) with LKRG from just prior to this PR

These 4 are running as a user. Added a 5th running as root. Still nothing. BTW, this is on 4 logical CPUs.

> I think the most strenuous test would involve [...]

Well, I'm not going to develop it now. Maybe you'd like to? ;-)

solardiz:
> These 4 are running as a user. Added a 5th running as root. Still nothing. BTW, this is on 4 logical CPUs.

Kept these tests running, but replaced LKRG with this PR's. Still nothing.

solardiz:
> The off flag can be set twice for a task via seccomp

Was this really an issue? How is the non-tsync setting of the flag is any different from setting it from other places, which we've already determined can result in it being set twice and @kerneltoast convinced me is fine because if this happens after tsync's setting of it then override paths will be enabled automatically in those other places.

kerneltoast:
> > The off flag can be set twice for a task via seccomp
> 
> Was this really an issue? How is the non-tsync setting of the flag is any different from setting it from other places, which we've already determined can result in it being set twice and @kerneltoast convinced me is fine because if this happens after tsync's setting of it then override paths will be enabled automatically in those other places.

Yeah, it was an issue because the off flag could remain incremented by cookie after the tsync override is gone.

Bug:
1. Thread A calls seccomp without tsync. Off flag is incremented by one. Off=1
2. Thread B (in the same thread group) calls seccomp with tsync. Off flag is incremented by 1. 

> > The off flag can be set twice for a task via seccomp
> 
> Was this really an issue? How is the non-tsync setting of the flag is any different from setting it from other places, which we've already determined can result in it being set twice and @kerneltoast convinced me is fine because if this happens after tsync's setting of it then override paths will be enabled automatically in those other places.

Great point! I'll drop `Ensure seccomp sets the off flag only up to one time per task` from this PR, and add the Fixes tag into your commit message.

kerneltoast:
> > The off flag can be set twice for a task via seccomp
> 
> Was this really an issue? How is the non-tsync setting of the flag is any different from setting it from other places, which we've already determined can result in it being set twice and @kerneltoast convinced me is fine because if this happens after tsync's setting of it then override paths will be enabled automatically in those other places.

Great point! I'll drop Ensure seccomp sets the off flag only up to one time per task from this PR, and add the Fixes tag into your commit message.

(PS: I had started typing out something earlier and thought I deleted it but my comment got posted with it, so I deleted that comment and made this new one to say what I had intended)

solardiz:
> Kept these tests running, but replaced LKRG with this PR's. Still nothing.

This still produced nothing new (was still running). Now replaced LKRG with the updated PR (one commit only).

Something I didn't mention before: whether with LKRG or not, only 99 out of 106 tests succeed. But this is stable, and unrelated to our work here. Also, running these tests produces frequent coredumps, maybe as intended or not. So LKRG's `profile_validate=4` results in many of:

```
[1467739.950134] LKRG: ALERT: BLOCK: UMH: Executing program name /usr/lib/systemd/systemd-coredump
[1467740.377630] LKRG: ALERT: BLOCK: UMH: Executing program name /usr/lib/systemd/systemd-coredump
[1467740.382161] LKRG: ALERT: BLOCK: UMH: Executing program name /usr/lib/systemd/systemd-coredump
[1467740.386596] LKRG: ALERT: BLOCK: UMH: Executing program name /usr/lib/systemd/systemd-coredump
[1467740.391168] LKRG: ALERT: BLOCK: UMH: Executing program name /usr/lib/systemd/systemd-coredump
[1467740.394828] LKRG: ALERT: BLOCK: UMH: Executing program name /usr/lib/systemd/systemd-coredump
[1467740.398382] LKRG: ALERT: BLOCK: UMH: Executing program name /usr/lib/systemd/systemd-coredump
[1467740.409125] LKRG: ALERT: BLOCK: UMH: Executing program name /usr/lib/systemd/systemd-coredump
[1467740.416266] LKRG: ALERT: BLOCK: UMH: Executing program name /usr/lib/systemd/systemd-coredump
[1467741.458161] LKRG: ALERT: BLOCK: UMH: Executing program name /usr/lib/systemd/systemd-coredump
```

while running these seccomp tests bundled with the kernel. (Latest kernel's tests on Rocky Linux 9.5 kernel.)

solardiz:
I'm not entirely sure this fully fixes the linked issue, but I'm going to merge it so that it receives more testing sooner rather than later. I'm also going to update the Rocky Linux SIG/Security package to latest git afterwards.

solardiz:
I guess this doesn't need a separate `CHANGES` entry because we already have:

```
*) Fix several lethal race conditions involving SECCOMP_FILTER_FLAG_TSYNC
```

However, we could possibly want to mention the user-visible effect, if we have in fact fixed #329. It is likely that multiple changes would have contributed to fixing that issue, not just this latest fix.

kerneltoast:
> I'm not entirely sure this fully fixes the linked issue [...]

If it doesn't, I think a new issue should be made instead since so much has changed to the point that any old reports of spurious off flag corruption are invalid. 

Combined with the fixes from PR #369, I think it makes sense to finally close the linked issue with this PR. Especially since the backtraces that were reported post-#369 implicate seccomp.

PR #386, opened 2025-05-01
UMH: Add missing call to unset off flag

kerneltoast:


solardiz:
Thank you @kerneltoast!

When was this bug introduced?

If this was in fact the problem, then why didn't it show up e.g. also on many UMH invocations of `systemd-coredump` (some were allowed, some not, as I was trying different LKRG settings) that I had in testing just a day before?

solardiz:
In the last 2 years, this file changed only slightly:

```diff
+++ b/src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
@@ -74,7 +74,6 @@ int p_call_usermodehelper_entry(struct kretprobe_instance *p_ri, struct pt_regs
    struct p_ed_process *p_tmp;
    struct subprocess_info *p_subproc = (struct subprocess_info *)p_regs_get_arg1(p_regs);
    unsigned char p_umh_allowed = 0;
-   unsigned long p_flags;
    size_t i;
 
    p_ed_enforce_validation();
@@ -84,13 +83,11 @@ int p_call_usermodehelper_entry(struct kretprobe_instance *p_ri, struct pt_regs
    } else if (P_CTRL(p_umh_validate) == 2) {
       goto p_call_usermodehelper_entry_not_allowed;
    }
-
-   p_tasks_write_lock(&p_flags);
-   if ( (p_tmp = p_find_ed_by_pid(task_pid_nr(current))) != NULL) {
+   if ((p_tmp = ed_task_lock_current())) {
       // This process is on the ED list - set temporary 'disable' flag!
       p_set_ed_process_off(p_tmp);
+      ed_task_unlock(p_tmp);
    }
-   p_tasks_write_unlock(&p_flags);
 
    for (i = 0; i < nitems(p_umh_global); i++) {
       if (!strcmp(p_umh_global[i], p_subproc->path)) {
```

Do you think the above somehow exposed the issue? Or do you know some other change that did?

solardiz:
> Or do you know some other change that did?

We may look for a possible EL9 kernel change that did, as I also upgraded the kernel from older EL9 to latest EL9 on the system where I first saw this issue.

solardiz:
> why didn't it show up e.g. also on many UMH invocations of `systemd-coredump`

Also no problem (with our latest merged code without this PR) on the very machine that exhibited the issue for `modprobe` during bootup when I try triggering coredumps with the machine already up, tried both with invocations of `systemd-coredump` allowed or blocked. So it looks like there's something special about those `modprobe`.

kerneltoast:
> ```diff
> +++ b/src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
> @@ -74,7 +74,6 @@ int p_call_usermodehelper_entry(struct kretprobe_instance *p_ri, struct pt_regs
>     struct p_ed_process *p_tmp;
>     struct subprocess_info *p_subproc = (struct subprocess_info *)p_regs_get_arg1(p_regs);
>     unsigned char p_umh_allowed = 0;
> -   unsigned long p_flags;
>     size_t i;
>  
>     p_ed_enforce_validation();
> @@ -84,13 +83,11 @@ int p_call_usermodehelper_entry(struct kretprobe_instance *p_ri, struct pt_regs
>     } else if (P_CTRL(p_umh_validate) == 2) {
>        goto p_call_usermodehelper_entry_not_allowed;
>     }
> -
> -   p_tasks_write_lock(&p_flags);
> -   if ( (p_tmp = p_find_ed_by_pid(task_pid_nr(current))) != NULL) {
> +   if ((p_tmp = ed_task_lock_current())) {
>        // This process is on the ED list - set temporary 'disable' flag!
>        p_set_ed_process_off(p_tmp);
> +      ed_task_unlock(p_tmp);
>     }
> -   p_tasks_write_unlock(&p_flags);
>  
>     for (i = 0; i < nitems(p_umh_global); i++) {
>        if (!strcmp(p_umh_global[i], p_subproc->path)) {
> ```
> 
> Do you think the above somehow exposed the issue?

> Or do you know some other change that did?

I don't. I kind of suspect this bug rarely appeared in the wild because it requires triggering the UMH after LKRG is loaded, along with some bad luck for the UMH program to hit another LKRG path that sets the off flag after the off flag is set for UMH and never unset.

To trigger the UMH, the most obvious example would be enumeration of new hardware (e.g. plugging something into the machine). Which is not exceedingly common after boot.

solardiz:
> bad luck for the UMH program to hit another LKRG path that sets the off flag after the off flag is set for UMH and never unset.

If it's so rare for a UMH program to hit such path, it suggests we may not be checking "`off` must be off" in some more places where we could (and already hook).

> To trigger the UMH, the most obvious example would be enumeration of new hardware (e.g. plugging something into the machine). Which is not exceedingly common after boot.

Yet common enough that we'd have seen it.

Here's another guess: besides the kernel, I also updated `kmod` on this system. Maybe the new `modprobe` does something different.

Adam-pi3:
I'm on the plane so can't check all the details. However, call_usermodehelper_exec_async calls exec* which we hook and during that hook we reset 'off' flag so I don't think setting it on is needed. Can you double check that path?

solardiz:
> However, call_usermodehelper_exec_async calls exec* which we hook and during that hook we reset 'off' flag so I don't think setting it on is needed. Can you double check that path?

Thank you for your comment @Adam-pi3. Yes, I had thought it'd be something like this, but I can't quickly find it now. Before d3276d45e7631288d7d7f060c315e1228160c560 (Jan 2021) we hooked `search_binary_handler` or `do_execveat_common` and we used the "override" versions of the `off` functions there. However, after that commit we hook `security_bprm_committing_creds` and `security_bprm_committed_creds`, which use non-"override", so I don't see why it wouldn't trigger the "corruption" detection.

src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.h
```diff
@@ -24,7 +24,7 @@
 
 /* per-instance private data */
 struct p_call_usermodehelper_data {
-    ktime_t entry_stamp;
+   bool off_flag_set;
```
solardiz, line 27: BTW, what's the deal with those `entry_stamp` fields? Looks like we don't use them at all, anywhere. Perhaps we should clean this up? (Not in this PR indeed.)

solardiz: Looks like those fields are a leftover from `samples/kprobes/kretprobe_example.c`. Also, none of these structs have any other fields. So maybe we want to remove the structs - except here in UMH, if we do start to use this one now.

solardiz:
Assuming what this PR tries to do is actually right and required, I think it does this too late (the new program is started _before_ `call_usermodehelper_exec` returns, but here we hook and unset the flag on this function's return) and maybe also not for the right task (not the ultimate child thread).

solardiz:
Actually, we had logic similar to this PR's before, but @Adam-pi3 removed it in c2c27f077fa18b9fa2455b8acfe72950eb180026 with no explanation in the commit message. Maybe it was in our emails...

solardiz:
> we had logic similar to this PR's before, but @Adam-pi3 removed it in [c2c27f0](https://github.com/lkrg-org/lkrg/commit/c2c27f077fa18b9fa2455b8acfe72950eb180026) with no explanation

Now that I realize we actually hook `call_usermodehelper_exec_async`, I assume the reason is that the task exits before the function returns, so `off` flag no longer matters.

Adam-pi3:
>Actually, we had logic similar to this PR's before, but @Adam-pi3 removed it in https://github.com/lkrg-org/lkrg/commit/c2c27f077fa18b9fa2455b8acfe72950eb180026 with no explanation in the commit message. Maybe it was in our emails...

I don't remember the actual reasoning behind that (it was too long time ago, but certainly there was a reason behind it).

Nevertheless, about this PR, I'm wondering if we do the right think here. We do disable `off` at the entrance of `call_usermodehelper_exec_async` function which at some point calls `kernel_execve`. This function will call at some point `security_bprm_committed_creds` which will reset the off flag (so effectively cancels `p_set_ed_process_off`).
There is a caveat when `kernel_execve` fails (it is possible). In that case `call_usermodehelper_exec_async` calls `exit` which we hook too and remove the task from ED list.
What is worth to keep in mind is `kernel_execve` overwrite the current process metadata (that's why we call an `update`) and if this PR adds `p_set_ed_process_on` after we reset the flag, it actually desynchronize it and set the `off` into invalid state. However, I would need to spin-up a kernel debugger to analyze the full flow to confirm or reject the theory which I can't do it now (I will be travelling for entire May :(). I'm wondering @solardiz or @kerneltoast can you verify / debug that?

solardiz:
@Adam-pi3 Good point about `security_bprm_committed_creds` resetting the flag, which I see it does with `p_reset_ed_flags`.

@kerneltoast I think we should close this PR as unneeded and potentially harmful, do you agree?

PR #387, opened 2025-05-02
UMH: Allow "off" to be set a second time

solardiz:
### Description

Allow `off` to be set a second time. First from our hook of `call_usermodehelper_exec_async` and then also from our hook of `security_bprm_committing_creds`.  Without this hack, the latter would falsely detect `off` flag corruption due to our "`off` most be off" check.

Works around #385

### How Has This Been Tested?

By multiple reboots of the bare metal system where the issue was almost reliably reproducible. With this change, the issue is not seen on that system anymore.

### DIFF
--- a/src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
+++ b/src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
@@ -84,6 +84,26 @@ int p_call_usermodehelper_entry(struct kretprobe_instance *p_ri, struct pt_regs 
       goto p_call_usermodehelper_entry_not_allowed;
    }
    if ((p_tmp = ed_task_lock_current())) {
+      /*
+       * Dirty hack to allow "off" to be set a second time from our hook of
+       * security_bprm_committing_creds().
+       *
+       * We do this before rather than after p_set_ed_process_off() here to
+       * work around known glitching where we'd be accessing a wrong ed task
+       * record here (as well as later).
+       *
+       * These hacks force p_set_ed_process_off() to take the "override" path
+       * both here and in the second call (from the later hook), so that it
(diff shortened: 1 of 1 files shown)


solardiz:
> p_set_ed_process_off(p_tmp);
> p_tmp->p_ed_task.p_sec.flag_sync_thread++;

I think we may want to swap these two lines. The way I currently put them (as above) in the PR is more correct in that it insists on "`off` must be off" before we got here, which is just right when everything else is working properly. However, we know that some other (than current) shadow task is somehow picked here, and it may already be in `off` for whatever other reasons. So we may want to start with the override for now. Makes sense to you, @kerneltoast @Adam-pi3?

solardiz:
> I think we may want to swap these two lines.

I'm testing this on the "problematic" machine now (doing several reboots, OK so far).

solardiz:
> I think we may want to swap these two lines.

I've just amended the commit here to do that, and also explained it in the comment.

I'll likely merge this PR once our CI bots are happy. Then will proceed to update and retest the package (which is how I had discovered the issue).

kerneltoast:
`p_reset_ed_flags()` doesn't reset `p_tmp->p_ed_task.p_sec.flag_sync_thread`, so the flag_sync_thread increment persists for the entire lifetime of the task, FWIW.

solardiz:
> `p_reset_ed_flags()` doesn't reset `p_tmp->p_ed_task.p_sec.flag_sync_thread`,

Do you think it should? Is it a safe thing to do?

> so the flag_sync_thread increment persists for the entire lifetime of the task, FWIW.

Yes, that's my expectation. If we hit the right task here, then this is fine - the task will exit soon anyway. If we hit the wrong task (like we seem to most or all of the time when we manage to hit a task at all, which is rare), well, we have to break its protection in order for us not to run into false positives further in UMH.

We really ought to figure this issue out and fix it for real.

PR #389, opened 2025-05-06
Ignore redundant task_struct argument when p_ed_process is provided

kerneltoast:


kerneltoast:
@solardiz Although this PR didn't fix #385, it is still an important change that prevents issues like the one fixed in https://github.com/lkrg-org/lkrg/commit/cc12744d05d96572d75c4b14421b5af96dbc39d7 from occurring again. Could we work on getting this merged?

solardiz:
> Could we work on getting this merged?

Yes, I think we should.

Reviewing these commits, I see that the "Ignore ..." one also introduces but does not yet use `p_pcfi_kill_ed_task`. Then the "Apply ..." commit makes use of that function, but leaves `p_pcfi_kill_task` unused. Perhaps we should be removing `p_pcfi_kill_task` somewhere - maybe in an extra commit yet to be added. Also, maybe the introduction of not-yet-used `p_pcfi_kill_ed_task` should be in its own commit or at least documented in the commit message. I understand and agree that it shouldn't be in the same commit that puts it to use because of separation of manual edits vs. coccinelle.

kerneltoast:
> Reviewing these commits, I see that the "Ignore ..." one also introduces but does not yet use `p_pcfi_kill_ed_task`. Then the "Apply ..." commit makes use of that function, but leaves `p_pcfi_kill_task` unused.

Good catch, I've updated the PR to put the introduction of `p_pcfi_kill_ed_task` into its own commit, and tacked on an extra commit to remove the dead `p_pcfi_kill_task` helper.

solardiz:
Thank you @kerneltoast! The changes also look good to me.

PR #391, opened 2025-05-08
Hooking: Switch many hooks to simple kprobes

solardiz:
### Description
This cleans up (and maybe speeds up) our use of kretprobes, introduces support for simple kprobes into our shared hooking functions, and actually switches many of the hooks to simple kprobes.

### How Has This Been Tested?
Works fine so far in the VM where I develop this. I also checked `log_level=4` messages.

### DIFF
--- a/src/modules/exploit_detection/syscalls/p_install.h
+++ b/src/modules/exploit_detection/syscalls/p_install.h
@@ -23,7 +23,7 @@
 
 struct lkrg_probe {
    enum { LKRG_PROBE_OFF, LKRG_PROBE_ON, LKRG_PROBE_ISRA } state;
-   enum { LKRG_KRETPROBE } type;
+   enum { LKRG_KPROBE, LKRG_KRETPROBE } type;
    struct kretprobe krp;
 };
 
--- a/src/modules/exploit_detection/syscalls/p_install.h
+++ b/src/modules/exploit_detection/syscalls/p_install.h
@@ -21,18 +21,24 @@
 #ifndef P_LKRG_EXPLOIT_DETECTION_INSTALL_H
 #define P_LKRG_EXPLOIT_DETECTION_INSTALL_H
 
+struct lkrg_probe {
+   enum { LKRG_PROBE_OFF, LKRG_PROBE_ON, LKRG_PROBE_ISRA } state;
+   enum { LKRG_KRETPROBE } type;
+   struct kretprobe krp;
+};
+
 int p_get_kprobe_maxactive(void);
 
-int p_install_hook(struct kretprobe *kretprobe, char *state, int p_isra);
-void p_uninstall_hook(struct kretprobe *kretprobe, char *state);
+int p_install_hook(struct lkrg_probe *probe, int p_isra);
+void p_uninstall_hook(struct lkrg_probe *probe);
 
 #define GENERATE_INSTALL_FUNC(name)                                                       \
    int p_install_##name##_hook(int p_isra) {                                              \
-      return p_install_hook(&p_##name##_kretprobe, &p_##name##_kretprobe_state, p_isra);  \
+      return p_install_hook(&p_##name##_probe, p_isra);                                   \
    }                                                                                      \
                                                                                           \
    void p_uninstall_##name##_hook(void) {                                                 \
-      return p_uninstall_hook(&p_##name##_kretprobe, &p_##name##_kretprobe_state);        \
+      return p_uninstall_hook(&p_##name##_probe);                                         \
    }
 
 #endif
--- a/src/modules/exploit_detection/syscalls/p_install.h
+++ b/src/modules/exploit_detection/syscalls/p_install.h
@@ -41,4 +41,8 @@ void p_uninstall_hook(struct lkrg_probe *probe);
       return p_uninstall_hook(&p_##name##_probe);                                         \
    }
 
+#define GENERATE_INSTALL_FUNC_PROTO(name) \
+   extern int p_install_##name##_hook(int p_isra); \
+   extern void p_uninstall_##name##_hook(void);
+
 #endif
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,8 @@ The following major changes have been made since 0.9.9:
 
  *) Support Linux 6.13+ by not hooking {override,revert}_creds() anymore, and
     limiting detection of cred pointer overwrite attacks on those kernels
+ *) Switch many hooks from kretprobes to simple kprobes for greater reliability
+    and improved performance
  *) Overhaul locking of per-task shadow data, using finer-grain locks
  *) Improve performance of per-task shadow data lookups by making them lockless
  *) Fix several lethal race conditions involving SECCOMP_FILTER_FLAG_TSYNC
--- a/src/modules/database/FTRACE/p_ftrace_enable_sysctl/p_ftrace_enable_sysctl.c
+++ b/src/modules/database/FTRACE/p_ftrace_enable_sysctl/p_ftrace_enable_sysctl.c
@@ -35,7 +35,6 @@ static struct kretprobe p_ftrace_enable_sysctl_kretprobe = {
     .kp.symbol_name = "ftrace_enable_sysctl",
     .handler = p_ftrace_enable_sysctl_ret,
     .entry_handler = p_ftrace_enable_sysctl_entry,
-    .data_size = sizeof(struct p_ftrace_enable_sysctl_data),
 };
 
 notrace int p_ftrace_enable_sysctl_entry(struct kretprobe_instance *p_ri, struct pt_regs *p_regs) {
--- a/src/modules/database/FTRACE/p_ftrace_enable_sysctl/p_ftrace_enable_sysctl.c
+++ b/src/modules/database/FTRACE/p_ftrace_enable_sysctl/p_ftrace_enable_sysctl.c
@@ -29,62 +29,23 @@
 
 #if defined(P_LKRG_FTRACE_ENABLE_SYSCTL_H)
 
-char p_ftrace_enable_sysctl_kretprobe_state = 0;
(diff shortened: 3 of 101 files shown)


solardiz:
It was a bit late during my work on this that I realized `struct kretprobe` includes a `struct kprobe` inside of it. I actually prepared an anonymous `union` of both structs and code to use it by that time - now dropped/simplified. Even more simplifications can be made with this understanding - instead of the `type` field, we could infer the probe type based on which pointer field(s) are set. This would reduce the amount of changes to struct initializers now, but OTOH adding yet another probe type (neither kprobe nor kretprobe) would then be a little more work (we'd need to reintroduce the `type` field and `union` then).

kerneltoast:
All of the probe scaffolding can be simplified by creating dedicated sections to store each probe type. With that, all install-function prototypes can be removed, along with `p_functions_hooks_array` in its entirety. Then, we can simplify registration a bit by using `register_kprobes()` and `register_kretprobes()` to register the entire kprobe and kretprobe arrays. What do you think @solardiz?

kerneltoast:
Also, here is the coccinelle version of `Hooking: Drop unused private data from kretprobes` I had made: https://github.com/kerneltoast/lkrg/commit/cfac75e75bb5a97ef117d087d2ec9bd096c9a663

It differs from your commit by only two newlines, which are removed by my commit:
```diff
+++ b/src/modules/exploit_detection/syscalls/exec/p_security_bprm_committed_creds/p_security_bprm_committed_creds.h
@@ -30,7 +30,6 @@
 #define P_MAX_PATH PATH_MAX + 0x20 /* For weirdos used by d_path */
 
 
-
 struct inode *p_get_inode_from_task(struct task_struct *p_arg);
 
 int p_security_bprm_committed_creds_ret(struct kretprobe_instance *p_ri, struct pt_regs *p_regs);
+++ b/src/modules/exploit_detection/syscalls/p_sel_write_enforce/p_sel_write_enforce.h
@@ -24,7 +24,6 @@
 #define P_LKRG_EXPLOIT_DETECTION_SEL_WRITE_ENFORCE_H
 
 
-
 int p_sel_write_enforce_ret(struct kretprobe_instance *p_ri, struct pt_regs *p_regs);
 int p_sel_write_enforce_entry(struct kretprobe_instance *p_ri, struct pt_regs *p_regs);
 int p_install_sel_write_enforce_hook(int p_isra);
 ```

We can go with whichever version you prefer.

**Comment by solardiz** [comment]
> All of the probe scaffolding can be simplified by creating dedicated sections to store each probe type.

I'm not sure what you mean by "sections" here, but I agree that further simplifications can and eventually should be made. I've just pushed my work-in-progress for you to see where I'm going for now. Maybe you'd want to implement your idea on top of these changes?

In fact, I think eventually all of those many header files could be removed. But then there could be occasional special cases where we'd need to re-introduce a few of those, if there's any unusual interaction needed with a certain hook.

**Comment by kerneltoast** [comment]
> > All of the probe scaffolding can be simplified by creating dedicated sections to store each probe type.
> 
> I'm not sure what you mean by "sections" here, [...]

I mean `__section()` (i.e. `__attribute__((__section__(foo)))`) to create an isolated section in the module binary to store probe structures. This is the same trick used by the kernel's init system, which is how e.g. `device_initcall(YOUR_INIT_FUNCTION);` enrolls a function into the init system. And the init system gets a pointer to the section using a linker script to define `__initcall_start` and `__initcall_end` variables indicating the beginning and end of the array of init function pointers (see the [INIT_CALLS linker script definition](https://github.com/torvalds/linux/blob/71b639af06ef1c422df87be36ad411bc31abf1ce/include/asm-generic/vmlinux.lds.h#L929-L941)).

> In fact, I think eventually all of those many header files could be removed. But then there could be occasional special cases where we'd need to re-introduce a few of those, if there's any unusual interaction needed with a certain hook.

All special attributes could be defined at the same location where the struct kprobe/kretprobe variable is defined.

**Comment by kerneltoast** [comment]
> > I mean `__section()` (i.e. `__attribute__((__section__(foo)))`) to create an isolated section in the module binary to store probe structures. [...] All special attributes could be defined at the same location where the struct kprobe/kretprobe variable is defined.
> 
> Oh, so you mean this as a way to group the probe structs together in the binary into an equivalent of an "array", while defining the "elements" where it's convenient for us? OK, maybe. But I'm also fine with defining an array manually (such as of pointers to such probe structs).

Defining an array manually means we'd need an `extern` declaration for every probe struct. Grouping all probe structs into a single binary section eliminates that need.

> What we need to get rid of in either case are the per-probe sets of install/uninstall functions.

Unless I'm missing something, this is naturally achieved by creating an array of pointers to every probe struct.

**Comment by solardiz** [comment]
> Defining an array manually means we'd need an `extern` declaration for every probe struct. Grouping all probe structs into a single binary section eliminates that need.

Yes.

> > What we need to get rid of in either case are the per-probe sets of install/uninstall functions.
> 
> Unless I'm missing something, this is naturally achieved by creating an array of pointers to every probe struct.

Right. There may be a few exceptions for probes installed at separate times, but for most of them this is the case.

**Comment by solardiz** [comment]
Our `do_exit` hook previously did a `p_ed_enforce_validation` on return. This didn't make sense since `do_exit` isn't supposed to return and since we'd have removed our shadow entry for the current task by that point. This is getting removed here along with the switch to simple kprobe.

**Comment by solardiz** [comment]
I'm now also testing these changes rebased on #389 (which GitHub will also do when we merge this - there are no conflicts).

**Comment by solardiz** [comment]
Performance testing of this PR's changes on top of #389 vs. clean 0.9.9 vs. no LKRG, on:

```
Intel(R) Xeon(R) CPU E3-1231 v3 @ 3.40GHz
(turbo to 3.8 GHz)
2x 8 GB DDR3-1600 ECC
```

otherwise idle, running Rocky Linux 9.5 (recently updated), kernel `5.14.0-503.40.1.el9_5.x86_64`. Using JtR bleeding-jumbo (source tree as of March 2025) build as a benchmark.

No LKRG:

```
Configure finished.  Now "make -s clean && make -sj8" to compile.

real    0m11.738s
user    0m7.900s
sys     0m3.970s

Make process completed.

real    0m45.180s
user    4m27.307s
sys     0m15.623s
```

LKRG 0.9.9

```
Configure finished.  Now "make -s clean && make -sj8" to compile.

real    0m13.059s
user    0m7.860s
sys     0m5.344s

Make process completed.

real    0m46.295s
user    4m28.145s
sys     0m22.324s
```

LKRG latest (#389 + this PR):

```
Configure finished.  Now "make -s clean && make -sj8" to compile.

real    0m12.761s
user    0m7.839s
sys     0m5.077s

Make process completed.

real    0m45.976s
user    4m27.584s
sys     0m21.002s
```

These results are reproducible (I ran each build several times, the differences between same-LKRG builds are smaller than what we see between two LKRG versions here).

So performance impact of LKRG reduces a little bit, but is still significant.

File: src/modules/exploit_detection/syscalls/p_install.h [hunk 1]
```diff
@@ -21,18 +21,24 @@
 #ifndef P_LKRG_EXPLOIT_DETECTION_INSTALL_H
 #define P_LKRG_EXPLOIT_DETECTION_INSTALL_H
 
+typedef struct {
```
kerneltoast, line 24: Typedef is against general kernel coding style. And on the rare occasions typedef _is_ used in the kernel, the type name is suffixed with _t (even though _t type suffixes are technically reserved by C).

kerneltoast:
> Our `do_exit` hook previously did a `p_ed_enforce_validation` on return. This didn't make sense since `do_exit` isn't supposed to return and since we'd have removed our shadow entry for the current task by that point. This is getting removed here along with the switch to simple kprobe.

I was aware of this too, and I think this deserves a call-out in its own commit.

src/modules/exploit_detection/syscalls/p_install.h
solardiz: I didn't add a `_t` precisely because I know it's reserved, but so are leading underscores, yet are used in the kernel.

LKRG already has stuff like:

```c
typedef struct p_lkrg_counter_lock {

   atomic_t p_counter;
   spinlock_t p_lock;

} p_lkrg_counter_lock;
```

Perhaps we'll need to switch them all to some consistent convention in a coding style update later.

solardiz: OK, I dropped the typedef, now we have `struct lkrg_probe`. This was rather time-consuming to change across the many commits moving these lines around, etc. I tried with `exec` between the rebase commands, but even then it was nasty.

solardiz:
> > Our `do_exit` hook previously did a `p_ed_enforce_validation` on return. This didn't make sense since `do_exit` isn't supposed to return and since we'd have removed our shadow entry for the current task by that point. This is getting removed here along with the switch to simple kprobe.
> 
> I was aware of this too, and I think this deserves a call-out in its own commit.

I've just added this into the commit message. I didn't make it a separate commit.

src/modules/exploit_detection/syscalls/p_install.h
kerneltoast: > I didn't add a _t precisely because I know it's reserved, but so are leading underscores, yet are used in the kernel.

Right, I was pointing out that the kernel diverges in this regard and does put _t, so the _t would've been needed to make it in line with the kernel's code conventions.

> Perhaps we'll need to switch them all to some consistent convention in a coding style update later.

Yes, and I think we should avoid perpetuating the old conventions in new code.

> This was rather time-consuming to change across the many commits moving these lines around, etc. I tried with exec between the rebase commands, but even then it was nasty.

Ouch, sorry to hear that. For the record, you could have done `git filter-repo --refs da4a7d982e0cd65b9692ae3576a461febe4529ee^..HEAD --replace-text <(echo 'lkrg_probe ==>struct lkrg_probe ')`

PR #394, opened 2025-05-25
Perform stricter cred pointer checks where no creds override is expected

solardiz:
### Description

Two changes/commits here:

1. Wrap integrity violation checks in unlikely() - this is just something I happened to start working on while implementing the more important change:
2. Compensate for the relaxed cred pointer checks in p_cmp_tasks() especially on 6.13+ by stricter checks in places where we also assert that our `off` flag must be off. This corresponds to a subset of our hooks where no creds override is expected. We may later expand these checks to more places. See #371

### How Has This Been Tested?
In the VM where/while I worked on completing these changes, and in our CI under my personal GitHub account.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,8 @@ The following major changes have been made since 0.9.9:
 
  *) Support Linux 6.13+ by not hooking {override,revert}_creds() anymore, and
     limiting detection of cred pointer overwrite attacks on those kernels
+ *) To compensate for the above and as an enhancement on older kernels, check
+    for cred pointer overwrites in certain other places where we did not before
  *) Switch many hooks from kretprobes to simple kprobes for greater reliability
     and improved performance
  *) Overhaul locking of per-task shadow data, using finer-grain locks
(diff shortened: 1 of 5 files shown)


solardiz:
Initial versions of these changes actually failed testing under our CI for [mkosi-boot (jammy)](https://github.com/solardiz/lkrg/actions/runs/15231216314/job/42839273296#logs) only. That's when I was checking both `cred` and `real_cred` against our shadow copies of those in `p_set_ed_process_off`. I've now relaxed to checking only `cred == real_cred` in there, but still checking against the shadow copies in `p_set_ed_process_on`. The main difference in our usage of these two functions is that in our hook of `security_bprm_committing_creds` we only use `p_set_ed_process_off` without a corresponding `on`, IIRC because we intend to reset the `off` flag for the newly started program.

So apparently on that version of kernel we happen to reach `security_bprm_committing_creds` with both `cred` and `real_cred` pointers already (equally) different from what we had saved (but the actual credentials that we track still the same). This may also be why we have the below weird logic in `p_cmp_tasks`:

```
    * Pre-check for *cred pointer corruption and report it if the credentials
    * also differ, but do not report those credentials differences just yet.
    * This may be weird, but it was preserved over code refactorings for now so
    * that no unintentional functional changes were made in those.  We may
    * reconsider it separately.
```

PR #395, opened 2025-05-26
CI: Fedora mainline build & boot

vt-alt:
### Description
Build (in `fedora:rawhide` docker image) for mainline kernels and quick boot using qemu-kvm over 9pfs mount and direct kernel & initrd binaries. Not the full system boot simulation like previous mkosi workflow used.

### How Has This Been Tested?
Tested in my GA fork.

This took 2 days.

### DIFF
--- a/.github/workflows/docker-boot.yml
+++ b/.github/workflows/docker-boot.yml
@@ -58,8 +58,6 @@ jobs:
             - name: Check that boot-tests script finished successfully
               run: grep 'run-boot-tests.sh - SUCCESS' boot.log
             - name: Check that boot.log contains shutdown sequence
-              run: |
-                  grep 'sysrq: Power Off' boot.log
-                  grep 'reboot: Power down' boot.log
+              run: "grep 'reboot: Power down' boot.log"
 
 # vim: sw=4
--- a/.github/workflows/fedora-mainline.yml
+++ b/.github/workflows/fedora-mainline.yml
@@ -1,0 +1,74 @@
+name: Fedora boot (mainline)
+on:
+    push:
+    pull_request:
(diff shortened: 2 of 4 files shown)


vt-alt:
Ok mkosi workflows failed because I changed poweroff sequence, I'll fix this. Perhaps, we should add `schedule.cron` entry like previous mainline workflow had. I thought about building only once per week, Torvalds releases around the midnight on every Sunday but who knows when Fedora updates the package.

solardiz:
@vt-alt Do you suggest we merge this now or are you planning any additional testing/changes first?

vt-alt:
I think this is mergeable. SIGBUS perhaps because the memory of a running process is deallocated and it crashes (because sysrq-triggered `o` is asynchronous by nature). The tests you propose will only clarify this. `powerof -f` is cleaner solution anyway.

vt-alt:
To the records I filled the related dracut bug https://github.com/dracutdevs/dracut/issues/2698 debugging this took almost 50% of the time and I didn't find the root cause.

vt-alt:
jfyi. Fedora mainline is strange, They have 6.15.0-0.rc7.58.fc43.x86_64 until May 29th, skipping 5.16 release (which was at May 25th), And next day, May 30th, they started to have 6.16.0-0.rc0.250528gfeacb1774bd5.5.fc43.x86_64 which is arbitrary commit (perhaps feacb1774bd5 from 2025-05-27) from merge window.

solardiz:
FWIW, these tests are failing now - looks like something changed in 6.16-rc compared to earlier kernels, and we need to adapt LKRG to that change.

Issue #400, opened 2025-06-17
Review/revise our actions on ftrace_modify_all_code


solardiz:
This is to continue that thread of the discussion started in #399.

solardiz:
I found that LKRG is very slow to unload when there are a lot of other kernel modules loaded - takes 5 minutes with 1275 modules loaded. It turns out that our hook of `ftrace_modify_all_code` is invoked 42 times on LKRG unload (perhaps once per removed k[ret]probe hook), detects a total of ~50k kernel + ~50k module changes, and seems to set all modules' bits in the bitmask - so all modules' hashes are recomputed 42 times (that's about 50k hash computations, then). The command code is 1, which is `FTRACE_UPDATE_CALLS`.

Simply swapping these lines:

```diff
-   p_exploit_detection_exit();
    p_unregister_arch_metadata();
+   p_exploit_detection_exit();
```

reduces the number of calls from 42 to 2. This may be a start towards making LKRG unload quicker, but perhaps there's more to the underlying issue. Maybe what we do on `ftrace_modify_all_code` should be revised.

In fact, not doing anything on this call doesn't have any ill effects in my quick tests so far (and makes LKRG unloading quick). So I need a testcase for the issue that us having this code is supposed to deal with.

solardiz:
Also this:

```diff
+++ b/src/modules/database/arch/p_arch_metadata.c
@@ -141,6 +141,12 @@ int p_unregister_arch_metadata(void) {
    p_uninstall_switch_idt_hook();
 #endif
 
+#if defined(CONFIG_DYNAMIC_FTRACE)
+   p_uninstall_ftrace_modify_all_code_hook();
+#endif
+#if defined(CONFIG_FUNCTION_TRACER)
+   p_uninstall_ftrace_enable_sysctl_hook();
+#endif
    /*
     * This is not an arch specific hook, but it's a good place to deregister it
     */
@@ -148,12 +154,6 @@ int p_unregister_arch_metadata(void) {
 #ifdef P_LKRG_CI_ARCH_JUMP_LABEL_TRANSFORM_APPLY_H
    p_uninstall_arch_jump_label_transform_apply_hook();
 #endif
-#if defined(CONFIG_DYNAMIC_FTRACE)
-   p_uninstall_ftrace_modify_all_code_hook();
-#endif
-#if defined(CONFIG_FUNCTION_TRACER)
-   p_uninstall_ftrace_enable_sysctl_hook();
-#endif
[...]

Issue #257, opened 2023-01-04
Repeated lkrg.trigger=1 makes system extremely unresponsive


solardiz:
On the same system as in #256, running as root one instance of:

```
while :; do sysctl lkrg.trigger=1; done
```

made the system extremely unresponsive, with many concurrent kernel threads, as seen in `top` output below. As I recall from past discussions with @Adam-pi3, this was supposed to result in just one instance of our integrity checking routine running at a time, so eating up just one logical CPU, unlike what we see here.

```
top - 18:38:07 up 46 min,  1 user,  load average: 455.96, 181.60, 67.30
Tasks: 751 total,   8 running, 743 sleeping,   0 stopped,   0 zombie
%Cpu(s):  0.0 us, 30.9 sy,  0.0 ni, 68.9 id,  0.0 wa,  0.1 hi,  0.0 si,  0.0 st
MiB Mem :  31990.0 total,  30378.6 free,   1227.3 used,   1003.1 buff/cache
MiB Swap:   1950.0 total,   1950.0 free,      0.0 used.  30762.7 avail Mem

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
  55465 root      20   0       0      0      0 D  10.1   0.0   0:02.44 kworker/u32:125+events_unbound
  55912 root      20   0       0      0      0 D   6.0   0.0   0:02.13 kworker/u32:322+events_unbound
  55206 root      20   0       0      0      0 D   5.8   0.0   0:01.76 kworker/u32:11+events_unbound
  55669 root      20   0       0      0      0 D   5.8   0.0   0:01.97 kworker/u32:215+events_unbound
  56055 root      20   0       0      0      0 D   5.5   0.0   0:01.65 kworker/u32:376+events_unbound
  55744 root      20   0       0      0      0 D   5.0   0.0   0:01.80 kworker/u32:247+events_unbound
  55691 root      20   0       0      0      0 D   4.6   0.0   0:02.28 kworker/u32:224+events_unbound
  55702 root      20   0       0      0      0 D   4.1   0.0   0:01.71 kworker/u32:486+events_unbound
     17 root      20   0       0      0      0 I   3.2   0.0   0:04.36 kworker/0:1-events
  55345 root      20   0       0      0      0 D   3.1   0.0   0:01.66 kworker/u32:74+events_unbound
    127 root      20   0       0      0      0 I   3.0   0.0   0:03.37 kworker/1:1-events
[...]

PR #401, opened 2025-06-18
Replace module bitmap with struct fields, and other related changes

solardiz:
### Description
This fixes a few issues in and simplifies the code handling JUMP_LABEL batch mode and ftrace updates.

### How Has This Been Tested?
In my development VM, and by our CI.


Adam-pi3:
@solardiz a few questions:
1. why replacing bitmask with `bool` per module?
2. why removing checks for the opcocdes?
3. Did you run stress tests on those changes? What is effective is to run parallel a few screens where:
   a. one screen executes: `while true; do sysctl lkrg.trigger=1; sleep 0; done`
   b. 4-5 other screens execute in the loop rmmod and modprobe parallel, for example 5 screens with: `while true; do rmmod joydev; rmmod serio_raw; modprobe serio_raw; rmmod pcspkr; modprobe joydev; modprobe pcspkr; done`
   c. Sometimes on a few screen you can do `while true; do echo 1 > /sys/kernel/debug/tracing/events/enable; echo 0 > /sys/kernel/debug/tracing/events/enable; done` if you want to check FTRACE functionality

solardiz:
> * why replacing bitmask with `bool` per module?

So that we don't need the separate memory allocation. The "Drop the former JUMP_LABEL and FTRACE module bitmap (no users left)" enabled by these changes removes 44 lines of code.

> * why removing checks for the opcocdes?

Because I saw no reason to have them. If you're aware of a reason, please let me know! I suspect the only reason we had them was to paper over the indexing bug (without realizing that).

> * Did you run stress tests on those changes?

I ran some, and some are running as I write this (insmod/rmmod of LKRG itself in a loop), but not exactly what you suggest - so I am going to try those as well. Thank you!

Edit: moved the rest of what I wrote in this comment into a comment in #400 instead, where it's more relevant.

solardiz:
> run stress tests on those changes? What is effective is to run parallel a few screens where:

I've just tried running a few instances of all of these commands concurrently. I started with a few of (b) and (c), and then added 2 or 3 of (a). When I added the last of (a) - don't recall if it was 2nd or 3rd - the system became extremely unresponsive. I think I reproduced #257:

```
$ w
 00:53:56 up 21:19, 14 users,  load average: 519.21, 344.32, 153.07
USER     TTY        LOGIN@   IDLE   JCPU   PCPU WHAT
root     pts/3     00:46    7:47   2:20   2:20  /bin/bash
root     pts/4     00:46    6:59   0.13s  0.00s rmmod serio_raw
root     pts/5     00:46    6:52   0.02s  0.02s modprobe serio_raw
root     pts/6     00:47    6:49   0.02s  0.02s rmmod joydev
root     pts/7     00:47    6:36   0.00s  0.00s rmmod pcspkr
root     pts/8     00:47    6:33   0.20s  0.00s modprobe pcspkr
root     pts/9     00:47    6:28   0.20s  0.00s modprobe joydev
root     pts/10    00:47    6:17  48.64s 48.64s /bin/bash
root     pts/11    00:47    6:16  55.80s 55.80s /bin/bash
root     pts/12    00:47    6:14   0.00s  0.00s /bin/bash
root     pts/13    00:47    6:01   0.00s  0.00s /bin/bash
root     pts/14    00:48    5:38   0.00s  0.00s /bin/bash
root     pts/15    00:48    5:24   1.76s  1.76s /bin/bash
root     pts/16    00:48    4:42   1.40s  1.40s /bin/bash
```

I copy-pasted `rmmod lkrg` to a root shell prompt I had open, which took a few minutes to actually be processed, but once it was things are back to normal. No other ill effects observed.

solardiz:
Also tested on CentOS 7, AlmaLinux 8.10, Rocky Linux 9.6.

PR #402, opened 2025-06-19
Never reset "off", check cred pointers even during seccomp sync

### LINKED ISSUE #388: p_reset_ed_flags() vs. SECCOMP_FILTER_FLAG_TSYNC
I'm afraid @kerneltoast's new SECCOMP_FILTER_FLAG_TSYNC handling code doesn't interact with `p_reset_ed_flags` correctly. It assumes that `off` for a non-current thread is only symmetrically incremented/decremented, but `p_reset_ed_flags` violates this assumption. We use `p_reset_ed_flags` in two places:

```
src/modules/exploit_detection/syscalls/p_wake_up_new_task/p_wake_up_new_task.c:               p_reset_ed_flags(p_tmp);
src/modules/exploit_detection/syscalls/exec/p_security_bprm_committed_creds/p_security_bprm_committed_creds.c:      p_reset_ed_flags(p_tmp);
```

Can these be reached for a thread also being manipulated by SECCOMP_FILTER_FLAG_TSYNC? I think so.

In `wake_up_new_task`, thread isn't woken up yet, but it's probably already visible to our handling of TSYNC? Should it be - that is, is it also visible to the kernel's handling of TSYNC?

The `p_reset_ed_flags` in `wake_up_new_task` is only reached when the task being woken up had previously been inserted into our shadow database. Why would this be? (In my recent testing, I observed that this is occasionally the case, but I don't know if this is normal and expected, or if it indicates a bug elsewhere.)

In `security_bprm_committed_creds`, I guess it's quite possible that a thread goes through `execve` while another thread does the TSYNC.

solardiz:
### Description
1. Resetting `off` was wrong during seccomp sync. Let's never do that. Fixes #388.
2. The `off` override during seccomp sync isn't the same as creds override, and no creds override is expected where we call the non-override form of `p_set_ed_process_on` / `p_set_ed_process_off`. So let's perform the creds pointer checks added in #394 even during seccomp sync.

### How Has This Been Tested?
The first change was tested for a while in my development VM. Both changes together are tested in our CI setup. I'll test these some more before merging.

### DIFF
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -1092,13 +1092,6 @@ notrace void p_set_ed_process_override_off(struct p_ed_process *p_source) {
    p_source->p_ed_task.p_off = p_off ^ p_global_off_cookie;
 }
 
-notrace void p_reset_ed_flags(struct p_ed_process *p_source) {
-
-   p_source->p_ed_task.p_off = p_global_cnt_cookie ^ p_global_off_cookie;
-   p_source->p_ed_task.p_off_count = 0;
-
-}
-
 static int p_insert_ed_task(struct p_ed_process *p_source) {
 
    int ret;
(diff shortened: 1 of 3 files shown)


kerneltoast (approved):
Looks good to me given your comments on #388.

solardiz:
Tested on a few systems for a day, on one more briefly (but including exploit detection). No issues. I'll merge.

PR #404, opened 2025-06-20
Fix a couple of CPU hotplug races against p_get_cpus()

kerneltoast:


src/modules/integrity_timer/p_integrity_timer.c
```diff
@@ -182,6 +165,12 @@ void p_check_integrity(struct work_struct *p_work) {
    read_lock(&p_config_lock);
    p_read_cpu_lock();
 
+   /* Find information about current CPUs in the system */
+   p_get_cpus(&p_tmp_cpu_info);
+   if (p_cmp_cpus(&p_db.p_cpu,&p_tmp_cpu_info)) {
+      p_print_log(P_LOG_ISSUE, "Using CPU number from original database");
```
solardiz, line 171: What does this message even mean - "Using CPU number from original database" - and is it still correct after these changes? Maybe it means that we were using our `p_nr_cpu_ids`, which the previous commit here removed? If so, this message should also be gone, right? And perhaps `p_cmp_cpus` should be changed to no longer return a status, which we'd have no use for anyway.

solardiz: Oh, the message could also refer to the online/possible/present/active CPU numbers, which we still check and print messages about in `p_cmp_cpus`. However, I don't see us actually use those for anything else. So I suggest we drop this message.

solardiz: FWIW, we had this message already in LKRG 0.0, as follows:

```c
   /* Find information about current CPUs in the system */
   p_get_cpus(&p_tmp_cpu_info);
   if (p_cmp_cpus(&p_db.p_cpu,&p_tmp_cpu_info)) {
      p_print_log(P_LKRG_WARN,
             "!!! INTEGRITY WARNING: Using CPU number from original database !!!\n");
   }
```

and it was followed with code not too different from what we have now. So I don't know what this message actually meant. Should we drop it, @Adam-pi3, or if not then what does it mean?

solardiz:
As part of testing this and related code in LKRG, I've just tried bringing some CPUs offline and back online via sysfs, and saw the CPU metadata hashes reported at log_level=4 change - without any alerts, meaning LKRG had updated the expected hashes.

src/modules/integrity_timer/p_integrity_timer.c
kerneltoast: I find the whole construction strange since it assumes an attacker may corrupt one of the global CPU maps, which is not practical in any exploitation scenario I can think of. There's also the CPU hotplug notifier, so I'm not sure why there's a need for extra checking on the online CPU mask.

solardiz: Maybe this wasn't meant to be directly about such attacks, but about LKRG somehow missing a hotplug event (some kind of incompatibility between LKRG and the kernel), so that we'd become aware of that and know that a hash mismatch is expected for that reason (is this what's going to happen?)

solardiz: In fact, we use `FAULT` inside `p_cmp_cpus` and `ISSUE` here, and not `ALERT`. Which means we're not claiming this is an attack - just that something went wrong.

Adam-pi3: The logic behind that was a bit stretch but let me shed some light. One of the main advantage of LKRG is that we are tracking metadata from each individual CPU/core where (at that time) any other EDR did not. As a malware writer you could change some of the critical MSR for not-main CPU/core and being invisible while LKRG would catch you.
As you notice, this variable (`nr_cpu_ids`) should never change. However, you can always dynamically put some of the CPU offline (so the kernel will never use it) and then you could modify this variable so the kernel will think there is less logical cores then actually exist. If you are malware, you don't need to rely on 'official' functionality to 'bring-back' this "lost" core into the pull of available one and you can implement that logic yourself (and never trigger notifier). Now this core might have been initialized with malicious MSRs. Because we were tracking `nr_cpu_ids` we would catch that scenario (but not after this PR).

solardiz: Adam, I almost started commenting on the specifics of this, but I feel it'd lead us to lengthy discussion of a topic that just isn't worth it for our project at this time.

Even with your comment, I still do not get what that message meant. It doesn't appear to match what LKRG ever did, but maybe I overlook something.

Anyway, with the current state of code, do you agree the `if` and the message should be removed, as it does not correspond with what LKRG actually does?

Adam-pi3: I looked at the history and how the message was changed and originally it was:

```
      p_print_log(P_LKRG_WARN,
             "!!! INTEGRITY WARNING: Using CPU number from original database !!!\n");
```

Then, first change was introduced here:
https://github.com/lkrg-org/lkrg/commit/24f7029ac39fa7b6278f12b7093d1259dfbd8a11#diff-04719e3825c189f146ceb5f17e69323a5f0c932b87c02584bd08c9a136a92782L157

which modified log level from WARN to ISSUE. Next change is here:
https://github.com/lkrg-org/lkrg/commit/e3badfaf738727854b419fd1a768f93b637a07ba#diff-04719e3825c189f146ceb5f17e69323a5f0c932b87c02584bd08c9a136a92782R156

both changes were part of the bigger set of changes unifying logging to bring remote logging capabilities.

We still gather `nr_cpu_ids` on database creation and I would keep this check.

solardiz: Reviewing the code we had prior to this PR, my current understanding is that we were using our copy of `nr_cpu_ids` only for memory allocations (including temporary in `p_check_integrity`) and for `memset` on enabling `msr_validate` via sysctl. Yet we'd run things `on_each_cpu` and use `smp_processor_id` in there, meaning we'd trust whatever the kernel thinks the current online CPU numbers are - both for what we validate and also even for not causing us to access out of bounds! Given this, the dropping of copying/validation of this variable in this PR is a minor change. And it was always misleading in that message (which we still have, and I think should drop) that we were "Using CPU number from original database". The user could think we base our validation on the original number, whereas we only base memory allocation and occasional `memset` on it (yet would happily access out of bounds if the kernel's number increases anyway).

So simply reverting the check removal commit would not bring us to a good state, but rather back to an inconsistent state. At least now we consistently trust the kernel's `nr_cpu_ids`, as opposed to having a weird mix of trust and distrust.

So once again, I'd like to proceed to clean this up further, dropping the "Using ..." message. In fact, we'd need to drop it even if we were to revert the check removal commit, because this message never reflected the reality.

solardiz: FWIW, I've just tried `nr_cpu_ids--;` on `lkrg.trigger` sysctl, and triggered it once, got this:

```
 [2025-07-01 00:18:00] [476674.902746] LKRG: FAULT: Number of possible CPUs changed unexpectedly (expected 4, actual 3)
 [2025-07-01 00:18:00] [476674.902766] LKRG: FAULT: Number of present CPUs changed unexpectedly (expected 4, actual 3)
 [2025-07-01 00:18:00] [476674.902787] LKRG: FAULT: Number of active CPUs changed unexpectedly (expected 4, actual 3)
 [2025-07-01 00:18:00] [476674.902800] LKRG: ISSUE: Using CPU number from original database
 [2025-07-01 00:18:00] [476674.902865] LKRG: ALERT: DETECT: CPU: Hash of CPU metadata has changed unexpectedly 
[2025-07-01 00:18:00] [476674.926727] LKRG: ALERT: DETECT: Kernel: 1 checksums changed unexpectedly
[2025-07-01 00:18:00] [476674.956712] BUG: kernel NULL pointer dereference, address: 0000000000000208 
```

and it crashed in `cpuidle_governor_latency_req`.

Since i had this on `trigger`, it means we validated kernel integrity immediately after the change, which is why those FAULTs and ALERTs above. The system crashed within a second, so if something like this happened from another context, we may not have had those messages - just the crash.

Anyway, it looks like:
1. If we do check in time (before system crash), then we do detect this sort of corruption through its various other effects.
2. The system crashes quickly... unless the corruption is also undone quickly maybe?

Adam-pi3: >FWIW, I've just tried nr_cpu_ids--;

can you also try `++`?

solardiz: OK, `++` is more fun. No visible effect the first few times, but when I ran `while :; do sysctl lkrg.trigger=1; done`, the system quickly crashed after maybe a few hundred increments. The logs show several userspace segfaults, some assorted BUGs, and several kernel space crashes, including one at a weird address (could maybe be mmap'able by exploit, if it had reason to expect control would be transferred there):

```
[2025-07-01 01:55:55] [ 5330.545262] RIP: 0033:0x76de56ebc08e
 [2025-07-01 01:55:55] [ 5330.545271] Code: Unable to access opcode bytes at 0x76de56ebc064.
```

No detection by LKRG, but plenty of other noise in the logs.

Is this worth checking that variable? What do you think @kerneltoast?

kerneltoast: > Is this worth checking that variable?

No.

There is an insurmountable synchronization hurdle with corrupting nr_cpu_ids on the fly. It is used in *many* places in the kernel due to the cpumask API, and since it isn't designed to be modified after smp init, no code in the kernel reads nr_cpu_ids using READ_ONCE(). Therefore, there are several potential data races on all nr_cpu_ids reads, where:
• the compiler may tear the nr_cpu_ids load into multiple smaller loads
• the compiler may redundantly reload nr_cpu_ids in a loop due to register allocation pressure
• weakly-ordered CPU architectures that lack multi-copy atomicity won't observe the updated nr_cpu_ids value (because every nr_cpu_ids load would need to be followed by an explicit read barrier instruction)

On top of that, since nr_cpu_ids isn't ever changed after kernel init, there aren't any ROP gadgets available to corrupt nr_cpu_ids.

I don't see any value in checking for nr_cpu_ids corruption.

solardiz: @kerneltoast While you're probably right that "There is an insurmountable synchronization hurdle with corrupting nr_cpu_ids on the fly", your 3 bullet points look weird to me in context of exploits, which often have far worse portability and reliability problems than these, yet are just successful enough to be a concern.

The point about lack of ROP gadgets makes more sense, but a write-what-where (or even a write-constant-where) primitive would be sufficient, if the right target address can be figured out and fed into it.

So I think the real question is whether changing the value of `nr_cpu_ids` is potentially useful for an exploit to succeed with a reasonable probability at whatever its reasonable goal is.

Another question is what we should do on such violations. So far, LKRG did very little - just log FAULTs.

Anyway, I do agree this discussion and experiments are mostly a waste of time. We have plenty of more important issues and concerns in and related to LKRG overall.

kerneltoast: > your 3 bullet points look weird to me in context of exploits, which often have far worse portability and reliability problems than these, yet are just successful enough to be a concern.

The implication of those bullet points is that they all easily lead to crashing the kernel. And since nr_cpu_ids is used in so many places in the kernel, it is difficult to make sure all CPU cores are executing code that won't cause a kernel crash when nr_cpu_ids changes.

> So I think the real question is whether changing the value of nr_cpu_ids is potentially useful for an exploit to succeed with a reasonable probability at whatever its reasonable goal is.

I think it isn't useful at all.

> Anyway, I do agree this discussion and experiments are mostly a waste of time. We have plenty of more important issues and concerns in and related to LKRG overall.

Agreed.

PR #407, opened 2025-08-06
pINT: No longer store/validate override creds

### LINKED ISSUE #374: A bogus cred structure pointer may be used (possible use-after-free)
When `p_update_ed_process()` is called for a task other than `current`, `p_update_ed_process()` may use and store a pointer to a bogus cred structure if the task in question is in the middle of an access/faccessat/faccessat2 syscall.

In the kernel, `do_faccessat()` may override the current task’s cred temporarily for the duration of an access/faccessat/faccessat2 syscall. A pointer to the temporary cred is stored in `struct task_struct::cred`, which can then be picked up by LKRG’s `p_update_ed_process()` function. This would not only result in LKRG reading and saving the wrong cred structure, but also result in a use-after-free possibility because the [temporary cred is freed without RCU protection](https://github.com/torvalds/linux/blob/adc218676eef25575469234709c2d87185ca223a/fs/open.c#L439-L457). Therefore, LKRG may dereference the bogus cred from `p_dump_creds()` after it is freed.

Here is the `do_faccessat()` [code in the kernel which may override the cred for a task.](https://github.com/torvalds/linux/blob/adc218676eef25575469234709c2d87185ca223a/fs/open.c#L411-L465)

This issue can be triggered either from seccomp with SECCOMP_FILTER_FLAG_TSYNC, or `p_iterate_lkrg_tasks_paranoid()`.

### LINKED ISSUE #406: Build failure on 6.17.0-0.rc0.250804gd2eedaa3909b.10.fc43.x86_64: undefined type 'struct module_use'
Our Fedora boot (mainline) CI job has just started failing today with:

```
    CC [M]  src/modules/self-defense/hiding/p_hiding.o
  In file included from /usr/src/kernels/6.17.0-0.rc0.250804gd2eedaa3909b.10.fc43.x86_64/include/linux/container_of.h:5,
                   from /usr/src/kernels/6.17.0-0.rc0.250804gd2eedaa3909b.10.fc43.x86_64/include/linux/kernel.h:22,
                   from src/modules/self-defense/hiding/../../../p_lkrg_main.h:24,
                   from src/modules/self-defense/hiding/p_hiding.c:18:
  src/modules/self-defense/hiding/p_hiding.c: In function 'p_unhide_itself':
  /usr/src/kernels/6.17.0-0.rc0.250804gd2eedaa3909b.10.fc43.x86_64/include/linux/container_of.h:21:54: error: invalid use of undefined type 'struct module_use'
     21 |         static_assert(__same_type(*(ptr), ((type *)0)->member) ||       \
        |                                                      ^~
```

and so on. https://github.com/solardiz/lkrg/actions/runs/16766497114/job/47472546373

Our affected code is:

```c
#define P_UNHIDE_FROM_KOBJ(p_mod,p_kset,p_ktype)                           \
do {                                                                       \
/* struct kobject *p_kobj; */                                              \
   struct module_use *p_use;                                               \
   int p_tmp;                                                              \
   p_debug_log(P_LOG_DEBUG, "Creating KOBJ for [%s]",                      \
                                              p_mod->name);                \
/* p_kobj = kset_find_obj(p_kset, p_mod->name);                            \
   if (p_kobj) {                                                           \
      p_debug_log(P_LOG_DEBUG, "Module [%s] is NOT hidden!",               \
                                              p_mod->name);                \
      kobject_put(p_kobj);                                                 \
      return;                                                              \
   } */                                                                    \
   p_mod->mkobj.mod = p_mod;                                               \
   memset(&p_mod->mkobj.kobj, 0, sizeof(p_mod->mkobj.kobj));               \
   p_mod->mkobj.kobj.kset = p_kset;                                        \
   if (kobject_init_and_add(&p_mod->mkobj.kobj, p_ktype, NULL,             \
                                              "%s", p_mod->name)) {        \
      p_debug_log(P_LOG_DEBUG, "FAILED :(");                               \
      goto p_unhide_itself_exit;                                           \
   }                                                                       \
   p_mod->holders_dir = kobject_create_and_add("holders",                  \
                                              &p_mod->mkobj.kobj);         \
   if (!p_mod->holders_dir) {                                              \
      p_debug_log(P_LOG_DEBUG, "FAILED :(");                               \
      goto p_unhide_itself_exit;                                           \
   }                                                                       \
   if ( (p_tmp = sysfs_create_files(&p_mod->mkobj.kobj,                    \
               (const struct attribute **)&p_mod->modinfo_attrs)) != 0) {  \
      p_debug_log(P_LOG_DEBUG, "FAILED :(");                               \
      goto p_unhide_itself_exit;                                           \
   }                                                                       \
   /* add_usage_links() */                                                 \
   list_for_each_entry(p_use, &p_mod->target_list, target_list) {          \
      p_tmp = sysfs_create_link(p_use->target->holders_dir,                \
                            &p_mod->mkobj.kobj, p_mod->name);              \
   }                                                                       \
```

and in particular `p_use->target->holders_dir`.

solardiz:
### Description
This fixes #374 and simplifies things at the same time. It also adds more Ubuntu CI jobs, and fixes #406.

### How Has This Been Tested?
In our CI setup and in my dev VM. I haven't yet tested that credentials overwrite detection still works after these changes - I will before merging this.

### DIFF
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -1332,18 +1332,39 @@ static int p_cmp_tasks(struct p_ed_process *p_orig, char p_kill) {
    get_cred(p_current_real_cred);
 
    /*
+    * Valid cases here:
+    * 1. No credentials change nor override, so both data and pointers match.
+    * 2. Untracked commit_creds(), but no change in tracked credentials, so
+    *    data matches, but pointers differ from saved, yet match each other.
+    * 3. Untracked override_creds() on 6.13+, so real_cred data and pointer
+    *    match what we saved, but cred differs from real_cred.
+    *
+    * Invalid cases here:
+    * 1. Untracked commit_creds() that changes our tracked credentials should
+    *    not exist - we're supposed to track all callers of such.
+    * 2. Untracked override_creds() on pre-6.13 should not exist - we track
+    *    override_creds() itself.
+    * 3. Tracked commit_creds() that's still in progress should set "off" for
+    *    the task, which we check for above, so would not reach here.
+    * 4. Tracked maybe-nested override_creds() on pre-6.13 should set "off" for
+    *    the task until a final/outer revert_creds(), so we would not reach
+    *    here with override still in effect.
+    */
+
+   /*
     * Pre-check for *cred pointer corruption and report it if the credentials
     * also differ, but do not report those credentials differences just yet.
-    * This may be weird, but it was preserved over code refactorings for now so
-    * that no unintentional functional changes were made in those.  We may
-    * reconsider it separately.
+    * This is needed because we do not hook some callers of commit_creds(),
+    * such as by LSMs, where only creds that we currently do not track may
+    * change (so we'd only detect the pointer change, which would be a false
+    * positive).
     *
     * We no longer hook override_creds() on Linux 6.13+, so we infer that an
-    * override is in effect through only the cred pointer being different, but
-    * real_cred staying unchanged.  If both have changed, commit_creds() may
-    * have been used.  Here we assume that uses of override_creds*() are
-    * legitimate, but uses of commit_creds() are not (we set our "off" flag on
-    * the legitimate ones, so would not reach this function in those cases).
+    * override is in effect through only the cred pointer being different,
+    * but real_cred staying unchanged.  If both have changed, commit_creds()
+    * may have been used.  Here we assume that uses of override_creds*() are
+    * legitimate on 6.13+, but uses of commit_creds() are not (we would not
+    * reach these final checks on legitimate uses).
     */
    if (unlikely(p_orig->p_ed_task.p_cred_ptr != p_current_cred)
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(6,13,0)
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -1371,6 +1371,16 @@ static int p_cmp_tasks(struct p_ed_process *p_orig, char p_kill) {
 #endif
          P_CMP_PTR(p_orig->p_ed_task.p_real_cred_ptr, p_current_cred, "cred")
    }
+#if LINUX_VERSION_CODE < KERNEL_VERSION(6,13,0)
+   else if (unlikely(p_current_real_cred != p_current_cred)) {
+      /*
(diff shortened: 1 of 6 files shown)


solardiz:
> I haven't yet tested that credentials overwrite detection still works after these changes

I confirmed this now.

solardiz:
@kerneltoast @Adam-pi3 Any review from you please?

src/modules/exploit_detection/p_exploit_detection.c
```diff
@@ -1448,14 +1429,12 @@ static inline void p_cmp_cred_ptr(struct p_ed_process *edp, bool full_check) {
...
    int p_ret = 0;
 
    if (full_check) {
-      P_CMP_PTR(edp->p_ed_task.p_cred_ptr, p_current_cred, "cred")
       P_CMP_PTR(edp->p_ed_task.p_real_cred_ptr, p_current_real_cred, "real_cred")
-   }
-   if (likely(!p_ret))
+      P_CMP_PTR(edp->p_ed_task.p_real_cred_ptr, p_current_cred, "cred")
+   } else
       P_CMP_PTR(p_current_real_cred, p_current_cred, "cred")
```
solardiz: I wonder if in this real_cred vs. cred comparison we should report "real_cred or cred" since we don't really know which one of these is corrupted. (Not new with this PR.)

solardiz:
I'm now adding this comment:

```
    * Valid cases here:
    * 1. No credentials change nor override, so both data and pointers match.
    * 2. Untracked commit_creds(), but no change in tracked credentials, so
    *    data matches, but pointers differ from saved, yet match each other.
    * 3. Untracked override_creds() on 6.13+, so real_cred data and pointer
    *    match what we saved, but cred differs from real_cred.
    *
    * Invalid cases here:
    * 1. Untracked commit_creds() that changes our tracked credentials should
    *    not exist - we're supposed to track all callers of such.
    * 2. Untracked override_creds() on pre-6.13 should not exist - ditto.
    * 3. Tracked commit_creds() that's still in progress should set "off" for
    *    the task, which we check for above, so would not reach here.
    * 4. Tracked maybe-nested override_creds() on pre-6.13 should set "off" for
    *    the task until a final/outer revert_creds(), so we would not reach
    *    here with override still in effect.
```

I'm also adding an extra `cred == real_cred` check on pre-6.13. This extra check may need some (re)testing (I had already done lots of testing of this PR on many different systems/kernels, but without this extra check).

@kerneltoast @Adam-pi3 If you'd like to review, please do it now. I may as well merge without review very soon as I feel my understanding of the underlying issues and constraints is sufficient now.

Issue #410, opened 2025-08-12
CI: ALT Sisyphus failing: E: Package kernel-headers-modules-mainline is a virtual package with multiple good providers


solardiz:
https://github.com/lkrg-org/lkrg/actions/runs/16917910308/job/47936386158 @vt-alt 

```
Package kernel-headers-modules-mainline is a virtual package provided by:
  kernel-headers-modules-6.17 sisyphus+392151.100.1.1@1754856761
  kernel-headers-modules-6.16 sisyphus+391001.100.1.1@1753662092
You should explicitly select one to install.
E: Package kernel-headers-modules-mainline is a virtual package with multiple good providers.
```

vt-alt:
Yea, it's because they don't release 6.16.1 more than 2 weeks and now there are two mainline kernels. I wasn't expecting that turn of events.

vt-alt:
Maybe we could fix CI instead of these unreliable apt-get install targets...

PR #411, opened 2025-08-13
CI: Replace unreliable kernel package targets for alt:sisyphus

vt-alt:
Use supposedly more robust install method.

Fixes: https://github.com/lkrg-org/lkrg/issues/410

### How Has This Been Tested?
GA.


solardiz:
Thank you for fixing this so promptly @vt-alt!

Issue #259, opened 2023-02-02
LKRG Trips kCFI During get_kallsyms_address (on-load)


sempervictus:
Finally managed to get a 6.1.8 built with kCFI and LTO on LLVM16, LKRG still as a module. Unfortunately, the modules attempt to look-up addresses (it was built in-tree - in the hopes of avoiding this exact type of nonsense) at load-time produces this effect (hey, CFI works):
```sh
# modprobe lkrg 
[ 1285.469868][  T440] CFI failure at get_kallsyms_address+0xe6/0x150 [lkrg] (target: kallsyms_lookup_name+0x4/0x210; expected type: 0xaa7e236f)
[ 1285.471017][  T440] invalid opcode: 0000 [#1] PREEMPT SMP
[ 1285.471475][  T440] CPU: 3 PID: 440 Comm: modprobe Tainted: G                TN 6.1.8 #1
[ 1285.472168][  T440] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS Arch Linux 1.16.1-1-1 04/01/2014
[ 1285.472984][  T440] RIP: 0010:get_kallsyms_address+0xe6/0x150 [lkrg]
[ 1285.473570][  T440] Code: 28 48 89 05 f4 7c 02 00 48 89 e7 e8 44 e4 81 ec 4c 8b 1d e5 7c 02 00 48 c7 c7 41 06 9a c0 41 ba 91 dc 81 55 45 03 53 fc 74 02 <0f> 0b 41 ff d3 66 90 48 89 05 74 7d 02 00 31 db eb 10 48 c7 c3 ff
[ 1285.475287][  T440] RSP: 0018:ffffb18881bbf998 EFLAGS: 00010202
[ 1285.475863][  T440] RAX: 0000000000000000 RBX: ffffffffc0911005 RCX: 0000000000000000
[ 1285.476512][  T440] RDX: 0000000000384a83 RSI: 0000000000000000 RDI: ffffffffc09a0641
[ 1285.477155][  T440] RBP: ffffb18881bbfd70 R08: ffff97d9847ee0c0 R09: ffff97d980042800
[ 1285.477834][  T440] R10: 0000000055a0ebf7 R11: ffffffffad176a94 R12: ffffffffaea19728
[ 1285.478500][  T440] R13: 0000000000000000 R14: ffff97d9848e02a0 R15: 0000000000000000
[...]

PR #415, opened 2025-08-21
Intel CET IBT support

### LINKED ISSUE #413: LKRG lacks/needs Intel CET IBT support (BUG/Oops/panic on loading LKRG into Arch's 6.15.9-hardened1-2-hardened x86_64: Missing ENDBR: kallsyms_lookup_name+0x4/0xc0)
Hello!
I followed the instructions to install lkrg from this github repo however when I came to the testing part I ran ```sudo insmod output/lkrg.ko kint_enforce=1``` and it triggered the kernel panic mode. I then tried to install it with ```sudo make install``` and loading it via systemd as if I were to deploy it via ```sudo systemctl start lkrg``` however that triggered the kernel panic mode too. Afterwards I tried checking the logs via ```sudo journalctl -k -b -1``` but nothing related showed up in it. I've seen that there is a logging documentation however that is for remote logging as far as I know. How can I check/make logs on the host? A couple details about my setup if it's needed: arch linux, hardened kernel, pci passthru setup with vfio on one of the gpus, encrypted root partition and apparmor is installed and loaded.
Thanks for the help in advance!

solardiz:
Fixes #413

### Description
Initial hack to workaround Intel CET IBT.

### How Has This Been Tested?
In our CI and in an Arch Linux VM, and also by forcing this code to be compiled in and/or running on other x86_64 systems (even without IBT support in the kernel). We do not currently know whether it actually solves the problem, but it should.

After proper testing, next steps may be making it more reliable (not depending on compiler specifics so much, as mentioned in the lengthy source code comment) and/or implementing a similar trick for ARM BTI.

### DIFF
--- a/src/p_lkrg_main.h
+++ b/src/p_lkrg_main.h
@@ -280,8 +280,8 @@ typedef struct _p_lkrg_global_symbols_structure {
 } p_lkrg_global_syms;
 
 #ifdef P_LKRG_UNEXPORTED_MODULE_ADDRESS
-#define LKRG_P_MODULE_ADDRESS(p_addr)      P_SYM(p___module_address)(p_addr)
-#define LKRG_P_MODULE_TEXT_ADDRESS(p_addr) P_SYM(p___module_text_address)(p_addr)
+#define LKRG_P_MODULE_ADDRESS(p_addr)      P_SYM_CALL(p___module_address, p_addr)
+#define LKRG_P_MODULE_TEXT_ADDRESS(p_addr) P_SYM_CALL(p___module_text_address, p_addr)
 #else
 #define LKRG_P_MODULE_ADDRESS(p_addr)      __module_address(p_addr)
 #define LKRG_P_MODULE_TEXT_ADDRESS(p_addr) __module_text_address(p_addr)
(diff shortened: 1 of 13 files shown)


solardiz:
@user889252627 @sempervictus and anyone else running recent Linux kernel on bare metal recent Intel CPU, please help test this. Thank you!

solardiz:
I've also just tested builds with `P_KERNEL_AGGRESSIVE_INLINING`, separately with and without the new IBT support code, including `insmod` (on a system without IBT). No issues. (This got me thinking that maybe we should drop `P_KERNEL_AGGRESSIVE_INLINING` and just use one set of functions or the other all the time. But for now it stays.)

```diff
+++ b/src/p_lkrg_main.h
@@ -133,7 +133,7 @@ static inline unsigned long get_random_long(void) {
  * by uncommenting following definition. However, not all of them
  * so you need to experiment.
  */
-//#define P_KERNEL_AGGRESSIVE_INLINING 1
+#define P_KERNEL_AGGRESSIVE_INLINING 1
 
 //#define p_lkrg_read_only __attribute__((__section__(".data..p_lkrg_read_only"),aligned(PAGE_SIZE)))
 #define __p_lkrg_read_only __attribute__((__section__(".p_lkrg_read_only")))
@@ -330,7 +330,7 @@ extern p_ro_page p_ro;
       goto p_sym_error; \
    }
 
-#ifdef CONFIG_X86_KERNEL_IBT
+#if 1 // def CONFIG_X86_KERNEL_IBT
 
 /*
  * When Intel CET IBT is in use, we can't call non-exported kernel functions
```

src/p_lkrg_main.h
```diff
@@ -330,6 +330,103 @@ extern p_ro_page p_ro;
...
+ *
+ * __attribute__ ((noipa)) isn't recognized by older compilers, which is a
+ * problem if the kernel is nevertheless recent enough to be built with IBT
+ * support.  In testing, noinline along with noclone wasn't good enough.  We
+ * could improve upon this by having these functions in their own translation
+ * unit, ideally coming from an assembly rather than C source file.
+ */
+#define GENERATE_CALL_FUNC(type, name, ...) \
+   notrace __attribute__ ((noipa)) type call_##name(__VA_ARGS__) { \
+      __asm__ ("pushq %0" :: "m" (P_SYM(name))); \
```
solardiz: `pushq` implies 64-bit, but IBT is also supported in 32-bit mode. I don't know if Linux supports it in a 32-bit build or not.

solardiz:
I've just merged this based on it currently having no known issues and one report of it actually solving the problem. This way, we'll start receiving testing by more people sooner, and will have greater confidence (or on the contrary, some problem reports) before our next release.

sempervictus:
> @user889252627 @sempervictus and anyone else running recent Linux kernel on bare metal recent Intel CPU, please help test this. Thank you!

Thanks for the tag, sorry been a bit underwater lately. Will add to our upstream builds though only non-kcfi target we still have are nv/mellanox systems

solardiz:
Thank you @sempervictus. Why do you mention "non-kcfi"? With the changes here, we also want testing _with_ kCFI. Oh, but I see we'll need to revise our compile-time check, perhaps:

```diff
-#ifdef CONFIG_X86_KERNEL_IBT
+#if defined(CONFIG_X86_KERNEL_IBT) || defined(CONFIG_CFI_CLANG)
```

Does this look right?

sempervictus:
The issue w any CFI is it doesn't like dynamic probes, rap sorta makes this moot which is most of our hosts but for client systems we usually build something like the GOS kernel for x86. Nv and mlx drivers don't play ball. There are some crafty patches in the works for the nv driver to exclude CFI but at least ofed hardcodes GCC and the move to DOCA is slow...

solardiz:
> The issue w any CFI is it doesn't like dynamic probes

Yet k(ret)probes work just fine with Intel CET IBT. Only our own indirect calls into the kernel were impacted. Your report in #259 shows the same, and KCTF is said to be forward edge only just like IBT. So I think our changes here are quite likely to help, or at least the failure mode would be different (e.g., clang may miscompile our low-level hack here, which we've only tested with gcc). I'd like to know what happens if you enable these same changes along with KCFI.

PR #416, opened 2025-08-22
pINT: Allow SECCOMP_MODE_DEAD

### LINKED ISSUE #408: ALERT: DETECT: Task: seccomp mode corruption (expected 2 vs. actual 3) for pid 3756586, name sshd-auth
I had a shell loop with insmod/rmmod of LKRG running in an Arch Linux VM for a couple of months. Checking the multi-gigabyte log (obtained via our remote logging, starts on 2025-06-13T05:46:48Z after I rotated the previous one) for anything besides the expected messages, I just found two instances of `seccomp mode corruption (expected 2 vs. actual 3) for pid 3756586, name sshd-auth` (only the PID varies), somehow on two adjacent days in June (maybe there was unusual load on the underlying system at this time, or something).

```
2025-06-26T03:11:46Z - Freezing user space processes completed (elapsed 0.001 seconds)
2025-06-26T03:11:46Z - OOM killer disabled.
2025-06-26T03:11:47Z - OOM killer enabled.
2025-06-26T03:11:47Z - Restarting tasks ... done.
2025-06-26T03:11:47Z - LKRG: DYING: LKRG unloaded
2025-06-26T03:11:47Z - printk: legacy console [lkrg0] disabled
2025-06-26T03:11:49Z - printk: legacy console [lkrg0] enabled
2025-06-26T03:11:49Z - LKRG: ALIVE: Loading LKRG
2025-06-26T03:11:49Z - Freezing user space processes
2025-06-26T03:11:49Z - Freezing user space processes completed (elapsed 0.001 seconds)
2025-06-26T03:11:49Z - OOM killer disabled.
2025-06-26T03:11:49Z - LKRG: ALIVE: LKRG initialized successfully
2025-06-26T03:11:49Z - OOM killer enabled.
2025-06-26T03:11:49Z - Restarting tasks ... done.
2025-06-26T03:11:49Z - audit: type=1326 audit(1750907509.644:3): auid=4294967295 uid=65534 gid=65534 ses=4294967295 pid=3756586 comm="sshd-auth" exe="/usr/lib/ssh/sshd-auth" sig=31 arch=c000003e syscall=219 compat=0 ip=0x70ad0249fe56 code=0x0
2025-06-26T03:11:49Z - LKRG: ALERT: DETECT: Task: seccomp mode corruption (expected 2 vs. actual 3) for pid 3756586, name sshd-auth
2025-06-26T03:11:49Z - LKRG: ALERT: BLOCK: Task: Killing pid 3756586, name sshd-auth
2025-06-26T03:11:49Z - LKRG: ALERT: DETECT: Task: seccomp mode corruption (expected 2 vs. actual 3) for pid 3756586, name sshd-auth
2025-06-26T03:11:49Z - LKRG: ALERT: BLOCK: Task: Killing pid 3756586, name sshd-auth
2025-06-26T03:11:49Z - LKRG: ALERT: BLOCK: Task: Killing pid 3756586, name sshd-auth
2025-06-26T03:11:49Z - systemd-coredump[3756592]: Failed to initialize pidref from pid 3756586: No such process
2025-06-26T03:11:50Z - LKRG: DYING: Unloading LKRG
2025-06-26T03:11:50Z - Freezing user space processes
2025-06-26T03:11:50Z - Freezing user space processes completed (elapsed 0.001 seconds)
2025-06-26T03:11:50Z - OOM killer disabled.
2025-06-26T03:11:51Z - OOM killer enabled.
2025-06-26T03:11:51Z - Restarting tasks ... done.
2025-06-26T03:11:51Z - LKRG: DYING: LKRG unloaded
2025-06-26T03:11:51Z - printk: legacy console [lkrg0] disabled
2025-06-26T03:11:53Z - printk: legacy console [lkrg0] enabled
2025-06-26T03:11:53Z - LKRG: ALIVE: Loading LKRG
2025-06-26T03:11:53Z - Freezing user space processes
2025-06-26T03:11:53Z - Freezing user space processes completed (elapsed 0.001 seconds)
2025-06-26T03:11:53Z - OOM killer disabled.
2025-06-26T03:11:53Z - LKRG: ALIVE: LKRG initialized successfully
```

```
2025-06-27T20:37:10Z - Freezing user space processes
2025-06-27T20:37:10Z - Freezing user space processes completed (elapsed 0.001 seconds)
2025-06-27T20:37:10Z - OOM killer disabled.
2025-06-27T20:37:10Z - OOM killer enabled.
2025-06-27T20:37:10Z - Restarting tasks ... done.
2025-06-27T20:37:10Z - LKRG: DYING: LKRG unloaded
2025-06-27T20:37:10Z - printk: legacy console [lkrg0] disabled
2025-06-27T20:37:13Z - printk: legacy console [lkrg0] enabled
2025-06-27T20:37:13Z - LKRG: ALIVE: Loading LKRG
2025-06-27T20:37:13Z - Freezing user space processes
2025-06-27T20:37:13Z - Freezing user space processes completed (elapsed 0.001 seconds)
2025-06-27T20:37:13Z - OOM killer disabled.
2025-06-27T20:37:13Z - LKRG: ALIVE: LKRG initialized successfully
2025-06-27T20:37:13Z - OOM killer enabled.
2025-06-27T20:37:13Z - Restarting tasks ... 
2025-06-27T20:37:13Z - LKRG: ALERT: DETECT: Task: seccomp mode corruption (expected 2 vs. actual 3) for pid 4090469, name sshd-auth
2025-06-27T20:37:13Z - LKRG: ALERT: BLOCK: Task: Killing pid 4090469, name sshd-auth
2025-06-27T20:37:13Z - coredump: 4090469(sshd-auth): |/usr/lib/systemd/systemd-coredump pipe failed
2025-06-27T20:37:13Z c done.
2025-06-27T20:37:13Z - audit: type=1326 audit(1751056633.368:4): auid=4294967295 uid=65534 gid=65534 ses=4294967295 pid=4090469 comm="sshd-auth" exe="/usr/lib/ssh/sshd-auth" sig=31 arch=c000003e syscall=219 compat=0 ip=0x7f552e69fe56 code=0x0
2025-06-27T20:37:13Z - LKRG: DYING: Unloading LKRG
2025-06-27T20:37:13Z - Freezing user space processes
2025-06-27T20:37:13Z - Freezing user space processes completed (elapsed 0.000 seconds)
2025-06-27T20:37:13Z - OOM killer disabled.
2025-06-27T20:37:14Z - OOM killer enabled.
2025-06-27T20:37:14Z - Restarting tasks ... done.
2025-06-27T20:37:14Z - LKRG: DYING: LKRG unloaded
2025-06-27T20:37:14Z - printk: legacy console [lkrg0] disabled
2025-06-27T20:37:16Z - printk: legacy console [lkrg0] enabled
2025-06-27T20:37:16Z - LKRG: ALIVE: Loading LKRG
2025-06-27T20:37:16Z - Freezing user space processes
2025-06-27T20:37:16Z - Freezing user space processes completed (elapsed 0.001 seconds)
```

solardiz:
Fixes #408

### Description

This allows a task's current seccomp mode to be `SECCOMP_MODE_DEAD`, which happens when the task is exiting.

It weakens our detection slightly as an exploit could set this mode and go undetected by LKRG, but doing so would instead trigger the kernel itself to detect this soon and do:

```c
        /* Surviving SECCOMP_RET_KILL_* must be proactively impossible. */
        case SECCOMP_MODE_DEAD:
                WARN_ON_ONCE(1);
                do_exit(SIGKILL);
                return -1;
```

So an exploit doing this doesn't appear to achieve anything - not a bypass, but task killed.

### How Has This Been Tested?

On my dev VM and in our CI setup. Reproducing the original issue took a couple of months, so I'm not waiting for that, and we do not have a test case developed specifically to trigger this quickly.

### DIFF
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -1423,7 +1423,8 @@ static int p_cmp_tasks(struct p_ed_process *p_orig, char p_kill) {
          p_ret++;
       }
 
-      if (unlikely(p_orig->p_ed_task.p_sec.sec.mode != p_current->seccomp.mode)) {
+      if (unlikely(p_orig->p_ed_task.p_sec.sec.mode != p_current->seccomp.mode) &&
+          likely(p_current->seccomp.mode != 3 /* SECCOMP_MODE_DEAD */)) {
          if (p_current->seccomp.mode < 0 || p_current->seccomp.mode > 2
              || p_orig->p_ed_task.p_sec.sec.mode < 0 || p_orig->p_ed_task.p_sec.sec.mode > 2) {
             p_print_log(P_LOG_ALERT, "DETECT: Task: seccomp mode corruption (expected %u vs. actual %u) for pid %u, name %s",
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -1288,6 +1288,7 @@ static int p_cmp_tasks(struct p_ed_process *p_orig, char p_kill) {
    register long p_off = p_orig->p_ed_task.p_off ^ p_global_off_cookie;
    const struct cred *p_current_cred = NULL;
    const struct cred *p_current_real_cred = NULL;
+   int p_seccomp_mode;
 
    if (p_off - p_global_cnt_cookie) {
       if (p_kill)
@@ -1423,16 +1424,22 @@ static int p_cmp_tasks(struct p_ed_process *p_orig, char p_kill) {
          p_ret++;
       }
 
-      if (unlikely(p_orig->p_ed_task.p_sec.sec.mode != p_current->seccomp.mode) &&
-          likely(p_current->seccomp.mode != 3 /* SECCOMP_MODE_DEAD */)) {
-         if (p_current->seccomp.mode < 0 || p_current->seccomp.mode > 2
+      /*
+       * Use READ_ONCE because SECCOMP_MODE_DEAD may be set async to our checks
+       * when we are validating other than the current task, yet we need our
+       * checks to be internally consistent.
+       */
+      p_seccomp_mode = READ_ONCE(p_current->seccomp.mode);
+      if (unlikely(p_orig->p_ed_task.p_sec.sec.mode != p_seccomp_mode) &&
+          p_seccomp_mode != 3 /* SECCOMP_MODE_DEAD */) {
+         if (p_seccomp_mode < 0 || p_seccomp_mode > 2
              || p_orig->p_ed_task.p_sec.sec.mode < 0 || p_orig->p_ed_task.p_sec.sec.mode > 2) {
             p_print_log(P_LOG_ALERT, "DETECT: Task: seccomp mode corruption (expected %u vs. actual %u) for pid %u, name %s",
-               p_orig->p_ed_task.p_sec.sec.mode, p_current->seccomp.mode, task_pid_nr(p_current), p_current->comm);
+               p_orig->p_ed_task.p_sec.sec.mode, p_seccomp_mode, task_pid_nr(p_current), p_current->comm);
          } else {
             static const char * const p_sec_strings[] = {"SECCOMP_MODE_DISABLED", "SECCOMP_MODE_STRICT", "SECCOMP_MODE_FILTER"};
             p_print_log(P_LOG_ALERT, "DETECT: Task: seccomp mode corruption (expected %s vs. actual %s) for pid %u, name %s",
-               p_sec_strings[p_orig->p_ed_task.p_sec.sec.mode], p_sec_strings[p_current->seccomp.mode],
+               p_sec_strings[p_orig->p_ed_task.p_sec.sec.mode], p_sec_strings[p_seccomp_mode],
                task_pid_nr(p_current), p_current->comm);
          }
          p_ret++;
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -1427,11 +1427,20 @@ static int p_cmp_tasks(struct p_ed_process *p_orig, char p_kill) {
       /*
        * Use READ_ONCE because SECCOMP_MODE_DEAD may be set async to our checks
        * when we are validating other than the current task, yet we need our
-       * checks to be internally consistent.
+       * checks to be internally consistent.  Limit SECCOMP_MODE_DEAD support
+       * to kernels that actually have it, so that this magic value cannot be
+       * used to silently bypass our validation on older kernels.  The kernels
+       * that support it will kill the SECCOMP_MODE_DEAD task on their own.
        */
       p_seccomp_mode = READ_ONCE(p_current->seccomp.mode);
-      if (unlikely(p_orig->p_ed_task.p_sec.sec.mode != p_seccomp_mode) &&
-          p_seccomp_mode != 3 /* SECCOMP_MODE_DEAD */) {
+      if (unlikely(p_orig->p_ed_task.p_sec.sec.mode != p_seccomp_mode)
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,16,10) || \
+    (LINUX_VERSION_CODE < KERNEL_VERSION(5,16,0) && LINUX_VERSION_CODE >= KERNEL_VERSION(5,15,24)) || \
+    (LINUX_VERSION_CODE < KERNEL_VERSION(5,11,0) && LINUX_VERSION_CODE >= KERNEL_VERSION(5,10,211)) || \
+    (LINUX_VERSION_CODE < KERNEL_VERSION(5,5,0) && LINUX_VERSION_CODE >= KERNEL_VERSION(5,4,180))
+          && p_seccomp_mode != 3 /* SECCOMP_MODE_DEAD on 5.17+ and backports */
+#endif
+          ) {
          if (p_seccomp_mode < 0 || p_seccomp_mode > 2
              || p_orig->p_ed_task.p_sec.sec.mode < 0 || p_orig->p_ed_task.p_sec.sec.mode > 2) {
             p_print_log(P_LOG_ALERT, "DETECT: Task: seccomp mode corruption (expected %u vs. actual %u) for pid %u, name %s",


solardiz:
I wonder if we need a `READ_ONCE` on `p_current->seccomp.mode` before all the checks when validating a non-current task (even though there's no `WRITE_ONCE` on the other end), or to limit this to the current task like we do in nearby check of the seccomp filter anyway. In the latter check, we also check `PF_EXITING`, which we probably could here too _if_ we limit to the current task (as I think `SECCOMP_MODE_DEAD` is set just _before_ `PF_EXITING`, but for the current task I think we'll only reach this code after both are set).

solardiz:
Added `READ_ONCE`.

kerneltoast:
> I wonder if we need a `READ_ONCE` on `p_current->seccomp.mode` before all the checks when validating a non-current task (even though there's no `WRITE_ONCE` on the other end), or to limit this to the current task like we do in nearby check of the seccomp filter anyway. In the latter check, we also check `PF_EXITING`, which we probably could here too _if_ we limit to the current task (as I think `SECCOMP_MODE_DEAD` is set just _before_ `PF_EXITING`, but for the current task I think we'll only reach this code after both are set).

No, `READ_ONCE` isn't needed for _this_ reason because—disregarding `SECCOMP_MODE_DEAD`—`p_current->seccomp.mode` is only changed from inside `do_seccomp()` and therefore our seccomp handling will ensure the off flag is set for any task that may have its seccomp mode changed.

For `SECCOMP_MODE_DEAD` alone this doesn't hold. But a transition to `SECCOMP_MODE_DEAD` is always followed by a fatal signal, so we _shouldn't_ need to accommodate `SECCOMP_MODE_DEAD` for all cases and instead only let it slide in `do_exit()`. _Shouldn't_, at least...

...but paranoid mode could observe `SECCOMP_MODE_DEAD` before the task hits `do_exit()`. :cry:

And if the `queue_work()` hook is permitted from interrupt contexts again, it could also run into the same issue.

So a `READ_ONCE` is mandatory at the very least to ensure a `p_current->seccomp.mode` transition to 3 (`SECCOMP_MODE_DEAD`) after the final `p_current->seccomp.mode` check won't trigger an out of bounds access into the `p_sec_strings` array.

Furthermore, I think permitting `SECCOMP_MODE_DEAD` should be limited only to kernels >=5.17 which is when `SECCOMP_MODE_DEAD` was introduced. Otherwise, in a convoluted attack scheme, `p_current->seccomp.mode` could be changed to 3 on a pre-5.17 kernel and get past the `seccomp_may_assign_mode()` check to override the existing filter, which LKRG may have been able to protect against before this change but not after.

src/modules/exploit_detection/p_exploit_detection.c
```diff
@@ -1423,15 +1424,17 @@ static int p_cmp_tasks(struct p_ed_process *p_orig, char p_kill) {
          p_ret++;
       }
 
-      if (unlikely(p_orig->p_ed_task.p_sec.sec.mode != p_current->seccomp.mode)) {
-         if (p_current->seccomp.mode < 0 || p_current->seccomp.mode > 2
+      p_seccomp_mode = READ_ONCE(p_current->seccomp.mode);
```
kerneltoast, line 1435: Add a comment explaining why `READ_ONCE` is needed here.

src/modules/exploit_detection/p_exploit_detection.c
```diff
@@ -1423,15 +1424,17 @@ static int p_cmp_tasks(struct p_ed_process *p_orig, char p_kill) {
...
+      if (unlikely(p_orig->p_ed_task.p_sec.sec.mode != p_seccomp_mode) &&
+          likely(p_seccomp_mode != 3 /* SECCOMP_MODE_DEAD */)) {
```
kerneltoast: Restrict this to kernels >=5.17.

solardiz:
Thank you for your review and suggestions @kerneltoast!

> ...but paranoid mode could observe `SECCOMP_MODE_DEAD` before the task hits `do_exit()`.

I think not only paranoid mode, but also integrity checking on timer.

solardiz:
> Furthermore, I think permitting `SECCOMP_MODE_DEAD` should be limited only to kernels >=5.17 which is when `SECCOMP_MODE_DEAD` was introduced.

This version check would need to be more complicated because `SECCOMP_MODE_DEAD` was also backported to some stable kernels.

solardiz:
> Furthermore, I think permitting SECCOMP_MODE_DEAD should be limited only to kernels >=5.17 which is when SECCOMP_MODE_DEAD was introduced.

Also 5.16.10+, 5.15.24+, 5.10.211+, 5.4.180+, but I worry about possible distro kernel backports, such as RHEL's.

solardiz:
> Also 5.16.10+, 5.15.24+, 5.10.211+, 5.4.180+, but I worry about possible distro kernel backports, such as RHEL's.

Well, I can't find `SECCOMP_MODE_DEAD` in CentOS 9 Stream's 5.14.0-611.el9. But since the Fixes tag mentions older kernels and since the fix was backported to older upstream stable branches, I guess RH may later realize they also need this fix - which would then break LKRG if we add a kernel version check now.

src/modules/exploit_detection/p_exploit_detection.c
solardiz: Comment added.

solardiz:
One option may be to condition our support for `SECCOMP_MODE_DEAD` on kernel versions that have the bug commit from the Fixes tag in the commit that introduces `SECCOMP_MODE_DEAD`. That way, the exposure to attack would be limited to kernels that have this bug but not the fix (that should but don't actually have `SECCOMP_MODE_DEAD`). Unfortunately, this would include RHEL9 until they backport the fix commit introducing `SECCOMP_MODE_DEAD`. So I don't really like this option, but I thought I'd mention. It'd be an improvement upon the current state of this PR, seemingly without drawbacks.

kerneltoast:
Given that this should be a very rare scenario, and the spurious detection by LKRG results in LKRG killing the task when it was going to be killed anyway, I think we should just limit the `SECCOMP_MODE_DEAD` support to kernel versions which actually have `SECCOMP_MODE_DEAD`. It doesn't need to be perfect—I think a best effort approach is sufficient.

Put differently, I think this bug is largely cosmetic, and therefore we should not potentially compromise security in pursuit of fixing this bug. What do you think?

solardiz:
@kerneltoast OK, this makes sense to me. I'll implement the version check for kernels currently known to have this mode.

src/modules/exploit_detection/p_exploit_detection.c
solardiz: @kerneltoast Kernel version check added, please re-review. Thank you!

Issue #215, opened 2022-07-23
ovl_create_or_link inlining leads to false positive 'off' flag corruption for dockerd


ajakk:
I'm running lkrg-0.9.4 on Gentoo. Docker is at 20.10.16.

```sh
$ docker run hello-world
docker: error during connect: Post "http://%2Fvar%2Frun%2Fdocker.sock/v1.41/containers/create": EOF.
See 'docker run --help'.
```

This seems to be the interesting part of dmesg, with log level set to 6:

```
Jul 23 12:35:23 sol kernel: LKRG: ALERT: DETECT: Task: 'off' flag corruption for pid 977874, name dockerd
Jul 23 12:35:23 sol kernel: LKRG: DEBUG: 'off' flag[0x35d561dd823507c9] (normalization via 0x5fb43c3475b39c1)
Jul 23 12:35:23 sol kernel: LKRG: ALERT: BLOCK: Task: Killing pid 977874, name dockerd
Jul 23 12:35:23 sol kernel: LKRG: WATCH: Removing pid 977875
Jul 23 12:35:23 sol kernel: LKRG: WATCH: Removing pid 977880
Jul 23 12:35:23 sol kernel: LKRG: WATCH: Removing pid 977879
Jul 23 12:35:23 sol kernel: LKRG: WATCH: Removing pid 977881
Jul 23 12:35:23 sol kernel: LKRG: WATCH: Removing pid 977910
Jul 23 12:35:23 sol kernel: LKRG: WATCH: Removing pid 977908
Jul 23 12:35:23 sol kernel: LKRG: WATCH: Removing pid 977884
Jul 23 12:35:23 sol kernel: LKRG: WATCH: Removing pid 977870
Jul 23 12:35:23 sol kernel: LKRG: WATCH: Removing pid 977883
Jul 23 12:35:23 sol kernel: LKRG: WATCH: Removing pid 977878
Jul 23 12:35:23 sol kernel: LKRG: WATCH: Removing pid 977876
Jul 23 12:35:23 sol kernel: LKRG: WATCH: Removing pid 977907
Jul 23 12:35:23 sol kernel: LKRG: WATCH: Removing pid 977882
Jul 23 12:35:23 sol kernel: LKRG: WATCH: Removing pid 977877
Jul 23 12:35:23 sol kernel: LKRG: WATCH: Removing pid 977868
Jul 23 12:35:23 sol kernel: LKRG: WATCH: Removing pid 978087
[...]

PR #418, opened 2025-08-26
Support overlayfs tmpfile introduced in 6.10

kerneltoast:


solardiz:
I've just installed this on Arch Linux with 6.14.3-arch1-1 running Alpine in Docker, no issues so far.

solardiz:
Thank you very much @kerneltoast for figuring out the likely problem and proposing this fix!

Reviewing this PR against Linux 6.12.43, I see that `ovl_dentry_init_reval` (which you propose we hook) is called from `ovl_instantiate`, which is called from `ovl_create_tmpfile` shortly before it'd call `revert_creds` in the non-error path (and in one error path). The (other) error paths worry me, but moreover I don't know whether all other callers of `ovl_instantiate` (which are `ovl_create_upper` and `ovl_create_over_whiteout`) are precisely the places where we need this hook. There's one more direct call to `ovl_dentry_init_reval` from `ovl_lookup`, in which I at least see that it proceed to `revert_creds` shortly after. Is this exactly the full set of places where we need this, or where it at least somehow does not hurt? What's your analysis?

solardiz:
The old discussion in https://github.com/lkrg-org/lkrg/issues/215#issuecomment-1193198756 suggests we needed this hook and magic at all because of duplicate `override_creds` in `ovl_create_or_link`, where the second/duplicate call was made via `ovl_setup_cred_for_create`. I see the latter function is also called from `ovl_create_tmpfile`, so it does appear we have the same kind of problem for `ovl_create_tmpfile` and thus `ovl_tmpfile`. So this part of the fix may be correct:

> I see that `ovl_dentry_init_reval` (which you propose we hook) is called from `ovl_instantiate`, which is called from `ovl_create_tmpfile` shortly before it'd call `revert_creds` in the non-error path

solardiz:
> I've just installed this on Arch Linux with 6.14.3-arch1-1 running Alpine in Docker, no issues so far.

I'm afraid this test doesn't mean much because I am unable to reproduce any issue (not even an old one) even when I remove this hook's main function:

```diff
+++ b/src/modules/exploit_detection/syscalls/override/overlayfs/p_ovl_override_sync/p_ovl_override_sync.c
@@ -42,7 +42,7 @@ static notrace int p_ovl_override_sync_ret(struct kretprobe_instance *ri, struct
 #ifdef P_LKRG_TASK_OFF_DEBUG
             p_debug_off_flag_override_on(p_tmp, 30, p_regs);
 #endif
-            p_set_ed_process_override_on(p_tmp);
+//            p_set_ed_process_override_on(p_tmp);
          }
          p_ed_validate_current(p_tmp);
          p_ed_enforce_pcfi(NULL, p_tmp, p_regs);
```

and I can start MariaDB in an Alpine container, no issue seen.

solardiz:
> I can start MariaDB in an Alpine container, no issue seen.

... but with debugging output added (on top of this PR), I see that `p_ovl_override_sync_ret` is called many times on MariaDB startup in container, and a few times on killing that MariaDB. Still no ALERTs or anything with `p_set_ed_process_override_on` commented out.

solardiz:
This:

```diff
+++ b/src/modules/exploit_detection/syscalls/override/overlayfs/p_ovl_override_sync/p_ovl_override_sync.c
@@ -36,9 +36,12 @@ static notrace int p_ovl_override_sync_ret(struct kretprobe_instance *ri, struct

    struct p_ed_process *p_tmp;

+   p_print_log(P_LOG_ISSUE, "p_ovl_override_sync_ret called");
+
    if (p_is_ed_task(current)) {
       if ((p_tmp = ed_task_lock_current())) {
          if (p_verify_ovl_override_sync(p_tmp)) {
+            p_print_log(P_LOG_ISSUE, "p_ovl_override_sync_ret p_verify_ovl_override_sync true");
 #ifdef P_LKRG_TASK_OFF_DEBUG
             p_debug_off_flag_override_on(p_tmp, 30, p_regs);
 #endif
```

shows the hook is called 16 times on MariaDB start, and 4 times on it being killed. But the inner print of "p_verify_ovl_override_sync true" is never reached, which at least explains why commenting out `p_set_ed_process_override_on` made no difference for me.

solardiz:
@kerneltoast Now that @roedie reported success with this PR, what do you suggest we do with it next? Merge it? What's your confidence level (and why) that these changes don't make things worse in other ways? Thank you!

kerneltoast:
`ovl_dentry_is_whiteout` was never the perfect probe to cover only the double `override_creds` cases we care about, because it's also called from `ovl_rename` which we _don't_ care about.

Similarly, probing `ovl_dentry_init_reval` is not perfect but it is the closest match that covers both `ovl_create_tmpfile` and `ovl_create_or_link`, with just a single extraneous call site that we don't care to probe—the one in `ovl_lookup`.

I don't think this PR makes things worse in other ways; rather, it keeps the existing overlayfs hacks on life support longer.

I think a better approach to the overlayfs probing would be to set up probes using the function pointers in `ovl_dir_inode_operations` for functions which call `ovl_create_tmpfile` and `ovl_create_or_link`. So that would be: `.mkdir`, `.symlink`, `.link`, `.create`, `.mknod`, and `.tmpfile`.

solardiz:
> I think a better approach to the overlayfs probing would be to set up probes using the function pointers in `ovl_dir_inode_operations` for functions which call `ovl_create_tmpfile` and `ovl_create_or_link`. So that would be: `.mkdir`, `.symlink`, `.link`, `.create`, `.mknod`, and `.tmpfile`.

Wouldn't this result in us assuming `override_creds` even in cases when it's not actually reached (such as these functions ending up on their error paths)? I worry that we may already have this sort of problem, but to a lesser extent.

kerneltoast:
> > I think a better approach to the overlayfs probing would be to set up probes using the function pointers in `ovl_dir_inode_operations` for functions which call `ovl_create_tmpfile` and `ovl_create_or_link`. So that would be: `.mkdir`, `.symlink`, `.link`, `.create`, `.mknod`, and `.tmpfile`.
> 
> Wouldn't this result in us assuming `override_creds` even in cases when it's not actually reached (such as these functions ending up on their error paths)? I worry that we may already have this sort of problem, but to a lesser extent.

We could just check the return value for error. When there's no error, a double `override_creds` is guaranteed to have occurred in all of those functions _except_ for `ovl_link`. `ovl_link` is an exception because it may set `attr->hardlink` to a non-NULL value, in which case `ovl_create_or_link` will only do a single `override_creds`. Even if we didn't care to accurately decipher the number of `override_creds` calls for `ovl_link`, I think this is already far superior in both reliability and precision compared to the existing scheme.

solardiz:
> > I've just installed this on Arch Linux with 6.14.3-arch1-1 running Alpine in Docker, no issues so far.
> 
> I'm afraid this test doesn't mean much because I am unable to reproduce any issue (not even an old one) even when I remove this hook's main function

I now realize that my testing here was ridiculous since on 6.13+ we don't hook override/revert_creds, so our overlayfs hooking is irrelevant. This also reminds me that on 6.13+ we may want not to hook anything overlayfs as a simplification/optimization.

roedie:
@solardiz You want me to re-test to be sure it is fixed/mitigated? Or is there any other way I can help?

solardiz:
@roedie I'd help if you test/use our latest code from this repo (`main` branch), which has these changes along with other changes in it.

I'd also help if you _separately_ re-test #420 (that PR's branch alone, _not_ on top of other recent changes). Right now, @kerneltoast and I suspect that you did not test correctly when you concluded the issue went away with those changes alone. It would help to know whether the issue is still reproducible with those changes (most likely) or not (puzzling).

Thank you very much!

Issue #409, opened 2025-08-10
Consider dropping certain hooks until we actually need them


solardiz:
@Adam-pi3 I don't know why we're hooking certain functions. As I recently found out, we assume that we have some untracked `commit_creds` anyway, where the credentials that we do track remain unchanged, so we suppress detection of cred pointer changes in those cases.

Should we perhaps drop some hooks while we can? We can always reintroduce them later as we start tracking more things, but for 1.0 I'd like to reduce the bloat, as well as avoid unjustified performance overhead, and to consistently only hook what we currently need to hook so that the logic is easier to understand.

LKRG with this patch (on top of #407) still passes all of our CI tests (and works in my dev VM now):

```diff
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -148,6 +148,7 @@ static const struct p_functions_hooks {
      NULL,
      0
    },
+#if 0
    { "set_current_groups",
      p_install_set_current_groups_hook,
      p_uninstall_set_current_groups_hook,
@@ -155,6 +156,7 @@ static const struct p_functions_hooks {
      NULL,
      1
    },
+#endif
    { "generic_permission",
      p_install_generic_permission_hook,
      p_uninstall_generic_permission_hook,
@@ -187,6 +189,7 @@ static const struct p_functions_hooks {
      NULL,
      1
    },
+#if 0
    /* Caps. */
    { "sys_capset",
      p_install_sys_capset_hook,
@@ -231,7 +234,8 @@ static const struct p_functions_hooks {
      NULL,
      0
    },
-#ifdef CONFIG_COMPAT
+#endif
+#if 0 // def CONFIG_COMPAT
    { "compat_sys_keyctl",
      p_install_compat_sys_keyctl_hook,
      p_uninstall_compat_sys_keyctl_hook,
@@ -265,7 +269,7 @@ static const struct p_functions_hooks {
   #endif /* CONFIG_X86 */
  #endif /* P_SYSCALL_LAYOUT_4_17 */
 #endif /* CONFIG_COMPAT */
-#ifdef CONFIG_X86_X32
+#if 0 // def CONFIG_X86_X32
  #ifdef P_SYSCALL_LAYOUT_4_17
    { "x32_sys_keyctl",
      p_install_x32_sys_keyctl_hook,
```

solardiz:
BTW, most of these hook functions look like they were created by copy-pasting one for `capset` (which I also propose to drop), and they inherit a wrong debug message:

```
[...]

PR #420, opened 2025-08-28
Improve cred pointer checks, drop many unnecessary hooks

solardiz:
### Description

The first commit here revises the cred pointer checks further, so that we perform them regardless of whether the actual credentials differ or not, yet don't report spurious *cred pointer corruption on untracked commit_creds(), not even when an exploit concurrently overwrote the actual credentials.

Further commits drop many hooks relying on the fact that we support untracked commit_creds() anyway, as long as no tracked credentials are modified.

### How Has This Been Tested?

As mentioned in #409, I've been testing equivalent changes (with `#if 0` of the hook installation) on several different systems for around a week. These specific changes have so far been tested in my dev VM, in a CentOS 7.9 VM (where I also ran some `keyctl` commands adding/showing keys as root and user), and in our CI setup. I intend to test them more.

### DIFF
--- a/src/modules/exploit_detection/p_exploit_detection.h
+++ b/src/modules/exploit_detection/p_exploit_detection.h
@@ -57,33 +57,18 @@
    *
    *  [*] Used
    *
-   * CONFIG_COMPAT / CONFIG_IA32_EMULATION and CONFIG_X86_X32 is covered here.
-   *
+   * CONFIG_X86_X32 is covered here.
    */
 
  #ifdef CONFIG_X86_64
 
-  #define P_SYSCALL_LAYOUT_4_17
-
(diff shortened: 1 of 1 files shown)


solardiz:
> I've been testing equivalent changes (with `#if 0` of the hook installation) on several different systems for around a week.

I need to clarify. I've only been testing the dropping of hooks. Not also the first commit here, which revises the cred pointer checks. I expect this commit to actually improve behavior (or I wouldn't be making those changes), especially in case of concurrent exploit overwriting creds and untracked legitimate `commit_creds` (which becomes more relevant with us dropping tracking of more `commit_creds`), but it does need more testing.

@Adam-pi3 @kerneltoast Your review would be appreciated.

PR #427, opened 2025-08-30
Don't hook anything OverlayFS on 6.13+

### LINKED ISSUE #426: Don't hook overlayfs on 6.13+
We hook an overlayfs function in order to compensate for unbalanced override/revert_creds. However, on 6.13+ we don't hook those, so we also don't need to compensate, and thus shouldn't need to hook anything overlayfs. Let's try to omit such hooking on those kernels as a simplification and optimization.

solardiz:
### Description

We can't/don't hook override/revert_creds() on 6.13+, so we also don't need to hook anything OverlayFS on those kernels.

Fixes #426

### How Has This Been Tested?

In my dev VM (pre-6.13 kernel), in an Arch Linux VM with 6.14+ and a Docker container, and in our CI setup.

### DIFF
--- a/src/modules/exploit_detection/syscalls/override/overlayfs/p_ovl_override_sync/p_ovl_override_sync.h
+++ b/src/modules/exploit_detection/syscalls/override/overlayfs/p_ovl_override_sync/p_ovl_override_sync.h
@@ -6,9 +6,9 @@
  *    'ovl_create_or_link'.
  *
  * Notes:
- *  - We are maintianing Red-Black tree of pid's for Exploit Detection feature.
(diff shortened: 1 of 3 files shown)


solardiz:
I also confirmed that in the Arch Linux VM with 6.14.3-arch1-1 there was:

```
# grep ovl /sys/kernel/debug/kprobes/list 
ffffffffc0e88434  r  ovl_dentry_init_reval+0x4  overlay [FTRACE]
```

with LKRG revision just prior to this PR, and with this PR's changes this hook is gone from there.

solardiz:
Added a commit, re-tested the same things with that commit included.

PR #431, opened 2025-09-01
pINT: Do not misuse get_cred(), get_task_struct(), memory barriers

### LINKED ISSUE #430: LKRG may trigger the kernel's __put_cred BUG_ON()
I observed that when LKRG is in paranoid mode and certain stress-test workloads are running at once, the kernel complains about a BUG in cred.c line 103, which is:

```c
/**
 * __put_cred - Destroy a set of credentials
 * @cred: The record to release
 *
 * Destroy a set of credentials on which no references remain.
 */
void __put_cred(struct cred *cred)
{
        kdebug("__put_cred(%p{%ld})", cred,
               atomic_long_read(&cred->usage));

        BUG_ON(atomic_long_read(&cred->usage) != 0);
```

The backtraces it prints from there do not include LKRG, but rather (at least in the case of my workloads) appear to include `security_task_prctl` (that's with a revision of LKRG where we no longer hook anything `prctl`, so must be unrelated to our hooks).

Indeed, in our `p_cmp_tasks` we have a couple of `get_cred`, which could be for a non-current task if that's what we're validating (in paranoid mode or on timer). I suppose that if that non-current task is exiting, it could "know" it's allowed to use `__put_cred`, which just assumes that no references remain - but we violate this assumption.

solardiz:
Fixes #430

### Description

1. Drop all uses of `get_cred` / `put_cred`.
2. Add `rcu_read_lock` / `rcu_read_unlock` to `p_cmp_tasks`, even though the callers either already have those (for other reasons) or are for the current task.
3. Drop the only remaining use of `get_task_struct`.
4. Drop two memory barriers.

### How Has This Been Tested?

Light testing in my dev VM, and by our CI.

The first commit also got ~3 hours of stress-testing on the system where I previously saw #430 (not triggering it anymore), which now continues to stress-tests all 3 commits.

### DIFF
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -404,7 +404,6 @@ notrace void p_update_ed_process(struct p_ed_process *p_source, char p_stack) {
    p_print_log(P_LOG_WATCH, "Updating pid %u", p_task->pid);
 
    rcu_read_lock();
-   get_task_struct(p_task);
    /* Track process's metadata */
    p_source->p_ed_task.p_pid                      = p_task->pid;
    p_source->p_ed_task.p_real_cred_ptr            = rcu_dereference(p_task->real_cred);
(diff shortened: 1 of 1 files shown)


solardiz:
BTW, we skip on RCU in `p_cmp_cred_ptr`, which in practice is for the current task only and does not dereference the pointers.

solardiz:
@kerneltoast Thanks. Commit message changed to what you suggested.

I didn't realize/recall that unregister_kprobe/unregister_kretprobe wait for hooks to finish running. I now vaguely recall you mentioning this before.

PR #433, opened 2025-09-01
pINT: Skip __queue_work() probe when called from any non-task context

### LINKED ISSUE #432: Occasional spurious "real_cred pointer corruption" for bwrap (Flatpak app startup) on 6.14.3
I was running more stress-tests for/after the #429 fix and to test #430, and this triggered:

```
2025-09-01T08:58:31Z - LKRG: ALERT: DETECT: Task: real_cred pointer corruption for pid 683336, name bwrap
2025-09-01T08:58:31Z - LKRG: ALERT: BLOCK: Task: Killing pid 683336, name bwrap
```

just once so far in 4+ hours of continuous testing of a Flatpak app startup running in a loop with a `sleep 0.1` between tries (so maybe 5 to 10 times per second).

This is similar to but not the same as #429 (that one prevented Flatpak app startup on this affected system all the time, not occasionally), and I think the #430 fix should make no difference for it (the presence or omission of `get_cred` would not affect the pointer, and the complaint above is only about the pointer).

So I suspect there may be a logic error in my recently added check:

```
    * commit_creds() changes both pointers, override_creds() only the cred
    * pointer, but nothing valid changes only the real_cred pointer, so that's
    * what we can detect as a violation on all kernels.  This violation implies
    * that the two current pointers differ from each other, but not the other
    * way around.
```

```c
       p_orig->p_ed_task.p_real_cred_ptr != p_current_real_cred &&
       p_orig->p_ed_task.p_real_cred_ptr == p_current_cred
```

This is like `override_creds`, but the other way around - changing only `real_cred`, but not `cred` - which I thought never existed. But maybe it does exist somewhere in the kernel?! Also, if it does exist in paths other than those involving `override_creds`, it may trigger our stricter check `p_current_real_cred != p_current_cred` on pre-6.13.

The specific error messages here, along with our current code, imply that only `real_cred` pointer changed. Not only the `cred` pointer did not change, but also none of our tracked credentials (such as uids) changed as read over either pointer. In other words, we know that the two pointers are different, and that only `real_cred` changed, but the actual credentials pointer to by `real_cred` are still the same as far as we're concerned.

solardiz:
### Description

Our task integrity checks assume that the current task is in consistent state, which was apparently not guaranteed here.

Amends 863b28d1ad7d74046b321956df76ec47ae4cfb94
Fixes #432

### How Has This Been Tested?

With extra debugging print on Arch Linux with 6.14.3 as described in #432. Also in a CentOS 7.9 VM just to see that there's no apparent breakage. Now waiting for our CI jobs to complete.

I don't know whether this actually fixes #432, but this is currently my only plausible guess as to what caused the symptoms seen there (just once, not reproduced yet).

### DIFF
--- a/src/modules/exploit_detection/syscalls/pCFI/p___queue_work/p___queue_work.c
+++ b/src/modules/exploit_detection/syscalls/pCFI/p___queue_work/p___queue_work.c
@@ -26,7 +26,7 @@ static int p_pcfi___queue_work_entry(struct kprobe *p_ri, struct pt_regs *p_regs
 
    p_ed_pcfi_cpu(0);
 
-   if (!in_irq() && p_is_ed_task(current)) {
+   if (in_task() && p_is_ed_task(current)) {
       /* Do not take ED lock */
       if ((p_tmp = ed_task_trylock_current())) {
          p_ed_validate_current(p_tmp);


kerneltoast:
NAK. Writing up a longer explanation now, but wanted to state that this shouldn't be merged first since you're merging fixes quickly at the moment.

solardiz:
Thank you for the quick NAK @kerneltoast, it really does help now!

kerneltoast:
The whole reason `__queue_work()` is probed is in order to get more "unpredictable" probe points into a running task, since LKRG doesn't cover every possible syscall right now.

The reason I ignored the `__queue_work()` probe from hard IRQ context was because it created a lot of synchronization issues that weren't worth the risk/hassle just for that one probe. So the hard rule for our synchronization is no hard IRQ context.

That doesn't mean we can't probe from softirq context, which is what this change achieves: `in_task()` is true in softirq context in addition to hard IRQ context. It _also_ picks up NMIs, which is useless for kprobes because kprobe handlers aren't ever run from NMI context.

The way in which softirq executes within the current task is when the current task disables softirqs and then re-enables them, such as `spin_lock_bh()` followed by `spin_unlock_bh()`: the unlock_bh will check for pending softirqs and service them from the current context if possible.

Since there's no such pattern in the kernel where creds are in an inconsistent state upon softirqs executing from task context (_because_ softirq won't randomly execute from task context), this change simply weakens the `__queue_work()` probe.

IOW: In order for this change to fix a problem, that problem must be related to softirqs running from task context, which only occurs when the task calls `local_bh_enable()`.

solardiz:
Thanks @kerneltoast. IOW, you're confident that #432 must be caused solely by something else, correct?

kerneltoast:
> Thanks @kerneltoast. IOW, you're confident that #432 must be caused solely by something else, correct?

Yes.

Issue #443, opened 2025-10-10
LKRG's p_main_error: cleanup code does not accommodate certain codebase changes in project forks


qhdhuowen:
The content below "p_main_error" in function p_lkrg_register() might be incorrect. Once the code flow reaches "p_main_error", the system kernel will crash, regardless of the kernel version. Has anyone encountered this problem?

Adam-pi3:
What kernel do you use? can you provide a log messages from the crash?

qhdhuowen:
Thank you for your reply.

```
   p_print_log(P_LOG_ALIVE, "LKRG initialized successfully");

   p_ret = P_LKRG_SUCCESS;
   p_ret = P_LKRG_GENERAL_ERROR;
   p_print_log(P_LOG_ALIVE, "this is a test code designed to test p_main_error.");

p_main_error:

   if (p_ret != P_LKRG_SUCCESS) {
      p_print_log(P_LOG_DYING, "Not loading LKRG (initialization failed)");
      P_CTRL(p_kint_validate) = 0;
```

You can trigger the questions in the above way. Of course, the above is just the test code for triggering questions and will not be used in the production environment.The crash information is as follows：

```
[   35.913581] LKRG: ALIVE: Loading LKRG
[   35.916194] Freezing user space processes
[   35.917410] Freezing user space processes completed (elapsed 0.001 seconds)
[   35.917413] OOM killer disabled.
[   35.983208] LKRG: ISSUE: register_k[ret]probe() for ovl_dentry_is_whiteout failed! [err=-2]
[   35.983211] LKRG: ISSUE: Can't hook 'ovl_dentry_is_whiteout'. This is expected when OverlayFS is not used.
[   36.056313] LKRG: ALIVE: LKRG initialized successfully
[   36.056315] LKRG: ALIVE: LKRG:this is a test code designed to test p_main_error.
[   36.056316] LKRG: DYING: Not loading LKRG (initialization failed)
[   36.056321] BUG: unable to handle page fault for address: ffffffffc0e18010
[   36.056335] #PF: supervisor write access in kernel mode
[   36.056343] #PF: error_code(0x0003) - permissions violation
[   36.056352] PGD 495c24067 P4D 495c24067 PUD 495c26067 PMD 12c853067 PTE 800000012c859121
[   36.056366] Oops: 0003 [#1] PREEMPT SMP NOPTI
[...]

Issue #281, opened 2023-07-13
Setting/unsetting lkrg.hide makes the module appear "in use" and not removable


solardiz:
After these two commands:

```shell
sysctl lkrg.hide=1
sysctl lkrg.hide=0
```

`lkrg` is correctly seen in `lsmod` again (between the commands, it correctly is not), but with `-2` in the column showing how many other modules use it. In that state, `rmmod lkrg` refuses to remove it saying `rmmod: ERROR: Module lkrg is in use`, but `rmmod -f lkrg` works fine.

In a separate test (not trying to unload the module yet), repeating the `sysctl` commands doesn't appear to change the usage count further (it's still at `-2` after two more commands like this).

Adam-pi3:
Yes, it was from the beginning of that feature. I'm not exactly sure why, but it is related to KOBJs which we clear and recreate. I don't see any issues with such behavior. `rmmod -f` is doing all the work fine and `-2` is probably a confusion for the kernel of loosing the track of original KOBJs which now appeared again. However, I never go through the details of it.

PR #444, opened 2025-10-10
Make cleanup on failed load more generic (accommodate project forks)

solardiz:
### Description

This makes cleanup on failed load more generic, such as to accommodate project forks as seen in #443.

I am moderately unhappy that this adds some complexity and lines of code. We could instead delete two lines (that I just found were misplaced anyway, fixed here) and declare such code changes in forks unsupported (it'd be their problem to add this code if they need it).

### How Has This Been Tested?

In my dev VM, including with the testcase from #443 and running the `sysctl` command after failed load. Also mixing failed and successful loads/unloads during a system uptime.

This is also being tested in our CI setup now (but I think we don't specifically test failed loads, and maybe not even unload).

### DIFF
--- a/src/p_lkrg_main.c
+++ b/src/p_lkrg_main.c
@@ -583,12 +583,6 @@ static int __init p_lkrg_register(void) {
       goto p_main_error;
    }
 
-#ifdef LKRG_WITH_HIDE
-   if (P_CTRL(p_hide_lkrg)) {
-      p_hide_itself();
-   }
-#endif
-
    p_integrity_timer();
    p_register_notifiers();
    p_init_page_attr();


solardiz:
Oh, the last commit here (for unhiding) just failed my testing with:

```diff
+++ b/src/p_lkrg_main.c
@@ -64,7 +64,7 @@ p_ro_page p_ro = {
       .p_trigger = 0,                     // trigger
       .p_block_modules = 0,               // block_modules
 #ifdef LKRG_WITH_HIDE
-      .p_hide_lkrg = 0,                   // hide_lkrg
+      .p_hide_lkrg = 1,                   // hide_lkrg
 #endif
       .p_heartbeat = 0,                   // heartbeat
 #if defined(CONFIG_X86)
@@ -595,6 +595,7 @@ static int __init p_lkrg_register(void) {
    p_print_log(P_LOG_ALIVE, "LKRG initialized successfully");
 
    p_ret = P_LKRG_SUCCESS;
+   p_ret = P_LKRG_GENERAL_ERROR;
 
 p_main_error:
 
```

It got a `list_add double add` (same `new` and `next` pointers) apparently in `p_unhide_itself`.

solardiz:
> It got a `list_add double add` (same `new` and `next` pointers) apparently in `p_unhide_itself`.

The reason appears to be that the hiding didn't work (but LKRG itself thought it did). There are indeed other issues related to that, even without that commit. Investigating.

solardiz:
> hiding didn't work (but LKRG itself thought it did). There are indeed other issues related to that, even without that commit. Investigating.

Well, I just tested that the normal hiding/unhiding via sysctl works on this system (except for the known issue #281). So I am simply dropping the broken support for hiding by default for now, which wasn't exposed for use without a change of the hard-coded default anyway.

solardiz:
Weird, CI job [mkosi-boot (plucky)](https://github.com/lkrg-org/lkrg/actions/runs/18433760464/job/52524500501#logs) failed after you merged this, looks unrelated, but possibly a deadlock bug in our use of kernel locks.

Edit: log excerpt removed since it's now in its separate issue.

solardiz:
> Weird, CI job [mkosi-boot (plucky)](https://github.com/lkrg-org/lkrg/actions/runs/18433760464/job/52524500501#logs) failed after you merged this, looks unrelated

I re-ran the job, and as expected it succeeded. I'll open a separate issue for this intermittent failure.

solardiz:
> > Weird, CI job [mkosi-boot (plucky)](https://github.com/lkrg-org/lkrg/actions/runs/18433760464/job/52524500501#logs) failed after you merged this, looks unrelated
> 
> I re-ran the job, and as expected it succeeded. I'll open a separate issue for this intermittent failure.

This is now #445

Issue #21, opened 2020-10-30
Add 'notrace' and 'static' attribute to the most of our internal functions


Adam-pi3:

solardiz:
Adam, please provide some rationale for this. Thanks!

Adam-pi3:
Sure. When FTRACE is enabled GCC compiles every code with -pg option. It adds a stub at the beginning of each function (in fact long NOP) which can be dynamically "overwritten" and place 'jump' to the specified code to execute when needed. It also applies to LKRG's functions (including critical functions).
However, some of the functions in the kernel are very critical and we do not want to allow such functionality. Linux kernel to protect from that introduced 'notrace' attribute and such functions can't be hooked (e.g. FTRACE's internal functions to avoid deadlocks).
All available functions for hooking are listed in user-mode via FTRACE interface (e.g. /sys/kernel/debug/tracing/available_filter_functions). From FTRACE perspective they are called 'filters'. We can set-up custom code for 'filters' and set-up various actions like tracking their perf etc. However, dynamic ftrace (from kernel) can also modify their context (e.g. registers, including PC). In general it's a potential security issue for LKRG. We should avoid that feature.

solardiz:
Adam, while at it please also be adding `static` to most functions. Thanks!

solardiz:
Now we also have many functions in the embedded libhydrogen. It'd be nasty to make changes to those as that would make comparisons and sync'ing to latest upstream more difficult. I don't have a good suggestion on what to do about that. But until we get there, we still have plenty of LKRG-specific functions to add `notrace` to.

solardiz:
It looks like more consistent use of `notrace` by us would break Singularity rootkit's current LKRG bypass, where it hooks our functions.

@kerneltoast suggests we could also try `ccflags-remove-y := $(CC_FLAGS_FTRACE)` instead of `notrace` in the source... but I guess it's better in addition, because some people integrate LKRG into the kernel perhaps without usage of our `Makefile`?

kerneltoast:
> because some people integrate LKRG into the kernel perhaps without usage of our `Makefile`?

I think it counts as user error to integrate LKRG into the kernel without usage of `scripts/copy-builtin.sh`, which indeed generates a `Makefile` for LKRG.

solardiz:
> I think it counts as user error to integrate LKRG into the kernel without usage of `scripts/copy-builtin.sh`

[...]

PR #454, opened 2026-01-27
Remove trace-related CFLAGS from the build

kerneltoast:
Tested both modular and in-tree builds to verify the trace-related flags are indeed removed from all of LKRG's objects.

### DIFF
--- a/src/modules/database/FTRACE/p_ftrace_enable_sysctl/p_ftrace_enable_sysctl.c
+++ b/src/modules/database/FTRACE/p_ftrace_enable_sysctl/p_ftrace_enable_sysctl.c
@@ -31,7 +31,7 @@
 
 #include "../../../exploit_detection/syscalls/p_install.h"
 
-static notrace int p_ftrace_enable_sysctl_entry(struct kprobe *p_ri, struct pt_regs *p_regs) {
+static int p_ftrace_enable_sysctl_entry(struct kprobe *p_ri, struct pt_regs *p_regs) {
 
    p_regs_set_arg2(p_regs, 0);
 
--- a/src/modules/database/FTRACE/p_ftrace_modify_all_code/p_ftrace_modify_all_code.c
+++ b/src/modules/database/FTRACE/p_ftrace_modify_all_code/p_ftrace_modify_all_code.c
@@ -46,7 +46,7 @@ unsigned int p_ftrace_tmp_mod;
  * static int ftrace_modify_all_code(unsigned long pc, unsigned long old,
  *                                   unsigned long new, bool validate)
  */
-static notrace int p_ftrace_modify_all_code_entry(struct kretprobe_instance *p_ri, struct pt_regs *p_regs) {
+static int p_ftrace_modify_all_code_entry(struct kretprobe_instance *p_ri, struct pt_regs *p_regs) {
 
    struct ftrace_rec_iter *p_iter;
    struct dyn_ftrace *p_rec;
@@ -103,7 +103,7 @@ static notrace int p_ftrace_modify_all_code_entry(struct kretprobe_instance *p_r
 }
 
 
-static notrace int p_ftrace_modify_all_code_ret(struct kretprobe_instance *ri, struct pt_regs *p_regs) {
+static int p_ftrace_modify_all_code_ret(struct kretprobe_instance *ri, struct pt_regs *p_regs) {
 
    unsigned int p_tmp,p_tmp2;
    unsigned char p_flag = 0;
--- a/src/modules/database/JUMP_LABEL/p_arch_jump_label_transform/p_arch_jump_label_transform.c
+++ b/src/modules/database/JUMP_LABEL/p_arch_jump_label_transform/p_arch_jump_label_transform.c
@@ -31,7 +31,7 @@
 p_lkrg_counter_lock p_jl_lock;
 
 
-static notrace int p_arch_jump_label_transform_entry(struct kretprobe_instance *p_ri, struct pt_regs *p_regs) {
+static int p_arch_jump_label_transform_entry(struct kretprobe_instance *p_ri, struct pt_regs *p_regs) {
 
    struct jump_entry *p_tmp = (struct jump_entry *)p_regs_get_arg1(p_regs);
    unsigned long p_addr = p_jump_entry_code(p_tmp);
@@ -85,7 +85,7 @@ static notrace int p_arch_jump_label_transform_entry(struct kretprobe_instance *
 }
 
 
-static notrace int p_arch_jump_label_transform_ret(struct kretprobe_instance *ri, struct pt_regs *p_regs) {
+static int p_arch_jump_label_transform_ret(struct kretprobe_instance *ri, struct pt_regs *p_regs) {
 
    unsigned int p_tmp,p_tmp2;
    unsigned char p_flag = 0;
--- a/src/modules/database/JUMP_LABEL/p_arch_jump_label_transform_apply/p_arch_jump_label_transform_apply.c
+++ b/src/modules/database/JUMP_LABEL/p_arch_jump_label_transform_apply/p_arch_jump_label_transform_apply.c
@@ -36,7 +36,7 @@
 static bool p_text, p_mod;
 
 
-static notrace int p_arch_jump_label_transform_apply_entry(struct kretprobe_instance *p_ri, struct pt_regs *p_regs) {
+static int p_arch_jump_label_transform_apply_entry(struct kretprobe_instance *p_ri, struct pt_regs *p_regs) {
 
 #ifdef TEXT_POKE_MAX_OPCODE_SIZE
    int p_nr = P_SYM(p_text_poke_array)->nr_entries;
@@ -109,7 +109,7 @@ static notrace int p_arch_jump_label_transform_apply_entry(struct kretprobe_inst
 }
 
 
-static notrace int p_arch_jump_label_transform_apply_ret(struct kretprobe_instance *ri, struct pt_regs *p_regs) {
+static int p_arch_jump_label_transform_apply_ret(struct kretprobe_instance *ri, struct pt_regs *p_regs) {
 
    struct module *p_module = NULL;
    unsigned int p_tmp,p_tmp2;
--- a/src/modules/database/TRACEPOINT/p_arch_static_call_transform/p_arch_static_call_transform.c
+++ b/src/modules/database/TRACEPOINT/p_arch_static_call_transform/p_arch_static_call_transform.c
@@ -32,7 +32,7 @@ static unsigned int p_module2_idx;
 
 p_lkrg_counter_lock p_static_call_spinlock;
 
-static notrace int p_arch_static_call_transform_entry(struct kretprobe_instance *p_ri, struct pt_regs *p_regs) {
+static int p_arch_static_call_transform_entry(struct kretprobe_instance *p_ri, struct pt_regs *p_regs) {
 
    unsigned long p_site = p_regs_get_arg1(p_regs);
    unsigned long p_tramp = p_regs_get_arg2(p_regs);
@@ -128,7 +128,7 @@ static notrace int p_arch_static_call_transform_entry(struct kretprobe_instance 
 }
 
 
-static notrace int p_arch_static_call_transform_ret(struct kretprobe_instance *ri, struct pt_regs *p_regs) {
+static int p_arch_static_call_transform_ret(struct kretprobe_instance *ri, struct pt_regs *p_regs) {
 
    unsigned int p_tmp;
    unsigned char p_flag = 0;
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -301,7 +301,7 @@ static void p_ed_wq_valid_cache_delete(void) {
    }
 }
 
-static notrace void p_dump_creds(struct p_cred *p_where, const struct cred *p_from) {
+static void p_dump_creds(struct p_cred *p_where, const struct cred *p_from) {
 
    /* Track process's IDs */
    p_set_uid(&p_where->uid, p_get_uid(&p_from->uid));
@@ -317,7 +317,7 @@ static notrace void p_dump_creds(struct p_cred *p_where, const struct cred *p_fr
 }
 
 #if defined(CONFIG_SECCOMP)
-static notrace void p_dump_seccomp(struct p_seccomp *p_sec, struct task_struct *p_task) {
+static void p_dump_seccomp(struct p_seccomp *p_sec, struct task_struct *p_task) {
 
    p_sec->sec.mode    = p_task->seccomp.mode;   // Mode
    p_sec->sec.filter  = p_task->seccomp.filter; // Filter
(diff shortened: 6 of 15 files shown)


solardiz:
Thank you @kerneltoast! Please look into the CI failures here.

kerneltoast:
@solardiz I don't understand why there are boot CI failures. I've tested these changes myself and LKRG loads just fine; any idea what's going on?

solardiz:
Looks like it's build time failure that we fail to detect right away @vt-alt 

```
  + echo $'::group::Building LKRG for 6.19.0-0.rc6.260123gc072629f05d7b.46.fc44.x86_64 \360\237\217\203'
  ++ nproc
  + make -O -j4 all install-module
Building LKRG for 6.19.0-0.rc6.260123gc072629f05d7b.46.fc44.x86_64 🏃
  make: Nothing to be done for 'all'.
  make -C /lib/modules/6.19.0-0.rc6.260123gc072629f05d7b.46.fc44.x86_64/build M=/workspace modules_install
    DEPMOD  /lib/modules/6.19.0-0.rc6.260123gc072629f05d7b.46.fc44.x86_64
  + echo '::group::Generate initrd image'
Generate initrd image
  + dracut -f --add virtfs --force-drivers '9p 9pnet_virtio lkrg' /boot/initrd.img 6.19.0-0.rc6.260123gc072629f05d7b.46.fc44.x86_64
  dracut[E]: No '/dev/log' or 'logger' included for syslog logging
  grep: /var/tmp/dracut.dkwW5b3/initramfs/etc/shadow: No such file or directory
  dracut-install: Failed to find module 'lkrg'
```

We then proceed to try and boot the system up anyway.

Makefile
```diff
@@ -83,9 +87,15 @@ $(TARGET)-objs += src/modules/ksyms/p_resolve_ksym.o \
...
+# Runs first during the "make" pass
+ifneq ($(REMOVE_FLAGS),)
+CFLAG_REMOVAL_OPTS := $(foreach x,$($(TARGET)-objs),CFLAGS_REMOVE_$(x)="$(REMOVE_FLAGS)")
+endif
+
 all:
-#	$(MAKE) -C $(KERNEL) M=$(P_PWD) modules CONFIG_DEBUG_SECTION_MISMATCH=y
-	$(MAKE) -C $(KERNEL) M=$(P_PWD) modules
+	$(MAKE) -C $(KERNEL) $(CFLAG_REMOVAL_OPTS) M=$(P_PWD) modules
+endif
```
solardiz: Looks like with these changes the `all` target is defined only if `KERNELRELEASE` is empty. Why is that?

kerneltoast: Oh, `KERNELRELEASE` should not be passed to `make` by the build runner. I'm using a Kbuild trick here.

From https://www.kernel.org/doc/Documentation/kbuild/modules.rst :
```
Shared Makefile
---------------

	An external module always includes a wrapper makefile that
	supports building the module using "make" with no arguments.
	This target is not used by kbuild; it is only for convenience.
	Additional functionality, such as test targets, can be included
	but should be filtered out from kbuild due to possible name
	clashes.

	Example 1::

		--> filename: Makefile
		ifneq ($(KERNELRELEASE),)
		# kbuild part of makefile
		obj-m  := 8123.o
		8123-y := 8123_if.o 8123_pci.o

		else
		# normal makefile
		KDIR ?= /lib/modules/`uname -r`/build

		default:
			$(MAKE) -C $(KDIR) M=$$PWD

		endif

	The check for KERNELRELEASE is used to separate the two parts
	of the makefile. In the example, kbuild will only see the two
	assignments, whereas "make" will see everything except these
	two assignments. This is due to two passes made on the file:
	the first pass is by the "make" instance run on the command
	line; the second pass is by the kbuild system, which is
	initiated by the parameterized "make" in the default target.
```

solardiz: Hmm, so should we change our CI setup not to pass `KERNELRELEASE`? Maybe pass the `KERNEL` path instead, like `lkrg.spec` does?

solardiz: BTW, I switched `lkrg.spec` from using `KERNELRELEASE` to `KERNEL` in daf7a964bda73cc0095c8a6cfc6ffaf271fea90f, but I think that was to avoid dependency on a symlink provided by the `kernel` package.

vt-alt:
Something wrong goes with CI in this PR. `modules` target does not run, only `modules_install`.

I thought to look how it shows `$(MAKE) -C $(KERNEL) $(CFLAG_REMOVAL_OPTS) M=$(P_PWD) modules` but there is none. 

Compare this https://github.com/lkrg-org/lkrg/actions/runs/21385286215/job/61560447426
with the old run https://github.com/lkrg-org/lkrg/actions/runs/20686717343/job/59388521375

Makefile
kerneltoast: Yes, we shouldn't shadow variables used by Kbuild, so we need to change the CI setup. The CI setup shouldn't be specifying a kernel path at all.

Looking at our Makefile, I see there is another change I should make to alter this logic:
```
ifneq ($(KERNELRELEASE),)
    KERNEL := /lib/modules/$(KERNELRELEASE)/build
else
    ## KERNELRELEASE not set.
    KERNEL := /lib/modules/$(P_KVER)/build
endif
```

kerneltoast:
@solardiz The CI passes now. One of the CI checks appears to be stuck downloading some build dependencies to install via apt, FYI.

vt-alt:
> I thought to look how it shows `$(MAKE) -C $(KERNEL) $(CFLAG_REMOVAL_OPTS) M=$(P_PWD) modules` but there is none.

The line looks horrible long now. And it will _silently miscompile_ the module if compiled manually with `make -C /lib/modules/$(uname -r)/build M=$PWD modules` like some people do (for example ALT have this command in spec). Why pass as arguments and not just define the options inside of Makefile?

vt-alt:
BTW, how to ensure that thing is actually working? I see something that looks like instrumentation (in `main` branch currently at b8b1418a6c1e7229cdf3dfa020fcc4945e108d83):

`p_verify_addr_limit` is defined with `notrace`:
```c
notrace void p_verify_addr_limit(struct p_ed_process *p_orig) {
```
```
$ make 
$ objdump -d lkrg.ko | less +/p_verify_addr_limit
0000000000010800 <__pfx_p_verify_addr_limit>:
   10800:       90                      nop
   10801:       90                      nop
   10802:       90                      nop
   10803:       90                      nop
   10804:       90                      nop
   10805:       90                      nop
   10806:       90                      nop
   10807:       90                      nop
   10808:       90                      nop
   10809:       90                      nop
   1080a:       90                      nop
   1080b:       90                      nop
   1080c:       90                      nop
   1080d:       90                      nop
   1080e:       90                      nop
   1080f:       90                      nop

0000000000010810 <p_verify_addr_limit>:
   10810:       e9 00 00 00 00          jmp    10815 <p_verify_addr_limit+0x5>
   10815:       66 66 2e 0f 1f 84 00 00 00 00 00        data16 cs nopw 0x0(%rax,%rax,1)
```
16 nops looks like patchable-function-entry, so the `notrace` attribute did not work for this function before?

Btw, when I compile with `V=1` it is not just `-fpatchable-function-entry` but `-fpatchable-function-entry=16,16`.

vt-alt:
Btw, there is `ccflags-remove-y` since Linux v5.9, with it would have been just `ccflags-remove-y = $(CC_FLAGS_FTRACE)`. (Not filtering `-fpatchable-function-entry` though, but maybe we don't need to remove it?)

The fallback (similar to [what used before](https://github.com/torvalds/linux/commit/15d5761ad31dfb194ebe76554e6af0437eb20424) its introduction) would be:

```
ORIG_CFLAGS := $(KBUILD_CFLAGS)
KBUILD_CFLAGS = $(filter-out $(CC_FLAGS_FTRACE), $(ORIG_CFLAGS))
```
People used `subst` instead of `filter-out` because they (`CC_FLAGS_FTRACE` options) are added in a chunk anyway.

kerneltoast:
> Btw, there is `ccflags-remove-y` since Linux v5.9, with it would have been just `ccflags-remove-y = $(CC_FLAGS_FTRACE)`. (Not filtering `-fpatchable-function-entry` though, but maybe we don't need to remove it?)

Yes, see our conversation in #21 where it was decided `ccflags-remove-y` wasn't useful since LKRG maintains support as far back as 3.10. Same issue with CC_FLAGS_FTRACE; it's not present in 3.10.

> The fallback (similar to [what used before](https://github.com/torvalds/linux/commit/15d5761ad31dfb194ebe76554e6af0437eb20424) its introduction) would be:
> 
> ```
> ORIG_CFLAGS := $(KBUILD_CFLAGS)
> KBUILD_CFLAGS = $(filter-out $(CC_FLAGS_FTRACE), $(ORIG_CFLAGS))
> ```
> 
> People used `subst` instead of `filter-out` because they (`CC_FLAGS_FTRACE` options) are added in a chunk anyway.

Oh, I didn't expect modifying KBUILD_CFLAGS like that to work, but it does since LKRG is just a module instead of something like arch code, so KBUILD_CFLAGS is already populated with the ftrace flags by the time make descends into LKRG. Thanks for the suggestion, I will go with this approach.

> > I thought to look how it shows `$(MAKE) -C $(KERNEL) $(CFLAG_REMOVAL_OPTS) M=$(P_PWD) modules` but there is none.
> 
> The line looks horrible long now. And it will _silently miscompile_ the module if compiled manually with `make -C /lib/modules/$(uname -r)/build M=$PWD modules` like some people do (for example ALT have this command in spec). Why pass as arguments and not just define the options inside of Makefile?

~There isn't a `modules` recipe in LKRG's Makefile, and the KERNELRELEASE trick is a very old Kbuild trick that just lets you identify which pass over the Makefile is occurring, since there are two passes over the Makefile.~

EDIT: Ah, I just realized what you mean. You're right, nevermind.

The reason I passed CFLAGS_REMOVE as arguments was to dynamically populate a CFLAGS_REMOVE for every object file, instead of hardcoding CFLAGS_REMOVE for every object. 

> BTW, how to ensure that thing is actually working? I see something that looks like instrumentation (in `main` branch currently at [b8b1418](https://github.com/lkrg-org/lkrg/commit/b8b1418a6c1e7229cdf3dfa020fcc4945e108d83)):
> 
> `p_verify_addr_limit` is defined with `notrace`:
> 
> ```c
> notrace void p_verify_addr_limit(struct p_ed_process *p_orig) {
> ```
> 
> ```
> $ make 
> $ objdump -d lkrg.ko | less +/p_verify_addr_limit
> 0000000000010800 <__pfx_p_verify_addr_limit>:
>    10800:       90                      nop
>    10801:       90                      nop
>    10802:       90                      nop
>    10803:       90                      nop
>    10804:       90                      nop
>    10805:       90                      nop
>    10806:       90                      nop
>    10807:       90                      nop
>    10808:       90                      nop
>    10809:       90                      nop
>    1080a:       90                      nop
>    1080b:       90                      nop
>    1080c:       90                      nop
>    1080d:       90                      nop
>    1080e:       90                      nop
>    1080f:       90                      nop
> 
> 0000000000010810 <p_verify_addr_limit>:
>    10810:       e9 00 00 00 00          jmp    10815 <p_verify_addr_limit+0x5>
>    10815:       66 66 2e 0f 1f 84 00 00 00 00 00        data16 cs nopw 0x0(%rax,%rax,1)
> ```
> 
> 16 nops looks like patchable-function-entry, so the `notrace` attribute did not work for this function before?
> 
> Btw, when I compile with `V=1` it is not just `-fpatchable-function-entry` but `-fpatchable-function-entry=16,16`.

Good catch, I was mostly looking for -pg getting filtered out. I can fix this properly now that I know we can just directly filter out flags from KBUILD_CFLAGS.

Also, `notrace` is defined differently depending on the arch's options. See include/linux/compiler_types.h:
```
#if defined(CC_USING_HOTPATCH)
#define notrace			__attribute__((hotpatch(0, 0)))
#elif defined(CC_USING_PATCHABLE_FUNCTION_ENTRY)
#define notrace			__attribute__((patchable_function_entry(0, 0)))
#else
#define notrace			__attribute__((__no_instrument_function__))
#endif
```

The presence of patchable-function-entry shouldn't matter on arches that lack CC_USING_PATCHABLE_FUNCTION_ENTRY, because those arches don't expect to use a patchable function entry for ftrace. So `notrace` probably works fine because it will disable the ftrace mechanism actually utilized by the arch.

For LKRG, we'll disable all flags related to function patching just to be safe anyway.

kerneltoast:
The implementation is now cleaner and more robust after @vt-alt's feedback. I also dug through GCC sources to find all flags that could cause profiling code to be emitted, just to be safe, even though some of them aren't currently used by the upstream kernel.

There aren't any patchable function entry nops now:
```
$ objdump -d lkrg.ko | less +/p_verify_addr_limit
000000000000f280 <p_verify_addr_limit>:
    f280:	e9 00 00 00 00       	jmpq   f285 <p_verify_addr_limit+0x5>
    f285:	66 66 2e 0f 1f 84 00 	data16 nopw %cs:0x0(%rax,%rax,1)
    f28c:	00 00 00 00 
```

scripts/copy-builtin.sh
```diff
@@ -39,10 +39,13 @@ MAKEFILE=$(cat <<EOC
...
 
 obj-\$(CONFIG_SECURITY_LKRG) := lkrg.o
-ifeq (\$(SECURITY_LKRG_DEBUG), on)
+ifneq (\$(CONFIG_SECURITY_LKRG_DEBUG),)
 ccflags-m := -ggdb -DP_LKRG_DEBUG_BUILD -finstrument-functions
 ccflags-y := \${ccflags-m}
 lkrg-objs += modules/print_log/p_lkrg_debug_log.o
+else
+$(grep -Pzo 'REMOVE_FLAGS := [\S\s]*?\n(?=\S)' "$BASEDIR/../Makefile")
+KBUILD_CFLAGS := \$(filter-out \$(REMOVE_FLAGS),\$(KBUILD_CFLAGS))
```
solardiz, line 48: How does the `grep -Pzo` usage work here? `-z` would produce NUL bytes, right? But where do we expect a NUL-separated list here?

Makefile
```diff
@@ -23,6 +23,21 @@ ifeq ($(DEBUG), on)
 ccflags-m := -ggdb -DP_LKRG_DEBUG_BUILD -finstrument-functions
 ccflags-y := ${ccflags-m}
 $(TARGET)-objs += src/modules/print_log/p_lkrg_debug_log.o
+else
+# Remove trace-related flags so ftrace cannot probe LKRG's functions. This is
+# equivalent to specifying 'notrace' on every function.
```
solardiz: Maybe "similar" rather than exactly "equivalent"? We remove so many different flags that we probably sometimes do more than `notrace` did?

solardiz (commented):
Thank you very much @vt-alt and @kerneltoast! I only have a couple of nitpick comments/questions now. I mostly rely on your testing, and on our CI.

scripts/copy-builtin.sh
kerneltoast: `-z` changes the delimiter from `'\n'` to `'\0'`. This allows the grep regex to do multi-line matching.

Makefile
kerneltoast: True. I'll adjust it.

scripts/copy-builtin.sh
solardiz: Thanks. So we use `-z` here not to actually use NULs, but to remove the special meaning of `\n`. Makes sense now.

Adam-pi3:
Thanks for that PR @kerneltoast and valuable input @vt-alt and @solardiz. Maybe we should consider adding Singularity to the standard functinoalit-tests to make sure we always break/detect this tool. At least now, we can do it manually (and maybe we should start automating those)

vt-alt:
> ```
> ORIG_CFLAGS := $(KBUILD_CFLAGS)
> KBUILD_CFLAGS = $(filter-out $(CC_FLAGS_FTRACE), $(ORIG_CFLAGS))
> ```

Ah yeah, with `:=` we don't even need to introduce `ORIG_CFLAGS` to workaround recursive variable error. Thanks.

Issue #455, opened 2026-01-30
Singularity rootkit


solardiz:
Singularity is a new Open Source rootkit by @MatheuZSecurity (thank you, Matheu!), which has recently added an LKRG bypass. This "issue" is to formulate our stance on the topic, including how it fits our threat model (or does not), and what we're doing about it.

1. LKRG bypasses from kernel vulnerability exploits are more relevant for us and for LKRG users than bypasses from rootkits, since in that case the attacker doesn't yet have legitimate-looking root privileges at the initial stage. This is something to recognize and defend against with our current threat model. But in the case of rootkits, there isn't yet:
2. Currently, nothing prevents legitimate root from simply unloading LKRG with `rmmod` or effectively disabling it via `sysctl` before loading the rootkit. Therefore, an LKM rootkit bypassing LKRG, and us avoiding or breaking such bypass, will only make practical sense if and when we add protection against unloading LKRG and modifying its configuration #439 @kcdev809. And even then, the question will be about protecting the entire system boot chain, including userland; otherwise, the rootkit could be loaded by modifying files used during the next boot-up. That is, Secure Boot, signatures on everything before LKRG is loaded and on LKRG itself (implemented in RLC-H from CIQ), and kernel `lockdown`. In such system configuration, an LKM rootkit that lacks a valid signature could not be loaded without exploiting a vulnerability even in absence of LKRG. So what LKRG could add on top is mitigation of kernel vulnerability exploits - the same thing that's already our focus - and then post-detection of the rootkit's dirty hacks if it does get loaded anyway.
3. LKRG monitors the integrity and behavior of the kernel, which often detects rootkits - for example, 8 out of 9 in one study. However, we intentionally allow kernel modules loaded by legitimate root user (not with root access just obtained through a kernel vulnerability) and using only official APIs. We even have additional code that intentionally allows kernel function hooking using `ftrace` (which Singularity uses). We try not to break useful functionality. Therefore, writing a rootkit that we don't detect can be done simply by making it as correct as possible, without hacks (perhaps giving up on hiding). I think this is why 1 out of 9 wasn't detected in that old study - it was just a keylogger, not a full-fledged rootkit, which was more likely implemented without hacks.
4. Nevertheless, Singularity (or some of its behavior) was apparently initially detected by LKRG. Hence the additional bypass code, added in a more recent commit after the project's initial announcement. In a sense, it's a success for LKRG that a rootkit written without regard for LKRG was detected, even if this did not really matter yet.
[...]

PR #459, opened 2026-03-15
Verify newly loaded modules are persisted in the database

Adam-pi3:
### Description
Detect cases where a malicious module hides itself during init()
by checking whether the notifier routine adds it to the database.
Part of #455.

### How Has This Been Tested?
Against Singularity module

### DIFF
--- a/src/modules/kmod/p_kmod_notifier.c
+++ b/src/modules/kmod/p_kmod_notifier.c
@@ -29,6 +29,11 @@ static struct notifier_block p_module_block_notifier = {
 
 };
 
+static const char * const p_mod_strings[] = {
+                          "New module is LIVE",
+                          "New module is COMING",
+                          "Module is GOING AWAY",
+                          "New module is UNFORMED yet" };
 
 static int p_block_always(void) {
 
@@ -47,6 +52,46 @@ static void p_module_notifier_wrapper(unsigned long p_event, struct module *p_km
 
    return;
 }
+
+static inline void p_verify_added_module(struct module *p_kmod) {
+
+   unsigned int p_tmp;
+   unsigned int p_found_mod = 0;
+   unsigned int p_found_obj = 0;
+
+   for (p_tmp = 0; p_tmp < p_db.p_module_list_nr; p_tmp++)
+      if (p_db.p_module_list_array[p_tmp].p_mod == p_kmod)
+         p_found_mod = 1;
+
+   for (p_tmp = 0; p_tmp < p_db.p_module_kobj_nr; p_tmp++)
+      if (p_db.p_module_kobj_array[p_tmp].p_mod == p_kmod)
+         p_found_obj = 1;
+
+   if (!p_found_mod || !p_found_obj) {
+      p_print_log(P_LOG_ALERT, "DETECT: Newly loaded module entering LIVE state is hidden!");
+      /* Singularity rootkit is hooking print function
+       * and removes any messages containing "LKRG" string.
+       * Because of that, we are using here raw printk()
+       */
+      printk(KERN_CRIT "Hidden module:\n");
+      printk(KERN_CRIT "\tPointer to struct module: [0x%lx]\n", (unsigned long)p_kmod);
+      printk(KERN_CRIT "\tName: [%s]\n", p_kmod->name);
+      printk(KERN_CRIT "\tState: [%d: %s]\n", p_kmod->state,
+                                              (p_kmod->state > MODULE_STATE_UNFORMED) ?
+                                              "WRONG STATE VALUE" :
+                                              p_mod_strings[p_kmod->state]);
+      /* We can do more here, e.g.,
+       *  - dump the entire module
+       *  - calculate hash from .text and behave like AV?
+       *  - more...
+       */
+      if (P_CTRL(p_kint_enforce) >= 2) {
+         // OK, we need to crash the kernel now
+         p_panic("Kernel: Newly loaded module [name: %s] in LIVE state is hidden!", p_kmod->name);
+      }
+   }
+}
+
 
 #if P_OVL_OVERRIDE_SYNC_MODE
 static void p_verify_module_live(struct module *p_mod);
@@ -71,12 +116,6 @@ static void p_verify_module_going(struct module *p_mod);
 static int p_module_event_notifier(struct notifier_block *p_this, unsigned long p_event, void *p_kmod) {
 
    struct module *p_tmp = p_kmod;
-
-   static const char * const p_mod_strings[] = {
-                             "New module is LIVE",
-                             "New module is COMING",
-                             "Module is GOING AWAY",
-                             "New module is UNFORMED yet" };
 
 // STRONG_DEBUG
    p_debug_log(P_LOG_FLOOD,
@@ -219,6 +258,8 @@ static int p_module_event_notifier(struct notifier_block *p_this, unsigned long 
          while(p_kmod_hash(&p_db.p_module_list_nr,&p_db.p_module_list_array,
                            &p_db.p_module_kobj_nr,&p_db.p_module_kobj_array, 2) != P_LKRG_SUCCESS)
             schedule();
+
+         p_verify_added_module(p_module_activity_ptr);
 
          /* Update global module list/kobj hash */
          p_db.p_module_list_hash = p_lkrg_fast_hash((unsigned char *)p_db.p_module_list_array,
--- a/src/modules/integrity_timer/p_integrity_timer.c
+++ b/src/modules/integrity_timer/p_integrity_timer.c
@@ -963,7 +963,7 @@ void p_check_integrity(struct work_struct *p_work) {
 
    if (p_hack_check) {
       p_print_log(P_LOG_ALERT, "DETECT: Kernel: %u checksums changed unexpectedly", p_hack_check);
-      if (P_CTRL(p_kint_enforce >= 2)) {
+      if (P_CTRL(p_kint_enforce) >= 2) {
          // OK, we need to crash the kernel now
          p_panic("Kernel: %u checksums changed unexpectedly", p_hack_check);
       }


Adam-pi3:
Look like I used corrupted branch. If you fine with the logic, I will directly push it to the repo instead of creating a new PR.

src/modules/kmod/p_kmod_notifier.c
```diff
@@ -48,6 +48,51 @@ static void p_module_notifier_wrapper(unsigned long p_event, struct module *p_km
...
+static inline void p_verify_added_module(struct module *p_kmod) {
+
+   unsigned int p_tmp;
+   unsigned int p_found_mod = 0;
+   unsigned int p_found_obj = 0;
+   static const char * const p_mod_strings[] = {
+                             "New module is LIVE",
+                             "New module is COMING",
+                             "Module is GOING AWAY",
+                             "New module is UNFORMED yet" };
```
solardiz: Should make `p_mod_strings` shared between two functions that use it - move it to global scope in this source file (but `static` to keep it local to the file).

src/modules/kmod/p_kmod_notifier.c
```diff
@@ -48,6 +48,51 @@ static void p_module_notifier_wrapper(unsigned long p_event, struct module *p_km
...
+       * and removes any messages containing "LKRG" string.
+       * Because of that, we are using here raw printk()
+       */
+      printk(KERN_CRIT "Hidden module:\n");
+      printk(KERN_CRIT "\tPointer to struct module: [0x%lx]\n", (unsigned long)p_kmod);
+      printk(KERN_CRIT "\tName: [%s]\n", p_kmod->name);
+      printk(KERN_CRIT "\tState: [%d: %s]\n", p_kmod->state,
+                                              (p_kmod->state > MODULE_STATE_UNFORMED) ?
+                                              "WRONG STATE VALUE" :
+                                              p_mod_strings[p_kmod->state]);
```
solardiz, line 82: It's confusing that we first report "in LIVE state" and then may report a different state here. As I understand, the former is based on event and the latter is on the state field, but if we report both we need to adjust the wording to make this non-confusing.

src/modules/kmod/p_kmod_notifier.c
```diff
@@ -48,6 +48,51 @@ static void p_module_notifier_wrapper(unsigned long p_event, struct module *p_km
...
+      /* We can do more here, e.g.,
+       *  - dump the entire module
+       *  - calculate hash from .text and behave like AV?
+       *  - more...
+       */
+      if (P_CTRL(p_kint_enforce >= 2)) {
+         // OK, we need to crash the kernel now
+         p_panic("Kernel: Newly loaded module [name: %s] in LIVE state is hidden!", p_kmod->name);
```
solardiz, line 90: BTW, if we assume the module is malicious, then also its name may be malicious - may attempt to inject something deliberately misleading in these messages. OTOH, we have the same problem with task names already, and we just log them anyway.

solardiz (changes requested):
Please also clean up the commits here - should be just one commit, no merge commits.

solardiz:
> Look like I used corrupted branch. If you fine with the logic, I will directly push it to the repo instead of creating a new PR.

Rather than push directly or create new PR, can you rebase and force-push to this PR?

solardiz:
> ### How Has This Been Tested?
> 
> Against Singularity module

I assume you got kernel panic immediately upon loading Singularity on top of LKRG, correct?

Have you also tried it in the other order? If so, what happens in that case?

src/modules/kmod/p_kmod_notifier.c
```diff
@@ -48,6 +48,51 @@ static void p_module_notifier_wrapper(unsigned long p_event, struct module *p_km
...
+      printk(KERN_CRIT "\tState: [%d: %s]\n", p_kmod->state,
+                                              (p_kmod->state > MODULE_STATE_UNFORMED) ?
+                                              "WRONG STATE VALUE" :
+                                              p_mod_strings[p_kmod->state]);
+      /* We can do more here, e.g.,
+       *  - dump the entire module
+       *  - calculate hash from .text and behave like AV?
+       *  - more...
+       */
+      if (P_CTRL(p_kint_enforce >= 2)) {
```
solardiz: Misplaced closing brace - should be right after `p_kint_enforce`. I see we have the same bug in `src/modules/integrity_timer/p_integrity_timer.c`, so should fix it there as well. Luckily, the macro is currently defined such that this didn't matter.

src/modules/kmod/p_kmod_notifier.c
Adam-pi3: Correct, the latter one is what's in the module state (which sometimes is still visible as LIVE though). What wording would you suggest?

src/modules/kmod/p_kmod_notifier.c
Adam-pi3: I don't get, could you elaborate?

Adam-pi3:
> I assume you got kernel panic immediately upon loading Singularity on top of LKRG, correct?

Correct

> Have you also tried it in the other order? If so, what happens in that case?

I didn't but I do not expect anything spectacular in that scenario. Singularity is not visible in the system so LKRG during load-time will not see it. Our Threat Model has assumption of the system being clean during LKRG load time - otherwise it does not make any sense

> Rather than push directly or create new PR, can you rebase and force-push to this PR?

Sorry, I'm not a git master, could you baby step me here?

src/modules/kmod/p_kmod_notifier.c
solardiz: Should be `P_CTRL(p_kint_enforce) >= 2`

src/modules/kmod/p_kmod_notifier.c
solardiz: Maybe in the first message (about the event) replace "in" with "entering"

src/modules/kmod/p_kmod_notifier.c
Adam-pi3: Fixed

src/modules/kmod/p_kmod_notifier.c
Adam-pi3: Oh... good catch! I fixed in both places

src/modules/kmod/p_kmod_notifier.c
Adam-pi3: Make sense, fixed

Adam-pi3:
> > Rather than push directly or create new PR, can you rebase and force-push to this PR?
> 
> Sorry, I'm not a git master, could you baby step me here?

I think i just done that, so we should be fine?

Adam-pi3:
> I think i just done that, so we should be fine?

I think there are still 2 old commits, have no idea how to clean it. Any help @solardiz ?

solardiz:
> I think there are still 2 old commits, have no idea how to clean it. Any help @solardiz ?

I've just removed them. I may also squash the remaining 4 into fewer, perhaps 2.

solardiz (approved):

solardiz:
> I may also squash the remaining 4 into fewer, perhaps 2.

Done. I also made the `P_CTRL` usage bug fix first commit because we need to have/retain it even if we for some reason revert the new addition, and not to duplicate the bug into the newly added code before fixing it.

We may also need a `CHANGES` entry. And ideally a review by @kerneltoast.

src/modules/kmod/p_kmod_notifier.c
```diff
@@ -48,6 +53,46 @@ static void p_module_notifier_wrapper(unsigned long p_event, struct module *p_km
...
+         p_found_obj = 1;
+
+   if (!p_found_mod || !p_found_obj) {
+      p_print_log(P_LOG_ALERT, "DETECT: Newly loaded module entering LIVE state is hidden!");
+      /* Singularity rootkit is hooking print function
+       * and removes any messages containing "LKRG" string.
+       * Because of that, we are using here raw printk()
+       */
+      printk(KERN_CRIT "Hidden module:\n");
+      printk(KERN_CRIT "\tPointer to struct module: [0x%lx]\n", (unsigned long)p_kmod);
```
solardiz, line 77: BTW, we don't normally log kernel pointers at default log level as they can be useful for exploits. However, in this case we're about to panic the kernel, so I suppose this does not matter.

solardiz:
> I think there are still 2 old commits, have no idea how to clean it. Any help @solardiz ?

For further occasions, you can use interactive rebase to do this, starting with `git rebase -i 2ad465c693e334dbcf49bf6f1188b90e11bcb4c6`. Removing those 2 commits was as simple as deleting 2 lines there.

MatheuZSecurity:
👀👀

solardiz:
> 👀👀 Waiting for the commit so I can release a patch

Yes, please @MatheuZSecurity. :-)

solardiz:
@Adam-pi3 A colleague at CIQ has just reported that this produces false positives, unfortunately. We'll need to investigate or revert.

solardiz:
This happens on Arch Linux boot-up with the `lkrg-dkms-git` package installed:

```
[    8.200588] LKRG: ALERT: DETECT: Newly loaded module entering LIVE state is hidden!
[    8.200992] Hidden module:
[    8.201298] 	Pointer to struct module: [0xffffffffc1caad80]
[    8.201564] 	Name: [ptp]
[    8.201814] 	State: [0: New module is LIVE]
[    8.202062] LKRG: ALERT: BLOCK: Kernel: Newly loaded module [name: ptp] in LIVE state is hidden!
[    8.202316] Kernel panic - not syncing: Kernel: Newly loaded module [name: ptp] in LIVE state is hidden!
[    8.202319] CPU: 17 UID: 0 PID: 826 Comm: (udev-worker) Tainted: G           OE       6.19.8-arch1-1 #1 PREEMPT(full)  72931f2c699f8285fbdd3bf1eafb70b6934b6d21
```

Also with another module on another boot.

src/modules/kmod/p_kmod_notifier.c
```diff
@@ -220,6 +259,8 @@ static int p_module_event_notifier(struct notifier_block *p_this, unsigned long
                            &p_db.p_module_kobj_nr,&p_db.p_module_kobj_array, 2) != P_LKRG_SUCCESS)
             schedule();
 
+         p_verify_added_module(p_module_activity_ptr);
```
solardiz, line 262: Maybe the local variable `p_tmp` would be more appropriate to use here than the global `p_module_activity_ptr`, but I don't see how this could lead to a problem either way because the mutex is taken before that global variable is updated and that mutex is still held here.

solardiz:
@Adam-pi3 In `src/modules/kmod/p_kmod.c` we're skipping modules where `!p_module_core(p_mod) || !p_core_text_size(p_mod)`, but here we expect them to be present in the list anyway. So if this condition is true for any module going live here, then we'll have a false positive like this. I doubt this condition would ever be true for a valid module, can it be? Regardless, this is a bug. We need to be consistent.

solardiz:
> I doubt this condition would ever be true for a valid module

Upon a second thought, I think it can be: for a module that only has an init function, but no other code. I actually wrote some of those myself in the past, such as to live-patch a bug.

Adam-pi3:
Interesting, let's try adding this logic and see if it helps. I will reply in the issue thread

solardiz:
Trying to load latest Singularity (commit 2b6df46e0d4dbd4fe3931e0eaa5c7e52f6fba12e "try harder") after the LKRG, the VM was locking up (I assume our kernel panic). After I disabled our enforcement, it's this:

```
[  288.955312] LKRG: ALIVE: Loading LKRG
[  288.955315] LKRG: ISSUE: System does not support SMEP, which won't be validated
[  288.955316] LKRG: ISSUE: System does not support SMAP, which won't be validated
[  288.956364] LKRG: ISSUE: register_k[ret]probe() for lkrg_dummy failed! [err=-22]
[  288.958497] Freezing user space processes
[  288.960113] Freezing user space processes completed (elapsed 0.001 seconds)
[  288.960118] OOM killer disabled.
[  289.010321] LKRG: ISSUE: register_k[ret]probe() for ovl_dentry_init_reval failed! [err=-2]
[  289.129441] LKRG: ALIVE: LKRG initialized successfully
[  289.143875] OOM killer enabled.
[  289.143878] Restarting tasks ... done.
[  335.652534] LKRG: STATE: Changing 'kint_enforce' from 2 (PANIC) to 1 (LOG ONLY (For SELinux and CR0.WP LOG & RESTORE))
[  351.493067] LKRG: ALERT: DETECT: Newly loaded module entering LIVE state is hidden!
[  351.493778] Hidden module:
[  351.494029] 	Pointer to struct module: [0xffffffffc0bc1400]
[  351.494813] 	State: [0: New module is LIVE]
```

So this seems to work as @Adam-pi3 intended it.

src/modules/kmod/p_kmod_notifier.c
solardiz: > BTW, we don't normally log kernel pointers at default log level as they can be useful for exploits. However, in this case we're about to panic the kernel, so I suppose this does not matter.

... but we may have disabled the panics, so it may matter a bit, after all.

Adam-pi3: I'm not sure we should care about such scenario. This message will appear only when something really bad happens...

PR #461, opened 2026-03-15
Try harder at killing tasks

solardiz:
### Description

When we SIGKILL a task on its integrity violation, that signal is merely left pending to take effect when the task (forcibly) re-enters the kernel or when it completes handling its current call into the kernel. This may not be good enough if a rootkit intercepts our SIGKILL or if the current syscall already does something sufficient for the attacker. So we try to do a bit more here, by redundantly forcing SIGKILL into the pending signals mask, overwriting the task's userspace address to -1, and settings some of its resource limits to 0.

### How Has This Been Tested?

Just that it builds and doesn't crash in normal usage. Need to test the actual task killing behavior.

### DIFF
--- a/src/modules/exploit_detection/p_exploit_detection.h
+++ b/src/modules/exploit_detection/p_exploit_detection.h
@@ -274,12 +274,12 @@ struct p_ed_global_variables {
 
 };
 
-static inline int p_kill_task_by_task(struct task_struct *p_task) {
+static inline void p_kill_task_by_task(struct task_struct *p_task) {
 
    p_print_log(P_LOG_ALERT, "BLOCK: Task: Killing pid %u, name %s",
                task_pid_nr(p_task), p_task->comm);
 
-   return send_sig_info(SIGKILL, SEND_SIG_PRIV, p_task);
+   send_sig_info(SIGKILL, SEND_SIG_PRIV, p_task);
 }
 
 #include "ed_task_tree.h"
@@ -374,7 +374,7 @@ static inline unsigned int p_is_ed_task(struct task_struct *p_task) {
    return p_task->mm && !is_global_init(p_task);
 }
 
-static inline int p_ed_kill_task(struct p_ed_process *p_source) {
+static inline void p_ed_kill_task(struct p_ed_process *p_source) {
 
    struct task_struct *p_task = p_source->p_ed_task.p_task;
 
@@ -401,11 +401,9 @@ static inline int p_ed_kill_task(struct p_ed_process *p_source) {
          break;
 
    }
-
-   return 0;
 }
 
-static inline int p_pcfi_kill_edp_or_task(struct p_ed_process *p_source, struct task_struct *p_task) {
+static inline void p_pcfi_kill_edp_or_task(struct p_ed_process *p_source, struct task_struct *p_task) {
 
    if (p_source)
       p_task = p_source->p_ed_task.p_task;
@@ -421,11 +419,12 @@ static inline int p_pcfi_kill_edp_or_task(struct p_ed_process *p_source, struct 
 
       /* Kill task */
       case 1:
-         if (!p_source)
-            return p_kill_task_by_task(p_task);
-
-         /* Zero kill_task_on_unlock to kill upon unlocking the ed task lock */
-         p_source->kill_task_on_unlock = 0;
+         if (p_source) {
+            /* Zero kill_task_on_unlock to kill upon unlocking the ed task lock */
+            p_source->kill_task_on_unlock = 0;
+         } else {
+            p_kill_task_by_task(p_task);
+         }
          break;
 
       /* Log */
@@ -435,13 +434,11 @@ static inline int p_pcfi_kill_edp_or_task(struct p_ed_process *p_source, struct 
          break;
 
    }
-
-   return 0;
 }
 
-static inline int p_pcfi_kill_ed_task(struct p_ed_process *p_source) {
+static inline void p_pcfi_kill_ed_task(struct p_ed_process *p_source) {
 
-   return p_pcfi_kill_edp_or_task(p_source, NULL);
+   p_pcfi_kill_edp_or_task(p_source, NULL);
 }
 
 /*
--- a/src/modules/exploit_detection/p_exploit_detection.h
+++ b/src/modules/exploit_detection/p_exploit_detection.h
@@ -363,12 +363,6 @@ void p_debug_off_flag_override_off(struct p_ed_process *p_source, unsigned int p
 void p_debug_off_flag_override_on(struct p_ed_process *p_source, unsigned int p_id, struct pt_regs *p_regs);
 #endif
 
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,3,0)
-#define p_force_sig(sig) force_sig((sig))
-#else
-#define p_force_sig(sig) force_sig((sig), current)
-#endif
-
 static inline unsigned int p_is_ed_task(struct task_struct *p_task) {
 
    return p_task->mm && !is_global_init(p_task);
--- a/src/modules/exploit_detection/p_exploit_detection.h
+++ b/src/modules/exploit_detection/p_exploit_detection.h
@@ -274,12 +274,16 @@ struct p_ed_global_variables {
 
 };
 
+static inline void p_crash_task_if_current(struct task_struct *p_task);
+
 static inline void p_kill_task_by_task(struct task_struct *p_task) {
 
    p_print_log(P_LOG_ALERT, "BLOCK: Task: Killing pid %u, name %s",
                task_pid_nr(p_task), p_task->comm);
 
    send_sig_info(SIGKILL, SEND_SIG_PRIV, p_task);
+
+   p_crash_task_if_current(p_task);
 }
 
 #include "ed_task_tree.h"
@@ -468,6 +472,44 @@ static inline unsigned int p_ed_pcfi_cpu(unsigned char p_kill) {
 
 #endif
 
+}
+
+#include "../wrap/p_struct_wrap.h"
+
+static inline void p_crash_task_if_current(struct task_struct *p_task) {
+
+   unsigned long flags;
+
+   /*
+    * Unfortunately, we cannot just take the sighand lock for another task
+    * without risking deadlock.  See __lock_task_sighand() in the kernel.
+    */
+   if (p_task != current)
+      return;
+
+   /*
+    * We assume we also send SIGKILL via send_sig_info(), but a rootkit may
+    * easily intercept that, so we redundantly enter SIGKILL into the pending
+    * signals mask directly.  For similar reasons, as well as to potentially
+    * limit what the currently running syscall will do, zero some rlimits -
+    * those known to be relevant yet safe to zero (unlike e.g. RLIMIT_FSIZE,
+    * which may be unsafe to zero in a privileged victim task).
+    */
+   spin_lock_irqsave(&p_task->sighand->siglock, flags);
+   p_task->signal->rlim[RLIMIT_NPROC].rlim_cur = p_task->signal->rlim[RLIMIT_NPROC].rlim_max = 0;
+   p_task->flags |= PF_NPROC_EXCEEDED; /* safe read-modify-write assuming p_task == current */
+   p_task->signal->rlim[RLIMIT_NOFILE].rlim_cur = p_task->signal->rlim[RLIMIT_NOFILE].rlim_max = 0;
+   p_task->signal->rlim[RLIMIT_AS].rlim_cur = p_task->signal->rlim[RLIMIT_AS].rlim_max = 0;
+   sigaddset(&p_task->pending.signal, SIGKILL);
+   spin_unlock_irqrestore(&p_task->sighand->siglock, flags);
+
+   /*
+    * The task shouldn't ever return to userspace due to pending SIGKILL, but
+    * if it does then let's (hopefully) crash it (although it may avoid that by
+    * trapping SIGSEGV).  This is just an added layer of protection in case the
+    * SIGKILL doesn't work for some (malicious) reason.
+    */
+   p_regs_set_ip(task_pt_regs(p_task), -1);
 }
 
 #endif
--- a/src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
+++ b/src/modules/exploit_detection/syscalls/p_call_usermodehelper/p_call_usermodehelper.c
@@ -102,7 +102,7 @@ p_call_usermodehelper_entry_not_allowed:
             /* Prevent execution */
             case 1:
                p_print_log(P_LOG_ALERT, "BLOCK: UMH: Executing program name %s", p_subproc->path);
-               p_force_sig(SIGKILL);
+               p_kill_task_by_task(current);
                break;
 
             /* Log only */


src/modules/exploit_detection/p_exploit_detection.h
```diff
@@ -421,11 +430,12 @@ static inline int p_pcfi_kill_edp_or_task(struct p_ed_process *p_source, struct
...
       /* Kill task */
       case 1:
-         if (!p_source)
-            return p_kill_task_by_task(p_task);
-
-         /* Zero kill_task_on_unlock to kill upon unlocking the ed task lock */
-         p_source->kill_task_on_unlock = 0;
+         if (!p_source) {
+            p_kill_task_by_task(p_task);
+         } else {
```
Adam-pi3: Previous logic didn't have "else" now we have it, is that correct?

solardiz: Yes. Previously we had `return`, so wouldn't reach what's now in `else` for that reason.

src/modules/exploit_detection/p_exploit_detection.h
```diff
@@ -274,12 +274,28 @@ struct p_ed_global_variables {
...
+   p_crash_task(p_task);
 
    p_print_log(P_LOG_ALERT, "BLOCK: Task: Killing pid %u, name %s",
                task_pid_nr(p_task), p_task->comm);
 
-   return send_sig_info(SIGKILL, SEND_SIG_PRIV, p_task);
+   send_sig_info(SIGKILL, SEND_SIG_PRIV, p_task);
+
+   /* Deliberately partially redundant with the above */
+   spin_lock_irqsave(&p_task->sighand->siglock, flags);
```
kerneltoast: This can deadlock with our seccomp hook because it can kill a task underneath the sighand lock. We can't use this lock here.

You can change this block of code to:
```c
memset(p_task->signal->rlim, 0, sizeof(p_task->signal->rlim));
while (unlikely(!sigismember(&p_task->pending.signal, SIGKILL)))
	sigaddset(&p_task->pending.signal, SIGKILL);
```

src/modules/exploit_detection/p_exploit_detection.h
```diff
@@ -421,11 +430,12 @@ static inline int p_pcfi_kill_edp_or_task(struct p_ed_process *p_source, struct
 
       /* Kill task */
       case 1:
-         if (!p_source)
-            return p_kill_task_by_task(p_task);
-
-         /* Zero kill_task_on_unlock to kill upon unlocking the ed task lock */
-         p_source->kill_task_on_unlock = 0;
+         if (!p_source) {
```
kerneltoast: Invert the if-statement to make it clearer, i.e. put `if (p_source)` first so we can immediately see how p_source is used when non-null.

kerneltoast (changes requested):

src/modules/exploit_detection/p_exploit_detection.h
solardiz: If this can deadlock, then doesn't the same apply to pre-existing call to `send_sig_info`, which I think acquires this same lock under the hood?

Yes, I also thought of this alternative way to write it, except with an `if` where you have a `while`, but we may. The thinking is that if something is so seriously wrong that `send_sig_info` didn't add the signal to the mask, we don't care about non-atomic read-modify-write possibly removing some other signal being added concurrently. If both of us arrived at this idea, then perhaps it's in fact worthwhile.

kerneltoast: > If this can deadlock, then doesn't the same apply to pre-existing call to send_sig_info, which I think acquires this same lock under the hood?

Ah, crap. Yeah, a deadlock can happen in the seccomp hook when the off flag for a thread in the group is corrupted, since they share the same sighand. So we could see a deadlock with this callchain:
```
p_handle_seccomp
  p_seccomp_filter_tsync(entry = false)
    spin_lock_irqsave(&current->sighand->siglock, flags) // sighand is locked
    ed_task_find_lock_rcu                                // ed lock is taken
    p_set_ed_process_on
      p_set_ed_process_override_on
        p_validate_off_flag
          p_ed_is_off_off
            p_ed_kill_task
              p_source->kill_task_on_unlock = 0          // armed to kill
    ed_task_unlock                                       // ed lock is released
      p_kill_task_by_task
        send_sig_info                                    // sighand still locked!!!
                                                         // DEADLOCK!!!
```
<br>

> Yes, I also thought of this alternative way to write it, except with an if where you have a while, but we may. The thinking is that if something is so seriously wrong that send_sig_info didn't add the signal to the mask, we don't care about non-atomic read-modify-write possibly removing some other signal being added concurrently. If both of us arrived at this idea, then perhaps it's in fact worthwhile.

Right. My thinking is that we'd want to make sure SIGKILL persisted over any other attempt to write into the signal mask, in case we weren't going to take the sighand lock to do it. And it is fine to trash any other signals in the signal mask because SIGKILL always preempts all other signals in the signal queue.

Adam-pi3: Interesting. However, it is a corner case for seccomp which maybe we can getaway using your idea of `while` loop. Another idea is we can verify if the lock is taken just not manually inject signal (skip that logic) but still manually crash the task and use offical signal sending API (like we do today)

solardiz: > Yeah, a deadlock can happen in the seccomp hook when the off flag for a thread in the group is corrupted, since they share the same sighand.

@kerneltoast Please open an issue for this. Thank you!

solardiz: @Adam-pi3 Yes, we could use logic similar to this PR's in case the sighand lock is already taken in that place. As to the `while` loop, I think the compiler will optimize it to `if` and that's fine. Even with a true `while` loop, an attacker constantly rewriting the value would succeed in doing so right after the loop finally ends, so it's not effective against that, nor against anything else.

kerneltoast: @solardiz If the sighand lock is primarily meant to protect the `sigaddset()`, then the `while` loop can be used in place of that. My intention with the loop was to effectively provide the same protection as the lock for the `sigaddset()`.

OTOH, if the lock is meant to protect the rlim modifications, then it's a different story. Are there some bugs you predicted would occur without the sighand lock?

@Adam-pi3 I don't think the compiler would optimize the `while` loop away because there is pointer chasing involved here.

solardiz: @kerneltoast I thought of the lock for both the signal set and the rlimits, as it's documented to also be used to protect the latter. I'm not aware of specific "bugs [that] would occur without the sighand lock", I acquired it just for correctness and to avoid discussion of what would happen without it - but we're having just this discussion now.

solardiz: @kerneltoast Regarding rlimits, when we zeroize them without locking and with not necessarily strictly sequential `memset` (or not strictly in-order propagation to another CPU if crossing a cache line on an arch with weak memory ordering) we may briefly make rlim_cur > rlim_max as seen from another context. I hope no readers are confused in a dangerous way by this.

Also regarding `memset` vs. resetting of individual limits, I wonder if any could possibly be unsafe to reset. That would be really weird, but we've seen that resetting of capabilities can be unsafe when it makes a program fail to drop (other) privileges then (IIRC, as first demonstrated against `sendmail` circa 2000). Which is part of why I don't do this here, even though the program is expected to die right after current syscall, so this could actually be safe.

solardiz:
@kerneltoast I've made the changes you suggested (also added your Co-authored-by to one commit) and force-pushed. Please take another look.

src/modules/exploit_detection/p_exploit_detection.h
solardiz: > I don't think the compiler would optimize the `while` loop away because there is pointer chasing involved here.

I confirm that (in my local test build) the generated code is slightly different for `while` vs. `if`. I didn't analyze exactly in what way, but it does contain one extra branch instruction in `p_exploit_detection.o` with the `while`.

solardiz: > My intention with the loop was to effectively provide the same protection as the lock for the `sigaddset()`.

OK, understood - so in case a concurrent non-malicious change (that does use the lock) happens and wipes our locklessly set SIGKILL due to it being read-modify-write and us not locking against it (so its read step could have been before our write), we reintroduce SIGKILL (and possibly wipe that other change).

kerneltoast:
> @kerneltoast I've made the changes you suggested (also added your Co-authored-by to one commit) and force-pushed. Please take another look.

@solardiz I've thought about it more and what we're dealing with here is unfortunately trickier than it seems. Because we may kill tasks outside of the context of the task to be killed (seccomp hook, integrity timer, and paranoid mode), that means the task could be actively running _anywhere_ (anywhere in the kernel or userspace) when `p_kill_task_by_task()` is called to kill it.

There is a lot we need to consider and verify with these changes, specifically the rlim zeroing and instruction pointer corruption.

The comment in `p_crash_task()`, `/* Crash the task if it returns to userspace */`, is misleading: the task might currently be in userspace!

I suspect that at the very least we may need to add back the sighand lock. And there was another problem with that, as it turns out: you may get either a NULL pointer dereference or use-after-free attempting to lock `p_task->sighand->siglock`. NULL pointer dereference because the `p_task->sighand` pointer is set to NULL in `__exit_signal()`, and use-after-free _when the task doing the killing isn't in the same thread group_ because the sighand pointer may have already been freed. That's why the kernel has the (non-exported) [__lock_task_sighand()](https://elixir.bootlin.com/linux/v6.19.10/source/kernel/signal.c#L1358-L1388) function, because of these races with task exit.

When the task doing the killing _is_ in the same thread group as the task getting killed, the use-after-free issue disappears because the sighand is still in use by the killer. But the NULL pointer dereference issue would still remain.

I was aware of these potential pitfalls with the sighand lock in the past but failed to mention them anywhere. The trick I did in the seccomp hook with `spin_lock_irqsave(&current->sighand->siglock, flags);` was deliberately careful: `current->sighand` will always be stable in the context of the seccomp syscall, hence it is safe to just lock `current->sighand->siglock` directly like that. The same applies to my `spin_lock_irqsave(&current->sighand->siglock, flags);` in `p_insert_ed_task()`: the new task getting inserted is guaranteed to have a non-NULL sighand pointer and it's guaranteed that the task won't go and exit while our hook runs from the context of that task itself.

solardiz:
> the task could be actively running _anywhere_ (anywhere in the kernel or userspace) when `p_kill_task_by_task()` is called to kill it.

Yes, I had this in mind.

> There is a lot we need to consider and verify with these changes, specifically the rlim zeroing and instruction pointer corruption.

Yes, it's in this context that I mentioned we may be violating the cur<=max invariant for some readers.

> The comment in `p_crash_task()`, `/* Crash the task if it returns to userspace */`, is misleading: the task might currently be in userspace!

Yes. My thinking was that our -1 would likely be overwritten when the task enters the kernel, then - so we would not crash such task, at least not in this way.

> I suspect that at the very least we may need to add back the sighand lock.

To achieve what exactly? Protection of rlim?

> And there was another problem with that, as it turns out: you may get either a NULL pointer dereference or use-after-free attempting to lock `p_task->sighand->siglock`. NULL pointer dereference because the `p_task->sighand` pointer is set to NULL in `__exit_signal()`, and use-after-free _when the task doing the killing isn't in the same thread group_ because the sighand pointer may have already been freed. That's why the kernel has the (non-exported) [__lock_task_sighand()](https://elixir.bootlin.com/linux/v6.19.10/source/kernel/signal.c#L1358-L1388) function, because of these races with task exit.

Oh. Non-exported, but also non-static and used from other source files, and simple enough that we could duplicate it.

I see we only have two other places where we take the sighand lock, and you've just commented on them here and in #467 that both of them are for `current`, so unaffected by the above risks.

I see these options:

1. Continue to work lockless and take some risks - we only reach this when something already went wrong (attack).
2. Use the kernel's `__lock_task_sighand`.
3. Use our own equivalent of `__lock_task_sighand`.
4. Limit the extra task killing logic introduced here to `current` task. Don't try this hard at killing other tasks (it's very rare that we'd try killing them anyway - like you say, only with seccomp or with non-default paranoid mode).

Sounds right? Any preference?

solardiz:
@kerneltoast We may also consider your plans at addressing #466 when deciding what to do here. It'd be silly to spend effort on a certain solution that you'd end up replacing.

kerneltoast:
> To achieve what exactly? Protection of rlim?

My concern is that we could end up triggering a crash in the kernel somewhere if we break the upstream's assumptions about values inside of task_struct.

> I see these options:
>
> 1. Continue to work lockless and take some risks - we only reach this when something already went wrong (attack).
> 2. Use the kernel's `__lock_task_sighand`.
> 3. Use our own equivalent of `__lock_task_sighand`.
> 4. Limit the extra task killing logic introduced here to `current` task. Don't try this hard at killing other tasks (it's very rare that we'd try killing them anyway - like you say, only with seccomp or with non-default paranoid mode).
>
> Sounds right? Any preference?

The integrity timer is another mechanism that may kill another task. You've pointed this out to me before when I thought it was only seccomp and paranoid mode that could kill non-`current` tasks. 😉 

Option 4 came to my mind earlier today. Let's do that.

> @kerneltoast We may also consider your plans at addressing #466 when deciding what to do here. It'd be silly to spend effort on a certain solution that you'd end up replacing.

@solardiz I just finished coding up a fix to #466 about an hour ago. I still need to artificially test it (hence no PR just yet), but the approach is that the actual killing of tasks in the seccomp hook will be moved outside of the sighand lock. So I am making it safe to keep using the sighand lock in `p_kill_task_by_task()`.

solardiz:
> The integrity timer is another mechanism that may kill another task. You've pointed this out to me before when I thought it was only seccomp and paranoid mode that could kill non-`current` tasks. 😉

Ah, sure.

> Option 4 came to my mind earlier today. Let's do that.

OK. So I add checks for `p_task == current` and re-add acquisition of sighand lock, right?

kerneltoast:
> > Option 4 came to my mind earlier today. Let's do that.
> 
> OK. So I add checks for `p_task == current` and re-add acquisition of sighand lock, right?

Yep, sounds good.

solardiz:
@kerneltoast I think I'll also revert the `memset` to my zeroing of just a few rlimits that I know are relevant and reasonably safe to zero assuming that the task may be a victim under attack. I am not so sure about some other rlimits, like zeroing of RLIMIT_FSIZE may result in truncation of a file being written by a privileged program (there were such CVEs before).

solardiz:
> > OK. So I add checks for `p_task == current` and re-add acquisition of sighand lock, right?
> 
> Yep, sounds good.

I've made the changes - please re-review. Thank you!

src/modules/exploit_detection/p_exploit_detection.h
```diff
@@ -383,6 +400,7 @@ static inline void p_ed_kill_task(struct p_ed_process *p_source) {
 
       /* Kill task */
       case 1:
+         p_crash_task_if_current(p_task);
```
kerneltoast: This `p_crash_task_if_current()` call here is redundant because the current task is running within the kernel. There is no chance of the current task returning to userspace before `p_kill_task_by_task()` runs. The `p_crash_task_if_current()` call in `p_kill_task_by_task()` is sufficient.

src/modules/exploit_detection/p_exploit_detection.h
```diff
@@ -274,12 +274,29 @@ struct p_ed_global_variables {
 
 };
 
+static inline void p_crash_task_if_current(struct task_struct *p_task);
+
 static inline void p_kill_task_by_task(struct task_struct *p_task) {
 
+   p_crash_task_if_current(p_task);
+
```
kerneltoast: Why is `p_crash_task_if_current()` called up here _and_ down below? I think this call should be removed.

src/modules/exploit_detection/p_exploit_detection.h
```diff
@@ -274,12 +274,29 @@ struct p_ed_global_variables {
...
+   if (p_task == current) {
+      /* Deliberately partially redundant with the above */
+      unsigned long flags;
+      spin_lock_irqsave(&p_task->sighand->siglock, flags);
+      p_task->signal->rlim[RLIMIT_NPROC].rlim_cur = p_task->signal->rlim[RLIMIT_NPROC].rlim_max = 0;
+      p_task->signal->rlim[RLIMIT_NOFILE].rlim_cur = p_task->signal->rlim[RLIMIT_NOFILE].rlim_max = 0;
+      p_task->signal->rlim[RLIMIT_AS].rlim_cur = p_task->signal->rlim[RLIMIT_AS].rlim_max = 0;
+      sigaddset(&p_task->pending.signal, SIGKILL);
+      spin_unlock_irqrestore(&p_task->sighand->siglock, flags);
+   }
```
kerneltoast: Move all of this into `p_crash_task_if_current()`. When you do, make `p_crash_task_if_current()` do an early return for `if (p_task != current)` to reduce the indentation by one level.

src/modules/exploit_detection/p_exploit_detection.h
```diff
@@ -274,12 +274,29 @@ struct p_ed_global_variables {
 
 };
 
+static inline void p_crash_task_if_current(struct task_struct *p_task);
```
kerneltoast, line 277: Why not move `#include "../wrap/p_struct_wrap.h"` up instead of creating this forward declaration? This makes me suspicious that there is some side-effect from including `p_struct_wrap.h` that would affect the code between here and where it is currently included at the bottom of this file (e.g., some kind of macro gore). There shouldn't be any issue like that; if there is, then it needs to be clearly documented.

src/modules/exploit_detection/p_exploit_detection.h
```diff
@@ -274,12 +274,29 @@ struct p_ed_global_variables {
...
 
+   p_crash_task_if_current(p_task);
+
    p_print_log(P_LOG_ALERT, "BLOCK: Task: Killing pid %u, name %s",
                task_pid_nr(p_task), p_task->comm);
 
    send_sig_info(SIGKILL, SEND_SIG_PRIV, p_task);
+
+   if (p_task == current) {
+      /* Deliberately partially redundant with the above */
```
kerneltoast: Please comment on this more rather than relying on links in the commit message to convey your intent. And clarify why the rlimits are zeroed as well as the justification for zeroing those specific rlimits (and possibly not other rlimits in addition).

src/modules/exploit_detection/p_exploit_detection.h
```diff
@@ -470,4 +488,14 @@ static inline unsigned int p_ed_pcfi_cpu(unsigned char p_kill) {
 
 }
 
+#include "../wrap/p_struct_wrap.h"
+
+static inline void p_crash_task_if_current(struct task_struct *p_task) {
+
+   if (p_task == current) {
+      /* Crash the task if it returns to userspace */
```
kerneltoast: Clarify that the task shouldn't ever return to userspace due to the SIGKILL, and that this is just an added layer of protection in case the SIGKILL doesn't work for some (malicious) reason.

kerneltoast (changes requested):

src/modules/exploit_detection/p_exploit_detection.h
```diff
@@ -470,4 +488,14 @@ static inline unsigned int p_ed_pcfi_cpu(unsigned char p_kill) {
...
+      p_regs_set_ip(task_pt_regs(p_task), -1);
```
kerneltoast: Setting the instruction pointer to -1 is less useful than it seems on the surface. The reason is because the program will fault and the kernel's fault handler will send SIGSEGV to the task. And a malicious program can simply catch SIGSEGV. Such a program could even partially restore the corrupted instruction pointer by inspecting the current stack frame and executing the instruction before the current return address. Or it can expect the SIGSEGV to be the last step for one of its functions and then simply execute `ret` to be on its merry way.

src/modules/exploit_detection/p_exploit_detection.h
solardiz: As I recall, I did in fact run into issues trying to avoid the forward declaration in various ways, so having it ended up the least invasive change.

src/modules/exploit_detection/p_exploit_detection.h
solardiz: My thinking was that a rootkit may do something so drastic in its intercepted `send_sig_info` that we don't get control back to here. In that case, we'd at least have already damaged the task a little bit.

src/modules/exploit_detection/p_exploit_detection.h
solardiz: Same reason as the redundant call in there - if we do it early, then it's already done and so unaffected by a rootkit stealing control in one of our later steps, where we make kernel function calls that are easy for it to identify as ours.

src/modules/exploit_detection/p_exploit_detection.h
kerneltoast: An attack like that is really not feasible, and if a rootkit is capable of that, then the rest of LKRG is "vulnerable" in the same way.

The reason why it's not feasible to permanently transfer control of the CPU out from here is because it would be trashing the kernel's internal state. For example, preempt_count() would need to be fixed, because preemption is disabled in kprobe hooks. A rootkit would need to try VERY hard to restore the kernel's internal state enough for the current task to return to the scheduler again.

solardiz:
@kerneltoast I've just made the changes you requested. Including removal of two redundant calls, although I'm not sure that's an improvement.

src/modules/exploit_detection/p_exploit_detection.h
solardiz: Well, maybe not "permanently transfer control of the CPU out from here" as far as the kernel is concerned, but just skip the rest of our hook - remove a stack frame and transfer control to where our hook was called from. Too low-level to be very portable, but still.

Nevermind, I mostly agree.

solardiz:
@kerneltoast Any re-review from you? I'm also adding `PF_NPROC_EXCEEDED`.

kerneltoast (approved):
This looks much better, thanks.

PR #470, opened 2026-04-10
Use task works to dump ED tasks instead of freezing userspace

kerneltoast:
Sanity tested by printing all tasks added to the ED tree.

### DIFF
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -1000,9 +1000,8 @@ static int p_insert_ed_task(struct p_ed_process *p_source) {
    return ret;
 }
 
-void p_dump_task_f(void *p_arg) {
+void p_dump_task_f(struct task_struct *p_task) {
 
-   struct task_struct *p_task = (struct task_struct *)p_arg;
    struct p_ed_process *p_tmp;
 
    if (unlikely((p_tmp = alloc_ed_task(p_task)) == NULL)) {
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -22,6 +22,8 @@ static unsigned long p_global_off_cookie;
 static unsigned long p_global_cnt_cookie;
 struct kmem_cache *p_ed_wq_valid_cache = NULL;
 static struct kmem_cache *p_ed_pcfi_cache = NULL;
+static atomic_t p_task_work_cnt = ATOMIC_INIT(0);
+static DECLARE_WAIT_QUEUE_HEAD(p_task_work_waitq);
 
 unsigned long p_pcfi_CPU_flags;
 
@@ -1038,29 +1040,136 @@ void p_dump_task_f(void *p_arg) {
    }
 }
 
+static void p_dump_task_work_cb(struct callback_head *head) {
+
+   struct task_struct *p_task = current;
+
+   /*
+    * Clear out our function pointer from the RCU callback head so module exit
+    * knows to not bother trying to cancel a task work for this task.
+    */
+   p_task->rcu.func = NULL;
+
+   /*
+    * When a task exits, exit_task_work() flushes all the queued task workers.
+    * This occurs after PF_EXITING is set for the exiting task. As such, we can
+    * check for PF_EXITING to know if we are getting called after the task has
+    * entered do_exit(), and the check isn't racy. This way, we won't leak
+    * memory allocating an ed struct after our do_exit() kprobe that frees it.
+    */
+   if (!(p_task->flags & PF_EXITING))
+      p_dump_task_f(p_task);
+
+   /*
+    * Decrement the task work counter and wake up the module exit thread if
+    * there are no more task works in flight and it happens to be waiting. The
+    * atomic_dec_and_test() executes a full memory barrier, providing the
+    * necessary ordering before the lockless waitqueue_active() check so that a
+    * wake-up won't be missed (which would make module exit hang forever).
+    */
+   if (atomic_dec_and_test(&p_task_work_cnt) &&
+       waitqueue_active(&p_task_work_waitq))
+      wake_up(&p_task_work_waitq);
+}
+
+static void p_dump_task_work_add(struct task_struct *p_task) {
+
+   /*
+    * Leverage the RCU callback head in task_struct for memory in order to queue
+    * a task work onto this task. This is safe so long as we don't trample over
+    * the RCU callback head while it's in use for RCU freeing the task, which is
+    * accomplished as described by the comment in p_dump_ed_tasks().
+    */
+   p_task->rcu.func = p_dump_task_work_cb;
+
+   /*
+    * Add the task work and trigger it right before the task context switches
+    * back to userspace. If the task is currently running in userspace, then
+    * this will kick the task into the kernel so it will run soon. If the task
+    * is sleeping then the task won't be woken up since there's no need to wake
+    * it up; if it's sleeping then it can't do anything malicious anyway.
+    *
+    * The task work counter is used to track task works in-flight for
+    * synchronization with module exit. We need to increment the task work
+    * counter before knowing if the task work add succeeded, because it may race
+    * with the task work callback's counter decrement. If the task work add
+    * failed (it only fails when the task is exiting), then we decrement the
+    * counter back to where it was. Since it's guaranteed that this function
+    * cannot be running anywhere after the probe hooks are uninstalled in module
+    * exit, decrementing the task work counter doesn't require a check to wake
+    * up the module exit waiter.
+    *
+    * On failure, we also uninit the RCU callback head so that we can track
+    * which tasks still have our task work queued.
+    */
+   atomic_inc(&p_task_work_cnt);
+   if (P_SYM_CALL(p_task_work_add, p_task, &p_task->rcu, 1 /* TWA_RESUME */)) {
+      atomic_dec(&p_task_work_cnt);
+      p_task->rcu.func = NULL;
+   }
+}
+
+static void p_ed_task_work_exit(void) {
+
+   struct task_struct *p_ptmp, *p_tmp;
+
+   if (!atomic_read(&p_task_work_cnt))
+      return;
+
+   /*
+    * Cancel all of our task works that are still enqueued. The task list lock
+    * is needed for the same reason described in p_dump_ed_tasks().
+    */
+   read_lock(P_SYM(p_tasklist_lock));
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 14, 0)
+   for_each_process_thread(p_ptmp, p_tmp) {
+#else
+   do_each_thread(p_ptmp, p_tmp) {
+#endif
+
+      if (P_SYM_CALL(p_task_work_cancel_func, p_tmp, p_dump_task_work_cb))
+         atomic_dec(&p_task_work_cnt);
+
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 14, 0)
+   }
+#else
+   } while_each_thread(p_ptmp, p_tmp);
+#endif
+   read_unlock(P_SYM(p_tasklist_lock));
+
+   /* Wait for any in-flight task work callbacks to finish running */
+   wait_event(p_task_work_waitq, !atomic_read(&p_task_work_cnt));
+}
+
 static void p_dump_ed_tasks(void) {
 
    struct task_struct *p_ptmp, *p_tmp;
 
-   rcu_read_lock();
+   /*
+    * Block tasks from getting removed from the task list and possibly getting
+    * queued up for RCU free. If we overwrite a task's rcu_head after the task
+    * has been queued up for RCU free, then we will corrupt RCU's object list.
+    * This can only happen after the task is removed from the task list, so by
+    * taking the task list lock it is guaranteed that none of the tasks we see
+    * will be allowed to get RCU freed.
+    */
+   read_lock(P_SYM(p_tasklist_lock));
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 14, 0)
    for_each_process_thread(p_ptmp, p_tmp) {
 #else
-   // tasklist_lock
    do_each_thread(p_ptmp, p_tmp) {
 #endif
 
       /* do not touch kernel threads or the global init */
       if (p_is_ed_task(p_tmp))
-         p_dump_task_f(p_tmp);
+         p_dump_task_work_add(p_tmp);
 
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 14, 0)
    }
 #else
-   // tasklist_unlock
    } while_each_thread(p_ptmp, p_tmp);
 #endif
-   rcu_read_unlock();
+   read_unlock(P_SYM(p_tasklist_lock));
 }
 
 static unsigned int p_iterate_lkrg_tasks_paranoid(void) {
@@ -1935,6 +2044,8 @@ void p_exploit_detection_exit(void) {
       p_fh_it->uninstall();
    }
 
+   /* Cancel and wait for all ED task works to finish running */
+   p_ed_task_work_exit();
    /* Delete cache for ED wq validation */
    p_ed_wq_valid_cache_delete();
    /* Delete cache for ED CFI validation */
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -22,6 +22,8 @@ static unsigned long p_global_off_cookie;
 static unsigned long p_global_cnt_cookie;
 struct kmem_cache *p_ed_wq_valid_cache = NULL;
 static struct kmem_cache *p_ed_pcfi_cache = NULL;
+static atomic_t p_task_work_cnt = ATOMIC_INIT(0);
+static DECLARE_WAIT_QUEUE_HEAD(p_task_work_waitq);
 
 unsigned long p_pcfi_CPU_flags;
 
@@ -70,13 +72,6 @@ static const struct p_functions_hooks {
      0,
      "Won't enforce validation on 'call_usermodehelper_exec'",
      1
-   },
-   { "wake_up_new_task",
-     p_install_wake_up_new_task_hook,
-     p_uninstall_wake_up_new_task_hook,
-     1,
-     NULL,
-     0
    },
    { "do_exit",
      p_install_do_exit_hook,
@@ -262,6 +257,19 @@ static const struct p_functions_hooks {
      0,
      "Won't enforce validation on 'scm_send'",
      1
+   },
+   /*
+    * !!! DO NOT ADD NEW HOOKS AFTER THIS COMMENT !!!
+    *
+    * wake_up_new_task MUST be the last hook installed. See the comment in
+    * p_exploit_detection_init().
+    */
+   { "wake_up_new_task",
+     p_install_wake_up_new_task_hook,
+     p_uninstall_wake_up_new_task_hook,
+     1,
+     NULL,
+     0
    },
    { NULL, NULL, NULL, 1, NULL, 0 }
 };
@@ -1018,18 +1026,12 @@ void p_dump_task_f(struct task_struct *p_task) {
 #endif
 
    if (unlikely(p_insert_ed_task(p_tmp))) {
-      struct p_ed_process *p_have;
-      rcu_read_lock();
-      if ((p_have = ed_task_find_lock_rcu(p_task))) {
-         p_print_log(P_LOG_FAULT, "Duplicate key for tracking pid %u, name %s (have pid %u, name %s)",
-                     p_tmp->p_ed_task.p_pid, p_tmp->p_ed_task.p_comm,
-                     p_have->p_ed_task.p_pid, p_have->p_ed_task.p_comm);
-         ed_task_unlock(p_have);
-      } else {
-         p_print_log(P_LOG_FAULT, "Duplicate key for tracking pid %u, name %s",
-                     p_tmp->p_ed_task.p_pid, p_tmp->p_ed_task.p_comm);
-      }
-      rcu_read_unlock();
+      /*
+       * The task is already in the ED tree. This is normal at init because the
+       * wake_up_new_task probe is installed before the initial task dump, so a
+       * newly forked task may be added by the probe and then get added again
+       * during the task dump.
+       */
       free_ed_task(p_tmp);
    } else {
       p_print_log(P_LOG_WATCH, "Inserting pid %u, name %s",
@@ -1037,17 +1039,110 @@ void p_dump_task_f(struct task_struct *p_task) {
    }
 }
 
+static void p_dump_task_work_cb(struct callback_head *work) {
+
+   struct task_struct *p_task = current;
+
+   kfree(work);
+
+   /*
+    * When a task exits, exit_task_work() flushes all the queued task workers.
+    * This occurs after PF_EXITING is set for the exiting task. As such, we can
+    * check for PF_EXITING to know if we are getting called after the task has
+    * entered do_exit(), and the check isn't racy. This way, we won't leak
+    * memory allocating an ed struct after our do_exit() kprobe that frees it.
+    */
+   if (!(p_task->flags & PF_EXITING))
+      p_dump_task_f(p_task);
+
+   /*
+    * Decrement the task work counter and wake up the module exit thread if
+    * there are no more task works in flight and it happens to be waiting. The
+    * atomic_dec_and_test() executes a full memory barrier, providing the
+    * necessary ordering before the lockless waitqueue_active() check so that a
+    * wake-up won't be missed (which would make module exit hang forever).
+    */
+   if (atomic_dec_and_test(&p_task_work_cnt) &&
+       waitqueue_active(&p_task_work_waitq))
+      wake_up(&p_task_work_waitq);
+}
+
+static void p_dump_task_work_add(struct task_struct *p_task,
+                                 struct callback_head **p_work) {
+
+   struct callback_head *work = p_work ? *p_work : NULL;
+
+   if (!work) {
+      work = kmalloc(sizeof(*work), GFP_ATOMIC);
+      if (!work)
+         return;
+   }
+
+   init_task_work(work, p_dump_task_work_cb);
+
+   /*
+    * Add the task work so it fires right before the task returns to userspace.
+    * If the task is currently running in userspace, it'll be kicked into the
+    * kernel so it runs soon. If the task is sleeping, it won't be woken up
+    * since there's no need to; a sleeping task can't do anything malicious.
+    *
+    * The task work counter tracks in-flight task works for synchronization
+    * with module exit. We need to increment the counter before task_work_add()
+    * because the callback's decrement may race with us. If task_work_add()
+    * fails (it only fails when the task is exiting), we just decrement the
+    * counter back. This function can't be running anywhere after the probe
+    * hooks are uninstalled in module exit, so the decrement doesn't need to
+    * check if the module exit waiter needs a wake-up.
+    */
+   atomic_inc(&p_task_work_cnt);
+   if (P_SYM_CALL(p_task_work_add, p_task, work, 1 /* TWA_RESUME */))
+      atomic_dec(&p_task_work_cnt);
+   else
+      work = NULL;
+
+   /* On failure, save the buffer for reuse if requested, otherwise free it */
+   if (p_work)
+      *p_work = work;
+   else
+      kfree(work);
+}
+
+static void p_ed_task_work_exit(void) {
+
+   struct task_struct *p_ptmp, *p_tmp;
+   struct callback_head *work;
+
+   if (!atomic_read(&p_task_work_cnt))
+      return;
+
+   rcu_read_lock();
+   for_each_process_thread(p_ptmp, p_tmp) {
+      work = P_SYM_CALL(p_task_work_cancel_func, p_tmp, p_dump_task_work_cb);
+      if (work) {
+         kfree(work);
+         atomic_dec(&p_task_work_cnt);
+      }
+   }
+   rcu_read_unlock();
+
+   /* Wait for any in-flight task work callbacks to finish running */
+   wait_event(p_task_work_waitq, !atomic_read(&p_task_work_cnt));
+}
+
 static void p_dump_ed_tasks(void) {
 
    struct task_struct *p_ptmp, *p_tmp;
+   struct callback_head *work = NULL;
 
    rcu_read_lock();
    for_each_process_thread(p_ptmp, p_tmp) {
       /* do not touch kernel threads or the global init */
       if (p_is_ed_task(p_tmp))
-         p_dump_task_f(p_tmp);
+         p_dump_task_work_add(p_tmp, &work);
    }
    rcu_read_unlock();
+
+   kfree(work);
 }
 
 static unsigned int p_iterate_lkrg_tasks_paranoid(void) {
@@ -1879,9 +1974,18 @@ int p_exploit_detection_init(void) {
       return P_LKRG_GENERAL_ERROR;
    }
 
-   // Dump processes and threads
-   p_dump_ed_tasks();
-
+   /*
+    * Install all probes *before* the initial task dump. If a task were added to
+    * the ED tree while in middle of a probed code path, then the task's shadow
+    * state could be garbage. Since no tasks are in the ED tree yet, the probe
+    * handlers will just see untracked tasks and safely skip them.
+    *
+    * The wake_up_new_task probe is installed last (it's at the end of the
+    * array) because it adds new tasks to the ED tree, so all other probes need
+    * to be in place first. The task dump uses task works, which fire when the
+    * target task returns to userspace, so the task is guaranteed to be outside
+    * of any probed code path when it is added to the ED tree.
+    */
    for (p_fh_it = p_functions_hooks_array; p_fh_it->name != NULL; p_fh_it++) {
       if (p_fh_it->install(p_fh_it->is_isra_safe)) {
          if (!p_fh_it->p_fatal) {
@@ -1892,6 +1996,8 @@ int p_exploit_detection_init(void) {
          return P_LKRG_GENERAL_ERROR;
       }
    }
+
+   p_dump_ed_tasks();
 
    return P_LKRG_SUCCESS;
 
@@ -1912,6 +2018,8 @@ void p_exploit_detection_exit(void) {
       p_fh_it->uninstall();
    }
 
+   /* Cancel and wait for all ED task works to finish running */
+   p_ed_task_work_exit();
    /* Delete cache for ED wq validation */
    p_ed_wq_valid_cache_delete();
    /* Delete cache for ED CFI validation */
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -1043,23 +1043,11 @@ static void p_dump_ed_tasks(void) {
    struct task_struct *p_ptmp, *p_tmp;
 
    rcu_read_lock();
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 14, 0)
    for_each_process_thread(p_ptmp, p_tmp) {
-#else
-   // tasklist_lock
-   do_each_thread(p_ptmp, p_tmp) {
-#endif
-
       /* do not touch kernel threads or the global init */
       if (p_is_ed_task(p_tmp))
          p_dump_task_f(p_tmp);
-
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 14, 0)
    }
-#else
-   // tasklist_unlock
-   } while_each_thread(p_ptmp, p_tmp);
-#endif
    rcu_read_unlock();
 }
 
@@ -1070,12 +1058,7 @@ static unsigned int p_iterate_lkrg_tasks_paranoid(void) {
    struct task_struct *p_ptmp, *p_task;
 
    rcu_read_lock();
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 14, 0)
    for_each_process_thread(p_ptmp, p_task) {
-#else
-   do_each_thread(p_ptmp, p_task) {
-#endif
-
       /* do not touch kernel threads or the global init */
       if (p_is_ed_task(p_task) && p_get_task_state(p_task) != TASK_DEAD) {
          if ((p_tmp = ed_task_find_lock_rcu(p_task))) {
@@ -1086,12 +1069,7 @@ static unsigned int p_iterate_lkrg_tasks_paranoid(void) {
             ed_task_unlock(p_tmp);
          }
       }
-
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 14, 0)
    }
-#else
-   } while_each_thread(p_ptmp, p_task);
-#endif
    rcu_read_unlock();
 
    /* Before leaving, verify current task */
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -1002,7 +1002,7 @@ static int p_insert_ed_task(struct p_ed_process *p_source) {
    return ret;
 }
 
-void p_dump_task_f(void *p_arg) {
+static void p_dump_task_f(void *p_arg) {
 
    struct task_struct *p_task = (struct task_struct *)p_arg;
    struct p_ed_process *p_tmp;
@@ -1072,7 +1072,7 @@ static void p_dump_task_work_cb(struct callback_head *head) {
       wake_up(&p_task_work_waitq);
 }
 
-static void p_dump_task_work_add(struct task_struct *p_task) {
+void p_dump_task_work_add(struct task_struct *p_task) {
 
    /*
     * Leverage the RCU callback head in task_struct for memory in order to queue
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -1008,14 +1008,9 @@ static int p_insert_ed_task(struct p_ed_process *p_source) {
    return ret;
 }
 
-void p_dump_task_f(struct task_struct *p_task) {
+static void p_dump_task_f(struct task_struct *p_task) {
 
-   struct p_ed_process *p_tmp;
-
-   if (unlikely((p_tmp = alloc_ed_task(p_task)) == NULL)) {
-      p_print_log(P_LOG_FAULT, "Can't allocate memory for tracking pid %u, name %s", p_task->pid, p_task->comm);
-      return;
-   }
+   struct p_ed_process *p_tmp = alloc_ed_task(p_task); /* Never fails */
 
    p_update_ed_process(p_tmp, 1);
 //   p_set_ed_process_on(p_tmp);
@@ -1067,8 +1062,8 @@ static void p_dump_task_work_cb(struct callback_head *work) {
       wake_up(&p_task_work_waitq);
 }
 
-static void p_dump_task_work_add(struct task_struct *p_task,
-                                 struct callback_head **p_work) {
+void p_dump_task_work_add(struct task_struct *p_task,
+                          struct callback_head **p_work) {
 
    struct callback_head *work = p_work ? *p_work : NULL;
 
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -1020,9 +1020,8 @@ static int p_insert_ed_task(struct p_ed_process *p_source) {
    return ret;
 }
 
-void p_dump_task_f(void *p_arg) {
+void p_dump_task_f(struct task_struct *p_task) {
 
-   struct task_struct *p_task = (struct task_struct *)p_arg;
    struct p_ed_process *p_tmp;
 
    if (unlikely((p_tmp = alloc_ed_task(p_task)) == NULL)) {
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -1063,23 +1063,11 @@ static void p_dump_ed_tasks(void) {
    struct task_struct *p_ptmp, *p_tmp;
 
    rcu_read_lock();
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 14, 0)
    for_each_process_thread(p_ptmp, p_tmp) {
-#else
-   // tasklist_lock
-   do_each_thread(p_ptmp, p_tmp) {
-#endif
-
       /* do not touch kernel threads or the global init */
       if (p_is_ed_task(p_tmp))
          p_dump_task_f(p_tmp);
-
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 14, 0)
    }
-#else
-   // tasklist_unlock
-   } while_each_thread(p_ptmp, p_tmp);
-#endif
    rcu_read_unlock();
 }
 
@@ -1090,12 +1078,7 @@ static unsigned int p_iterate_lkrg_tasks_paranoid(void) {
    struct task_struct *p_ptmp, *p_task;
 
    rcu_read_lock();
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 14, 0)
    for_each_process_thread(p_ptmp, p_task) {
-#else
-   do_each_thread(p_ptmp, p_task) {
-#endif
-
       /* do not touch kernel threads or the global init */
       if (p_is_ed_task(p_task) && p_get_task_state(p_task) != TASK_DEAD) {
          if ((p_tmp = ed_task_find_lock_rcu(p_task))) {
@@ -1106,12 +1089,7 @@ static unsigned int p_iterate_lkrg_tasks_paranoid(void) {
             ed_task_unlock(p_tmp);
          }
       }
-
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 14, 0)
    }
-#else
-   } while_each_thread(p_ptmp, p_task);
-#endif
    rcu_read_unlock();
 
    /* Before leaving, verify current task */
--- a/src/modules/exploit_detection/ed_task_tree.c
+++ b/src/modules/exploit_detection/ed_task_tree.c
@@ -266,10 +266,7 @@ struct p_ed_process *alloc_ed_task(struct task_struct *tsk)
 {
 	struct p_ed_process *edp;
 
-	edp = kmem_cache_zalloc(ed_task_cache, GFP_ATOMIC);
-	if (!edp)
-		return NULL;
-
+	edp = kmem_cache_zalloc(ed_task_cache, GFP_KERNEL | __GFP_NOFAIL);
 	raw_spin_lock_init(&edp->lock);
 	edp->kill_task_on_unlock = cookie_kill_task_on_unlock;
 	edp->p_ed_task.p_task = tsk;
--- a/src/modules/exploit_detection/ed_task_tree.c
+++ b/src/modules/exploit_detection/ed_task_tree.c
@@ -266,7 +266,7 @@ struct p_ed_process *alloc_ed_task(struct task_struct *tsk)
 {
 	struct p_ed_process *edp;
 
-	edp = kmem_cache_zalloc(ed_task_cache, GFP_ATOMIC);
+	edp = kmem_cache_zalloc(ed_task_cache, GFP_KERNEL | __GFP_NOFAIL);
 	if (!edp)
 		return NULL;
 
--- a/src/modules/exploit_detection/p_exploit_detection.h
+++ b/src/modules/exploit_detection/p_exploit_detection.h
@@ -326,7 +326,7 @@ static inline void p_kill_task_by_task(struct task_struct *p_task) {
 extern struct p_ed_global_variables p_ed_guard_globals;
 extern unsigned long p_pcfi_CPU_flags;
 
-void p_dump_task_f(void *p_arg);
+void p_dump_task_f(struct task_struct *p_task);
 
 void p_verify_addr_limit(struct p_ed_process *p_orig);
 void p_update_ed_process(struct p_ed_process *p_source, char p_stack);
--- a/src/modules/exploit_detection/p_exploit_detection.h
+++ b/src/modules/exploit_detection/p_exploit_detection.h
@@ -326,7 +326,8 @@ static inline void p_kill_task_by_task(struct task_struct *p_task) {
 extern struct p_ed_global_variables p_ed_guard_globals;
 extern unsigned long p_pcfi_CPU_flags;
 
-void p_dump_task_f(struct task_struct *p_task);
+void p_dump_task_work_add(struct task_struct *p_task,
+                          struct callback_head **p_work);
 
 void p_verify_addr_limit(struct p_ed_process *p_orig);
 void p_update_ed_process(struct p_ed_process *p_source, char p_stack);
--- a/src/modules/exploit_detection/p_exploit_detection.h
+++ b/src/modules/exploit_detection/p_exploit_detection.h
@@ -326,7 +326,7 @@ static inline void p_kill_task_by_task(struct task_struct *p_task) {
 extern struct p_ed_global_variables p_ed_guard_globals;
 extern unsigned long p_pcfi_CPU_flags;
 
-void p_dump_task_f(void *p_arg);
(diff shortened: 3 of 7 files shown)


src/modules/exploit_detection/p_exploit_detection.c
```diff
@@ -1038,29 +1040,136 @@ void p_dump_task_f(void *p_arg) {
...
+    * is needed for the same reason described in p_dump_ed_tasks().
+    */
+   read_lock(P_SYM(p_tasklist_lock));
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 14, 0)
+   for_each_process_thread(p_ptmp, p_tmp) {
+#else
+   do_each_thread(p_ptmp, p_tmp) {
+#endif
+
+      if (P_SYM_CALL(p_task_work_cancel_func, p_tmp, p_dump_task_work_cb))
```
kerneltoast: I forgot an optimization here to check for `p_tmp->rcu.func == p_dump_task_work_cb` since `task_work_cancel_func()` is pretty heavy and should be skipped when unneeded.

solardiz:
Thank you @kerneltoast! Unfortunately, this is too much of a change for inclusion in the very next release, which we should make in a week or so. I think we should make tiny fixes to what we already have for now, and consider this PR for merging after the release.

kerneltoast:
Ah, this is buggy because the kernel already uses the rcu_head member of task_struct for another field:
```c

	union {
		refcount_t		rcu_users;
		struct rcu_head		rcu;
	};
```

So the memory for the `rcu` member is used throughout the lifetime of a task_struct, not just when the task_struct is getting freed. And that's why there are those CI test failures. Darn. Guess we'll have to allocate memory for the task work object using GFP_ATOMIC.

kerneltoast:
@solardiz This PR is ready for review now.

kerneltoast:
@PlaidCat It'd be great if you could test this. You should see improved boot time due to the absence of freezing (and freezing retries).

PlaidCat:
## 2 boots ago
```
[plaidcat] $ journalctl --no-hostname  -b  -2 | grep -ie lkrg
May 10 17:58:10 kernel: lkrg: loading out-of-tree module taints kernel.
May 10 17:58:10 kernel: lkrg: module verification failed: signature and/or required key missing - tainting kernel
May 10 17:58:10 kernel: LKRG: ALIVE: Loading LKRG
May 10 17:58:10 kernel: LKRG: ISSUE: Freezing failed 1 time(s), sleeping 100 ms before next try with timeout of 2000 ms
May 10 17:58:10 kernel: LKRG: ALIVE: LKRG initialized successfully
May 10 17:58:10 modprobe[762]: insmod /lib/modules/7.0.5-arch1-1/updates/dkms/lkrg.ko.zst
May 10 17:58:38 kernel:  i2c_smbus k10temp libarc4 thunderbolt pps_core gpio_generic cfg80211 mac_hid rfkill lkrg(OE) ntsync i2c_dev crypto_user nfnetlink xfs amdgpu amdxcp i2c_algo_bit drm_ttm_helper ttm drm_exec drm_panel_backlight_quirks gpu_sched drm_suballoc_helper nvme drm_buddy nvme_core drm_display_helper video nvme_keyring cec nvme_auth hkdf wmi
May 10 18:32:44 modprobe[19531]: rmmod lkrg
May 10 18:32:44 kernel: LKRG: DYING: Unloading LKRG
May 10 18:32:45 kernel: LKRG: DYING: LKRG unloaded
May 10 18:32:45 systemd[1]: lkrg.service: Deactivated successfully.
```

## PR boot
Built from github.com/kerneltoast/lkrg/tree/sultan/ed-task-work
```
[plaidcat]$ uname -a
Linux HK-416a5 7.0.5-arch1-1 #1 SMP PREEMPT_DYNAMIC Fri, 08 May 2026 09:29:23 +0000 x86_64 GNU/Linux
[plaidcat]$ pacman -Q lkrg-dkms-git
lkrg-dkms-git r788.880e422-1
[plaidcat]$ journalctl --no-hostname  -b | grep -ie lkrg
May 11 14:43:46 kernel: lkrg: loading out-of-tree module taints kernel.
May 11 14:43:46 kernel: lkrg: module verification failed: signature and/or required key missing - tainting kernel
May 11 14:43:46 kernel: LKRG: ALIVE: Loading LKRG
May 11 14:43:46 kernel: LKRG: ALIVE: LKRG initialized successfully
May 11 14:43:46 modprobe[750]: insmod /lib/modules/7.0.5-arch1-1/updates/dkms/lkrg.ko.zst
May 11 14:44:14 kernel:  rfkill lkrg(OE) ntsync i2c_dev crypto_user nfnetlink xfs amdgpu amdxcp i2c_algo_bit drm_ttm_helper ttm drm_exec drm_panel_backlight_quirks gpu_sched drm_suballoc_helper nvme drm_buddy nvme_core drm_display_helper nvme_keyring video cec nvme_auth hkdf wmi
```

```
[plaidcat] $ ls -ltr ~/.cache/yay/lkrg-dkms-git/
total 1364
-rw-r--r-- 1 plaidcat plaidcat    195 Dec 13 17:11 dkms.conf
-rw-r--r-- 1 plaidcat plaidcat 149393 Mar 16 18:47 lkrg-dkms-git-r768.07efa4a-1-any.pkg.tar.zst
-rw-r--r-- 1 plaidcat plaidcat 149409 Mar 17 09:32 lkrg-dkms-git-r770.6c5d8ac-1-any.pkg.tar.zst
-rw-r--r-- 1 plaidcat plaidcat 151094 Apr  8 12:50 lkrg-dkms-git-r773.b343507-1-any.pkg.tar.zst
-rw-r--r-- 1 plaidcat plaidcat 151591 Apr 11 09:42 lkrg-dkms-git-r776.d44b347-1-any.pkg.tar.zst
-rw-r--r-- 1 plaidcat plaidcat 151704 Apr 13 10:54 lkrg-dkms-git-r778.bf818a8-1-any.pkg.tar.zst
-rw-r--r-- 1 plaidcat plaidcat 151770 Apr 19 11:40 lkrg-dkms-git-r779.236faa1-1-any.pkg.tar.zst
-rw-r--r-- 1 plaidcat plaidcat 152083 Apr 21 12:48 lkrg-dkms-git-r781.56ab932-1-any.pkg.tar.zst
-rw-r--r-- 1 plaidcat plaidcat 152132 Apr 25 11:34 lkrg-dkms-git-r784.e18822f-1-any.pkg.tar.zst
drwxr-xr-x 6 plaidcat plaidcat    180 May 11 14:41 lkrg-dkms-git
drwxr-xr-x 3 plaidcat plaidcat     56 May 11 14:41 src
-rw-r--r-- 1 plaidcat plaidcat   1669 May 11 14:41 PKGBUILD
drwxr-xr-x 4 plaidcat plaidcat     66 May 11 14:41 pkg
-rw-r--r-- 1 plaidcat plaidcat 153351 May 11 14:41 lkrg-dkms-git-r788.880e422-1-any.pkg.tar.zst

**Comment by PlaidCat** [comment]
May not be ideal today things started to fail
```
[plaidcat ~]$ uname -r
7.0.5-arch1-1
[plaidcat~]$ pacman -Q lkrg-dkms-git
lkrg-dkms-git r788.880e422-1
```

```
[plaidcat@HK-416a5 ~]$ journalctl --no-hostname  -b  -1 | grep -ie lkrg
May 12 20:02:15 kernel: lkrg: loading out-of-tree module taints kernel.
May 12 20:02:15 kernel: lkrg: module verification failed: signature and/or required key missing - tainting kernel
May 12 20:02:15 kernel: LKRG: ALIVE: Loading LKRG
May 12 20:02:16 kernel: RIP: 0010:p_arch_static_call_transform_ret+0x44/0x270 [lkrg]
May 12 20:02:16 kernel: Modules linked in: snd_pcm kvm_amd(+) snd_timer asus_nb_wmi(-) snd asus_armoury fjes(-) spd5118 asus_ec_sensors kvm ccp soundcore eeepc_wmi firmware_attributes_class sp5100_tco asus_wmi irqbypass mousedev platform_profile ghash_clmulni_intel joydev sparse_keymap aesni_intel i2c_piix4 rfkill intel_wmi_thunderbolt rapl wmi_bmof pcspkr k10temp thunderbolt(+) gpio_amdpt i2c_smbus gpio_generic mac_hid lkrg(OE+) ntsync i2c_dev crypto_user nfnetlink xfs amdgpu amdxcp i2c_algo_bit drm_ttm_helper ttm drm_exec drm_panel_backlight_quirks nvme gpu_sched drm_suballoc_helper drm_buddy nvme_core video drm_display_helper nvme_keyring nvme_auth cec hkdf wmi
May 12 20:02:16 kernel: RIP: 0010:p_arch_static_call_transform_ret+0x44/0x270 [lkrg]
May 12 20:02:16 kernel: LKRG: ALERT: DETECT: Task: Stack base invalid for pid 939, name (spawn)
May 12 20:02:16 kernel: LKRG: ALERT: DETECT: Task: Stack size mismatch for pid 939, name (spawn)
May 12 20:02:16 kernel: LKRG: ALERT: DETECT: Task: Stack base invalid for pid 939, name (spawn)
May 12 20:02:16 kernel: LKRG: ALERT: DETECT: Task: Stack size mismatch for pid 939, name (spawn)
May 12 20:02:16 kernel: LKRG: ALERT: BLOCK: Task: Killing pid 939, name (spawn)
```

### Longer log
```
May 12 20:02:16 mtp-probe[919]: bus: 5, device: 8 was not an MTP device
May 12 20:02:16 kernel: kvm_amd: TSC scaling supported
May 12 20:02:16 kernel: kvm_amd: Nested Virtualization enabled
May 12 20:02:16 kernel: kvm_amd: Nested Paging enabled
May 12 20:02:16 kernel: kvm_amd: LBR virtualization supported
May 12 20:02:16 kernel: kvm_amd: AVIC enabled
May 12 20:02:16 kernel: kvm_amd: x2AVIC enabled (max 512 vCPUs)
May 12 20:02:16 kernel: kvm_amd: Virtual VMLOAD VMSAVE supported
May 12 20:02:16 kernel: kvm_amd: Virtual GIF supported
May 12 20:02:16 kernel: kvm_amd: Virtual NMI enabled
May 12 20:02:16 kernel: BUG: kernel NULL pointer dereference, address: 0000000000000050
May 12 20:02:16 kernel: #PF: supervisor read access in kernel mode
May 12 20:02:16 kernel: #PF: error_code(0x0000) - not-present page
May 12 20:02:16 kernel: PGD 0 P4D 0
May 12 20:02:16 kernel: Oops: Oops: 0000 [#1] SMP NOPTI
May 12 20:02:16 kernel: CPU: 21 UID: 0 PID: 823 Comm: (udev-worker) Tainted: G           OE       7.0.5-arch1-1 #1 PREEMPT(full)  0d14cfb831c>
May 12 20:02:16 kernel: Tainted: [O]=OOT_MODULE, [E]=UNSIGNED_MODULE
May 12 20:02:16 kernel: Hardware name: ASUS System Product Name/ProArt X670E-CREATOR WIFI, BIOS 3603 03/09/2026
May 12 20:02:16 kernel: RIP: 0010:p_arch_static_call_transform_ret+0x44/0x270 [lkrg]
May 12 20:02:16 kernel: Code: 00 0f 84 d2 00 00 00 83 3d 7c 4d 2d 00 03 0f 87 92 81 00 00 e8 3d e4 ff ff 8b 1d c3 d0 2c 00 48 6b db 68 48 03 >
May 12 20:02:16 kernel: RSP: 0018:ffffd50685be7a80 EFLAGS: 00010246
May 12 20:02:16 kernel: RAX: ffffffffc0686270 RBX: 0000000000000000 RCX: ffffd50685be7ae0
May 12 20:02:16 kernel: RDX: ffffffffb861f78b RSI: ffffd50685be7ae0 RDI: ffffd50685a63210
May 12 20:02:16 kernel: RBP: ffffd50685a63220 R08: ffffffffb9f3d940 R09: ffff8dad009dd2b0
May 12 20:02:16 kernel: R10: ffffffffbaaf13a0 R11: 0000000000000000 R12: ffffffffb861f78b
May 12 20:02:16 kernel: R13: ffffd50685be7b88 R14: ffff8dad1c26a200 R15: ffffd50685be7ae0
May 12 20:02:16 kernel: FS:  00007f4ab7d97880(0000) GS:ffff8dc3e46a7000(0000) knlGS:0000000000000000
May 12 20:02:16 kernel: CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
May 12 20:02:16 kernel: CR2: 0000000000000050 CR3: 000000010f89d000 CR4: 0000000000f50ef0
May 12 20:02:16 kernel: PKRU: 55555554
May 12 20:02:16 kernel: Call Trace:
May 12 20:02:16 kernel:  <TASK>
May 12 20:02:16 kernel:  kretprobe_rethook_handler+0x42/0x60
May 12 20:02:16 kernel:  rethook_trampoline_handler+0x9d/0x180
May 12 20:02:16 kernel:  ? __pfx_svm_check_processor_compat+0x10/0x10 [kvm_amd 3c1f102c19623ac2379a8b5f1cd5727a264bacd7]
May 12 20:02:16 kernel:  ? __SCT__tp_func_kvm_rmp_fault+0x8/0x8 [kvm e17663227b10ec08b9bd7b991dd3e43137e3655a]
May 12 20:02:16 kernel:  arch_rethook_trampoline_callback+0x3b/0x50
May 12 20:02:16 kernel:  arch_rethook_trampoline+0x2c/0x60
May 12 20:02:16 kernel:  ? __SCT__tp_func_kvm_rmp_fault+0x8/0x8 [kvm e17663227b10ec08b9bd7b991dd3e43137e3655a]
May 12 20:02:16 kernel:  ? __pfx_svm_check_processor_compat+0x10/0x10 [kvm_amd 3c1f102c19623ac2379a8b5f1cd5727a264bacd7]
May 12 20:02:16 kernel:  ? __static_call_update+0x8b/0x210
May 12 20:02:16 kernel:  osnoise_arch_unregister+0x210/0x210
May 12 20:02:16 kernel:  ? __SCT__tp_func_kvm_rmp_fault+0x8/0x8 [kvm e17663227b10ec08b9bd7b991dd3e43137e3655a]
May 12 20:02:16 kernel:  ? __pfx_svm_check_processor_compat+0x10/0x10 [kvm_amd 3c1f102c19623ac2379a8b5f1cd5727a264bacd7]
May 12 20:02:16 kernel:  ? __pfx_svm_init+0x10/0x10 [kvm_amd 3c1f102c19623ac2379a8b5f1cd5727a264bacd7]
May 12 20:02:16 kernel:  kvm_x86_vendor_init+0x279/0x1950 [kvm e17663227b10ec08b9bd7b991dd3e43137e3655a]
May 12 20:02:16 kernel:  ? __pfx_svm_init+0x10/0x10 [kvm_amd 3c1f102c19623ac2379a8b5f1cd5727a264bacd7]
May 12 20:02:16 kernel:  svm_init+0x32/0x70 [kvm_amd 3c1f102c19623ac2379a8b5f1cd5727a264bacd7]
May 12 20:02:16 kernel:  do_one_initcall+0x8e/0x400
May 12 20:02:16 kernel:  do_init_module+0x6a/0x260
May 12 20:02:16 kernel:  ? init_module_from_file+0xc8/0x180
May 12 20:02:16 kernel:  init_module_from_file+0xc8/0x180
May 12 20:02:16 kernel:  idempotent_init_module+0x23e/0x360
May 12 20:02:16 kernel:  __x64_sys_finit_module+0x7f/0x110
May 12 20:02:16 kernel:  ? __x64_sys_close+0x47/0xa0
May 12 20:02:16 kernel:  do_syscall_64+0x119/0x1640
May 12 20:02:16 kernel:  ? do_user_addr_fault+0x237/0x710
May 12 20:02:16 kernel:  ? exc_page_fault+0x90/0x1f0
May 12 20:02:16 kernel:  ? __irq_exit_rcu+0x4c/0xf0
May 12 20:02:16 kernel:  entry_SYSCALL_64_after_hwframe+0x76/0x7e
May 12 20:02:16 kernel: RIP: 0033:0x7f4ab751af9d
May 12 20:02:16 kernel: Code: ff c3 66 2e 0f 1f 84 00 00 00 00 00 90 f3 0f 1e fa 48 89 f8 48 89 f7 48 89 d6 48 89 ca 4d 89 c2 4d 89 c8 4c 8b >
May 12 20:02:16 kernel: RSP: 002b:00007ffd0255b158 EFLAGS: 00000246 ORIG_RAX: 0000000000000139
May 12 20:02:16 kernel: RAX: ffffffffffffffda RBX: 0000563cbf1f3c30 RCX: 00007f4ab751af9d
May 12 20:02:16 kernel: RDX: 0000000000000004 RSI: 00007f4ab76c72f2 RDI: 000000000000004d
May 12 20:02:16 kernel: RBP: 00007ffd0255b1f0 R08: 0000000000000000 R09: 0000563cbf1eb690
May 12 20:02:16 kernel: R10: 0000000000000000 R11: 0000000000000246 R12: 00007f4ab76c72f2
May 12 20:02:16 kernel: R13: 0000000000020000 R14: 0000563cbf1ebe20 R15: 0000563cbf1f3c30
May 12 20:02:16 kernel:  </TASK>
May 12 20:02:16 kernel: Modules linked in: snd_pcm kvm_amd(+) snd_timer asus_nb_wmi(-) snd asus_armoury fjes(-) spd5118 asus_ec_sensors kvm c>
May 12 20:02:16 kernel: CR2: 0000000000000050
May 12 20:02:16 systemd-udevd[792]: cpu7: Worker [823] terminated by signal 9 (KILL).
May 12 20:02:16 kernel: ---[ end trace 0000000000000000 ]---
May 12 20:02:16 kernel: RIP: 0010:p_arch_static_call_transform_ret+0x44/0x270 [lkrg]
May 12 20:02:16 kernel: Code: 00 0f 84 d2 00 00 00 83 3d 7c 4d 2d 00 03 0f 87 92 81 00 00 e8 3d e4 ff ff 8b 1d c3 d0 2c 00 48 6b db 68 48 03 >
May 12 20:02:16 kernel: RSP: 0018:ffffd50685be7a80 EFLAGS: 00010246
May 12 20:02:16 kernel: RAX: ffffffffc0686270 RBX: 0000000000000000 RCX: ffffd50685be7ae0
May 12 20:02:16 kernel: RDX: ffffffffb861f78b RSI: ffffd50685be7ae0 RDI: ffffd50685a63210
May 12 20:02:16 kernel: RBP: ffffd50685a63220 R08: ffffffffb9f3d940 R09: ffff8dad009dd2b0
May 12 20:02:16 kernel: R10: ffffffffbaaf13a0 R11: 0000000000000000 R12: ffffffffb861f78b
May 12 20:02:16 kernel: R13: ffffd50685be7b88 R14: ffff8dad1c26a200 R15: ffffd50685be7ae0
May 12 20:02:16 kernel: FS:  00007f4ab7d97880(0000) GS:ffff8dc3e46a7000(0000) knlGS:0000000000000000
May 12 20:02:16 kernel: CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
May 12 20:02:16 kernel: CR2: 0000000000000050 CR3: 000000010f89d000 CR4: 0000000000f50ef0
May 12 20:02:16 kernel: PKRU: 55555554
May 12 20:02:16 kernel: note: (udev-worker)[823] exited with irqs disabled
May 12 20:02:16 kernel: note: (udev-worker)[823] exited with preempt_count 1
May 12 20:02:16 mtp-probe[929]: checking bus 5, device 5: "/sys/devices/pci0000:00/0000:00:02.1/0000:05:00.0/0000:06:08.0/0000:08:00.0/0000:0>
May 12 20:02:16 mtp-probe[929]: bus: 5, device: 5 was not an MTP device
May 12 20:02:16 mtp-probe[930]: checking bus 5, device 3: "/sys/devices/pci0000:00/0000:00:02.1/0000:05:00.0/0000:06:08.0/0000:08:00.0/0000:0>
May 12 20:02:16 mtp-probe[931]: checking bus 5, device 6: "/sys/devices/pci0000:00/0000:00:02.1/0000:05:00.0/0000:06:08.0/0000:08:00.0/0000:0>
May 12 20:02:16 mtp-probe[931]: bus: 5, device: 6 was not an MTP device
May 12 20:02:16 mtp-probe[933]: checking bus 5, device 9: "/sys/devices/pci0000:00/0000:00:02.1/0000:05:00.0/0000:06:08.0/0000:08:00.0/0000:0>
May 12 20:02:16 mtp-probe[934]: checking bus 5, device 7: "/sys/devices/pci0000:00/0000:00:02.1/0000:05:00.0/0000:06:08.0/0000:08:00.0/0000:0>
May 12 20:02:16 mtp-probe[933]: bus: 5, device: 9 was not an MTP device
May 12 20:02:16 mtp-probe[934]: bus: 5, device: 7 was not an MTP device
May 12 20:02:16 mtp-probe[936]: checking bus 5, device 8: "/sys/devices/pci0000:00/0000:00:02.1/0000:05:00.0/0000:06:08.0/0000:08:00.0/0000:0>
May 12 20:02:16 mtp-probe[936]: bus: 5, device: 8 was not an MTP device
May 12 20:02:16 kernel: LKRG: ALERT: DETECT: Task: Stack base invalid for pid 939, name (spawn)
May 12 20:02:16 kernel: LKRG: ALERT: DETECT: Task: Stack size mismatch for pid 939, name (spawn)
May 12 20:02:16 kernel: LKRG: ALERT: DETECT: Task: Stack base invalid for pid 939, name (spawn)
May 12 20:02:16 kernel: LKRG: ALERT: DETECT: Task: Stack size mismatch for pid 939, name (spawn)
May 12 20:02:16 kernel: LKRG: ALERT: BLOCK: Task: Killing pid 939, name (spawn)
May 12 20:02:16 (udev-worker)[824]: hidraw3: Process 'fido_id' terminated by signal KILL.
May 12 20:02:16 (udev-worker)[824]: hidraw3: Failed to wait for spawned command 'fido_id': Input/output error
May 12 20:02:16 (udev-worker)[824]: hidraw3: /usr/lib/udev/rules.d/60-fido-id.rules:5 IMPORT{program}="fido_id": Failed to execute "fido_id",>
May 12 20:02:16 mtp-probe[930]: bus: 5, device: 3 was not an MTP device
May 12 20:02:17 systemd[1]: Mounting /boot...
```

My Kernel and LKRG did not update from the last update.
I had to set `nolkrg` 

Not sure if this is directly related.  let me know if there are things I should update

Claudes Analysis 
[lkrg-boot-crash-analysis-2026-05-12.md](https://github.com/user-attachments/files/27666067/lkrg-boot-crash-analysis-2026-05-12.md)

**Comment by PlaidCat** [comment]
Just did an update and a new kernel came it and it booted fine
```
[plaidcat~]$ journalctl --no-hostname -b | grep -i lkrg
May 14 18:45:55 kernel: lkrg: loading out-of-tree module taints kernel.
May 14 18:45:55 kernel: lkrg: module verification failed: signature and/or required key missing - tainting kernel
May 14 18:45:55 kernel: LKRG: ALIVE: Loading LKRG
May 14 18:45:55 kernel: LKRG: ALIVE: LKRG initialized successfully
May 14 18:45:55 modprobe[755]: insmod /lib/modules/7.0.6-arch1-1/updates/dkms/lkrg.ko.zst
May 14 18:46:24 kernel:  rfkill lkrg(OE) ntsync crypto_user i2c_dev nfnetlink xfs amdgpu amdxcp i2c_algo_bit drm_ttm_helper ttm drm_exec drm_panel_backlight_quirks gpu_sched drm_suballoc_helper nvme drm_buddy drm_display_helper nvme_core video nvme_keyring cec nvme_auth hkdf wmi
[plaidcat ~]$ uname -r
7.0.6-arch1-1
[plaidcat ~]$ pacman -Q lkrg-dkms-git
lkrg-dkms-git r788.880e422-1
```

Correction something ... maybe race is happening, cold boots cause it to happen more often.
I'll rebuild against the mainline to validate its not this commit.

**Comment by solardiz** [comment]
> Not sure if this is directly related. let me know if there are things I should update
> 
> Claudes Analysis [lkrg-boot-crash-analysis-2026-05-12.md](https://github.com/user-attachments/files/27666067/lkrg-boot-crash-analysis-2026-05-12.md)

The analysis looks reasonable at first glance. Please create a separate issue for this as it looks unrelated to this PR. Thank you, @PlaidCat!

**Comment by solardiz** [comment]
I'm testing this with a shell loop loading/unloading LKRG on CentOS 7 (to choose something perhaps very different from where @kerneltoast developed this), no issues seen so far.

**Comment by solardiz** [comment]
> I'm testing this with a shell loop loading/unloading LKRG on CentOS 7

Running this for ~3 hours, I caught one false positive detection `unexpected 'off' flag depth -1 for pid 18175, name bash` between `Unloading LKRG` and `LKRG unloaded`. No idea if this is related to the changes in this PR or not, but it may be!

**Comment by solardiz** [comment]
Caught one more issue on the same CentOS 7 when stress-testing LKRG with this PR's changes:

```
[16236.072864] LKRG: ALIVE: Loading LKRG
[16236.427859] LKRG: ISSUE: register_k[ret]probe() for ovl_dentry_is_whiteout failed! [err=-2]
[16236.427859] LKRG: ISSUE: Can't hook 'ovl_dentry_is_whiteout'. This is expected when OverlayFS is not used.
[16236.675725] LKRG: ALIVE: LKRG initialized successfully
[16236.690217] LKRG: DYING: Unloading LKRG
[16237.250477] LKRG: ALERT: DETECT: Task: unexpected 'off' flag depth -1 for pid 21591, name bash
[16237.250479] LKRG: ALERT: BLOCK: Task: Killing pid 21591, name bash
[16237.250604] LKRG: ALERT: DETECT: Task: unexpected 'off' flag depth -1 for pid 21592, name bash
[16237.250605] LKRG: ALERT: BLOCK: Task: Killing pid 21592, name bash
[16237.251668] LKRG: ALERT: DETECT: Task: unexpected 'off' flag depth -1 for pid 21590, name bash
[16237.251694] LKRG: ALERT: BLOCK: Task: Killing pid 21590, name dmesg
[16237.258724] LKRG: ALERT: DETECT: Task: unexpected 'off' flag depth -1 for pid 498, name systemd-journal
[16237.258859] LKRG: ALERT: BLOCK: Task: Killing pid 498, name systemd-journal
[16237.268864] LKRG: ALERT: BLOCK: Task: Killing pid 21594, name (journald)
[16237.271472] LKRG: ALERT: BLOCK: Task: Killing pid 21594, name (journald)
[16237.272206] LKRG: ALERT: BLOCK: Task: Killing pid 21594, name (journald)
[16237.272453] LKRG: ALERT: BLOCK: Task: Killing pid 21594, name (journald)
[16237.274780] LKRG: ALERT: BLOCK: Task: Killing pid 21596, name (journald)
[16237.275582] LKRG: ALERT: BLOCK: Task: Killing pid 21596, name (journald)
[16237.276366] LKRG: ALERT: BLOCK: Task: Killing pid 21596, name (journald)
[16237.276567] LKRG: ALERT: BLOCK: Task: Killing pid 21596, name (journald)
[16237.279383] LKRG: ALERT: DETECT: Task: Stack base invalid for pid 21598, name (journald)
[16237.280128] LKRG: ALERT: DETECT: Task: Stack size mismatch for pid 21598, name (journald)
[16237.280275] LKRG: ALERT: BLOCK: Task: Killing pid 21598, name (journald)
[16237.281539] LKRG: ALERT: DETECT: Task: Stack base invalid for pid 21598, name (journald)
[16237.282218] LKRG: ALERT: DETECT: Task: Stack size mismatch for pid 21598, name (journald)
[16237.282511] LKRG: ALERT: BLOCK: Task: Killing pid 21598, name (journald)
[16237.282511] LKRG: ALERT: DETECT: Task: Stack base invalid for pid 21598, name (journald)
[16237.282511] LKRG: ALERT: DETECT: Task: Stack size mismatch for pid 21598, name (journald)
[16237.282511] LKRG: ALERT: BLOCK: Task: Killing pid 21598, name (journald)
[16237.285691] LKRG: ALERT: DETECT: Task: Stack base invalid for pid 21598, name (journald)
[16237.286360] LKRG: ALERT: DETECT: Task: Stack size mismatch for pid 21598, name (journald)
[16237.286663] LKRG: ALERT: BLOCK: Task: Killing pid 21598, name (journald)
[16237.289531] LKRG: ALERT: DETECT: Task: Stack base invalid for pid 21600, name (journald)
[16237.290205] LKRG: ALERT: DETECT: Task: Stack size mismatch for pid 21600, name (journald)
[16237.290425] LKRG: ALERT: BLOCK: Task: Killing pid 21600, name (journald)
[16237.291561] LKRG: ALERT: DETECT: Task: Stack base invalid for pid 21600, name (journald)
[16237.292273] LKRG: ALERT: DETECT: Task: Stack size mismatch for pid 21600, name (journald)
[16237.292546] LKRG: ALERT: BLOCK: Task: Killing pid 21600, name (journald)
[16237.292546] LKRG: ALERT: DETECT: Task: Stack base invalid for pid 21600, name (journald)
[16237.292546] LKRG: ALERT: DETECT: Task: Stack size mismatch for pid 21600, name (journald)
[16237.292546] LKRG: ALERT: BLOCK: Task: Killing pid 21600, name (journald)
[16237.295576] LKRG: ALERT: DETECT: Task: Stack base invalid for pid 21600, name (journald)
[16237.296226] LKRG: ALERT: DETECT: Task: Stack size mismatch for pid 21600, name (journald)
[16237.296553] LKRG: ALERT: BLOCK: Task: Killing pid 21600, name (journald)
[16237.310631] systemd-journald[21602]: File /run/log/journal/6a3b6a99892341b6a4fd07af813a018c/system.journal corrupted or uncleanly shut down, renaming and replacing.
[16237.519907] LKRG: DYING: LKRG unloaded
```

Then in a couple more hours, I couldn't trigger an issue on the same CentOS 7 with latest LKRG as currently in `main` branch here, but this may be because before this PR the loading and unloading of LKRG is much slower, so I ran fewer iterations in the same amount of real time.

**Comment by solardiz** [comment]
Looking at previous comments here, I notice that we also had this in @PlaidCat's issue report:

> May 12 20:02:16 kernel: LKRG: ALERT: DETECT: Task: Stack base invalid for pid 939, name (spawn)
> May 12 20:02:16 kernel: LKRG: ALERT: DETECT: Task: Stack size mismatch for pid 939, name (spawn)

I think this is not explained in his `lkrg-boot-crash-analysis-2026-05-12.md` attached here. That analysis talks about two other issues. Sure a NULL pointer dereference crash could have other consequences later, but I don't see how that specific crash would result in the above. So it could be a separate issue.

File: src/modules/exploit_detection/p_exploit_detection.c [hunk 2]
```diff
@@ -1039,48 +1041,155 @@ void p_dump_task_f(void *p_arg) {
...
+
+void p_dump_task_work_add(struct task_struct *p_task,
+                          struct callback_head **p_work) {
+
+   struct callback_head *work = p_work ? *p_work : NULL;
+
+   if (!work) {
+      work = kmalloc(sizeof(*work), GFP_ATOMIC);
+      if (!work)
+         return;
```
solardiz, line 1102: Don't we want to log a message in the (very unlikely) event that this allocation fails?

solardiz:
@kerneltoast I'm afraid I failed to fully understand this part of the commit message:

```
All probes are installed before the initial task dump so that the probe
handlers see untracked tasks and safely skip them. New tasks forked after
probe installation are caught by the wake_up_new_task probe, which may
cause the initial task dump to redundantly add a task, but duplicate add
attempts are handled gracefully by the ED tree helper so it's fine.

For the same reason that the initial task dump must come after probe
installation, the wake_up_new_task probe specifically must be installed
after all other probes are installed.
```

We want to start tracking all tasks, both those that were already running when we started and those new tasks that are appearing while we're starting up. Is the existence of these two categories of tasks the primary motivation behind the ordering described in the above two paragraphs?

The first sentence reads weird to me - surely we're not doing that "so that" or in other words with the goal to have the "handlers see untracked tasks and safely skip them", this is just something we accept would be happening as an alternative to something worse, right? Perhaps the word "that" is extraneous, and only "so" should be left?

The second paragraph starts with "For the same reason", but we gave no reason other than via "so that" above... which surely is not the actual reason? Maybe it's in fact "the same" unspecified reason, but stating merely that it's "the same" is not as helpful as actually giving the reason would be.

I don't mean to nitpick, but this affects my incomplete understanding of what this tries to say, and more importantly of why we're actually doing things in this way.

What worse would happen in what other (wrong) design, which this description implicitly compares this design against? Is this problem fully avoided by this design, and why?

solardiz:
OK, "the same reason" refers to what the commit message said further above, before these two paragraphs I just quoted. It said:

```
The task work callback runs before the target task returns to
userspace, at a place in the kernel unaffected by LKRG's probes. This is
important because tasks need to be added to the ED tree at a point where
they can't be in the middle of a probed code path; if a task were added
while in the middle of a cred operation, for example, LKRG wouldn't be able
to accurately track the depth of the cred operation, and the task's shadow
state would be garbage.
```

and we add this code comment:

```
    * Install all probes *before* the initial task dump. If a task were added to
    * the ED tree while in middle of a probed code path, then the task's shadow
    * state could be garbage. Since no tasks are in the ED tree yet, the probe
    * handlers will just see untracked tasks and safely skip them.
    *
    * The wake_up_new_task probe is installed last (it's at the end of the
    * array) because it adds new tasks to the ED tree, so all other probes need
    * to be in place first. The task dump uses task works, which fire when the
    * target task returns to userspace, so the task is guaranteed to be outside
    * of any probed code path when it is added to the ED tree.
```

solardiz:
I still don't fully understand. Given this:

```
    * [...] The task dump uses task works, which fire when the
    * target task returns to userspace, so the task is guaranteed to be outside
    * of any probed code path when it is added to the ED tree.
```

wouldn't it have been safe to dump the existing tasks before probe installation? I'm not saying we should, just that "the same reason" wording in the commit message may be wrong.

Further, the fourth commit here does:

```
    Now that we have task work infrastructure in place for adding ED tasks, use
    it even from the wake_up_new_task hook.
```

which seems to make the ordering of hook installation unimportant - but given it's a later separate commit, sure we can't rely on that in the third commit's message yet.

solardiz:
In better news, I think I now understand the task cred corruption false positives we got on LKRG unload with this PR's changes. We no longer freeze on unload, so we're removing hooks non-atomically with respect to actions by the tasks and while we still have tracked tasks.

Moreover, with the move of `wake_up_new_task` to the end of array, we're still potentially adding tasks to tracking during unload, even when we've already partially removed other hooks and so are updating our shadow creds only partially (on some code paths, not on all where we should).

We should maybe remove the `wake_up_new_task` hook first, then drop all tasks from tracking, and only then proceed to drop the rest of tasks.

Alternatively, we could set a flag that we're shutting down and inconsistencies are to be ignored, but that flag would be too good of a target for attacking LKRG.

kerneltoast:
@solardiz Thanks for the thorough review!

> I'm afraid I failed to fully understand this part of the commit message:
> 
> ```
> All probes are installed before the initial task dump so that the probe
> handlers see untracked tasks and safely skip them. New tasks forked after
> probe installation are caught by the wake_up_new_task probe, which may
> cause the initial task dump to redundantly add a task, but duplicate add
> attempts are handled gracefully by the ED tree helper so it's fine.
> 
> For the same reason that the initial task dump must come after probe
> installation, the wake_up_new_task probe specifically must be installed
> after all other probes are installed.
> ```
> 
> We want to start tracking all tasks, both those that were already running when we started and those new tasks that are appearing while we're starting up. Is the existence of these two categories of tasks the primary motivation behind the ordering described in the above two paragraphs?

Yes, that's correct with one nitpick: there is technically a third category, which is new tasks that appear after we've fully started up. The third category is the easy case; it's the other two categories that are tricky to deal with.

> The first sentence reads weird to me - surely we're not doing that "so that" or in other words with the goal to have the "handlers see untracked tasks and safely skip them", this is just something we accept would be happening as an alternative to something worse, right? Perhaps the word "that" is extraneous, and only "so" should be left?

True. It's tough to phrase this well. The problem: if we expect a task to always go through probe B after it has gone through probe A, and probe B is not registered by the time the task goes through B, then any stateful tracking started by probe A will remain imbalanced.

IOW, if all of LKRG's probes were stateless then this wouldn't be a problem.

> The second paragraph starts with "For the same reason", but we gave no reason other than via "so that" above... which surely is not the actual reason? Maybe it's in fact "the same" unspecified reason, but stating merely that it's "the same" is not as helpful as actually giving the reason would be.
> 
> I don't mean to nitpick, but this affects my incomplete understanding of what this tries to say, and more importantly of why we're actually doing things in this way.
> 
> What worse would happen in what other (wrong) design, which this description implicitly compares this design against? Is this problem fully avoided by this design, and why?

The "same reason" is that we don't want to start tracking tasks until all probes are registered, for the reason mentioned above.

> I still don't fully understand. Given this:
> 
> ```
>     * [...] The task dump uses task works, which fire when the
>     * target task returns to userspace, so the task is guaranteed to be outside
>     * of any probed code path when it is added to the ED tree.
> ```
> 
> wouldn't it have been safe to dump the existing tasks before probe installation? I'm not saying we should, just that "the same reason" wording in the commit message may be wrong.

No, because we cannot install all probes atomically with respect to those dumped tasks... unless we were to freeze all of those tasks. By "atomically," I mean that tasks will be able to observe some probes without observing others that are yet to be installed, and thus run into "the same reason" why all probes need to be installed before we start tracking any tasks.

> Further, the fourth commit here does:
> 
> ```
>     Now that we have task work infrastructure in place for adding ED tasks, use
>     it even from the wake_up_new_task hook.
> ```
> 
> which seems to make the ordering of hook installation unimportant - but given it's a later separate commit, sure we can't rely on that in the third commit's message yet.

The ordering of hook installation is still important. After the fourth commit, the wake_up_new_task hook **still** only runs once all hooks are installed.

The change to make wake_up_new_task use the task work could be done in the third commit, yes, but it would be pointless at the third commit. It would be deferring the task dump for no benefit at all, as the task dump can just be done directly from inside wake_up_new_task.

This changes at the fourth commit though: the fourth commit introduces a clear incentive to use the task work even when deferral *isn't strictly necessary.* And that incentive: replacing one large GFP_ATOMIC allocation with one tiny GFP_ATOMIC allocation plus a large **no-fail** GFP_KERNEL allocation. A small GFP_ATOMIC allocation is drastically more likely to succeed than a large one, especially when the allocation size is larger than one page. And the large allocation is now guaranteed to never fail, since it's performed in a sleepable context.

> In better news, I think I now understand the task cred corruption false positives we got on LKRG unload with this PR's changes. We no longer freeze on unload, so we're removing hooks non-atomically with respect to actions by the tasks and while we still have tracked tasks.

Oooof, good catch.

> Moreover, with the move of `wake_up_new_task` to the end of array, we're still potentially adding tasks to tracking during unload, even when we've already partially removed other hooks and so are updating our shadow creds only partially (on some code paths, not on all where we should).
>
> We should maybe remove the `wake_up_new_task` hook first, then drop all tasks from tracking, and only then proceed to drop the rest of tasks.

Yes, this sounds like the best approach, with the full version in my mind looking like this:
1. Remove the wake_up_new_task hook
2. Call p_ed_task_work_exit to cancel all queued task works and wait for in-flight ones to finish running
3. Delete every task in the ED task tree under the edtree lock

Hilariously, I asked Claude Fable 5 to simply fetch your comments on this PR and address them, and let it run in the background unsupervised. Then, while I was writing this comment, Claude independently implemented exactly that best approach I had in mind before the idea had even occurred to me.

solardiz:
> We should maybe remove the `wake_up_new_task` hook first, then drop all tasks from tracking, and only then proceed to drop the rest of tasks.

Oh, of course I meant "remove the rest of hooks", not "drop the rest of tasks".

> Yes, this sounds like the best approach, with the full version in my mind looking like this:

Looks good to me. I assume you'll update this PR shortly. Thank you!

kerneltoast:
@solardiz I've pushed what I've got so far. As I mentioned to you before, this is currently incomplete due to material issue findings by GPT-6 Astra. The report from Astra is printed below for you to skim through.

---

Here are six findings from reviewing this series through [c053dce](https://github.com/kerneltoast/lkrg/commit/c053dced4b9b0a9e000793f2c5dcea432630c95d). The snapshot race in item 5 is fixed in that commit; the separate seccomp enrollment race remains open. These are source-level analyses without runtime reproducers. Source links target this tip; links to earlier code identify the history of each issue.

### 1. The task-work counter can reach zero before the callback has returned

**Severity:** major. **Confidence:** high. **Affected:** fully preemptible kernels without a usable `synchronize_rcu_tasks()`.

**Origin: introduced by the task-work conversion.** The ED task-work callback and its counter are new; the [old ED teardown](https://github.com/kerneltoast/lkrg/blob/a357f1a0cd2cc42fdd8cce7858681eaa7b85056d/src/modules/exploit_detection/p_exploit_detection.c#L1923-L1940) had no task-work callback lifetime to drain.

The last task-work callback decrements `p_task_work_cnt` and then still has to check the waitqueue, possibly wake the waiter, and return. Teardown can observe zero before those instructions finish. The additional `synchronize_rcu_tasks()` call is optional. See [p_exploit_detection.c:1087–1090](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/exploit_detection/p_exploit_detection.c#L1087-L1090), [the counter wait at line 1161](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/exploit_detection/p_exploit_detection.c#L1161), and [the optional grace period at lines 1178–1186](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/exploit_detection/p_exploit_detection.c#L1178-L1186).

The failure sequence is: the callback decrements to zero, gets preempted, and remains suspended while teardown completes and LKRG's module memory is freed. When it resumes, it accesses the waitqueue or executes code in freed module memory. Ordinary RCU and scheduler grace periods do not wait for an arbitrary preempted task outside their read-side regions. This also applies to a later initialization failure that takes the same [ED teardown path in p_lkrg_main.c:631](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/p_lkrg_main.c#L629-L634).

Insertion has already finished before the decrement, so this finding concerns callback lifetime. It does not establish insertion after tree detachment. A fix needs either a required Tasks RCU facility on these configurations or another protocol that proves the callback has returned to kernel-owned code. Moving the decrement to the last C statement still leaves its function return unaccounted for.

### 2. The validation gate does not reliably drain validators on older preemptible RCU kernels

**Severity:** major. **Confidence:** high. **Concrete affected case:** x86 Linux 4.19 with preemptible RCU and ftrace-backed kprobes.

**Origin: exposed by removing the freezer.** The replacement validation gate and RCU wait do not fully preserve the old protection. [Userspace used to be frozen before ED teardown](https://github.com/kerneltoast/lkrg/blob/a357f1a0cd2cc42fdd8cce7858681eaa7b85056d/src/p_lkrg_main.c#L725-L746), preventing the legitimate userspace credential changes used in the failure sequence below.

`p_ed_enforce_validation_paranoid()` reads `pint_validate` [at p_exploit_detection.c:1615](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/exploit_detection/p_exploit_detection.c#L1608-L1619), before the walk takes `rcu_read_lock()` [at line 1211](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/exploit_detection/p_exploit_detection.c#L1205-L1229). The enclosing `p_ed_enforce_validation()` also reads the gate without explicit RCU protection. One caller is the [plain `p_scm_send` kprobe at p_scm_send.c:26–32](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/exploit_detection/syscalls/p_scm_send/p_scm_send.c#L26-L32).

On 4.19, the [ftrace kprobe handler disables preemption but leaves interrupts enabled](https://github.com/torvalds/linux/blob/v4.19/arch/x86/kernel/kprobes/ftrace.c#L28). That version's [`synchronize_rcu()` does not cover preemption-disabled regions](https://github.com/torvalds/linux/blob/v4.19/kernel/rcu/tree_plugin.h#L790). A handler can read “enabled,” then a scheduler interrupt can report a quiescent state before it enters the explicit RCU region. Its subsequent walk need not delay teardown's already-started grace period.

Teardown can then [detach a bucket before that walk finishes, while the probes are still installed](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/exploit_detection/p_exploit_detection.c#L2097-L2105). A validator retaining an old entry can compare against a shadow that no longer receives legitimate credential updates: for example, the [setuid hooks find no entry in the live tree](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/exploit_detection/syscalls/p_sys_setuid/p_sys_setuid.c#L28-L51). This permits a false violation and task kill. Unregistering the probes afterward is too late to prevent that comparison.

Take explicit RCU protection **before** reading the shutdown gate and retain it through the whole validation operation, including the current-task check after the paranoid walk. Use `READ_ONCE()`/`WRITE_ONCE()` for the gate. RCU on 5.0 and later also covers preemption-disabled regions, so the specific probe interleaving above should not be generalized to those kernels.

### 3. The module notifier is unregistered after the ED state it uses is destroyed

**Severity:** major. **Confidence:** high. **Affected:** configurations with the OverlayFS override hook, including ordinary 4.7–6.12 kernels; [this handling is disabled on 6.13+](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/exploit_detection/syscalls/override/overlayfs/p_ovl_override_sync/p_ovl_override_sync.h#L29-L53).

**Origin: existing ordering, exposed by removing the freezer.** ED teardown already preceded module-notifier unregistration, but [that interval was inside the userspace freeze](https://github.com/kerneltoast/lkrg/blob/a357f1a0cd2cc42fdd8cce7858681eaa7b85056d/src/p_lkrg_main.c#L725-L758). For ordinary userspace module loads, dropping that freeze exposes the race described below.

Both [normal unload at p_lkrg_main.c:693–694](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/p_lkrg_main.c#L691-L695) and [error unwind at lines 631–633](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/p_lkrg_main.c#L629-L634) tear down ED before unregistering the module notifier.

One failure is a hook installed after teardown has already passed it. A concurrent OverlayFS load enters [`p_verify_module_live()` and sees the hook is OFF](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/kmod/p_kmod_notifier.c#L299-L320), then pauses before installation. ED teardown [skips uninstalling an OFF hook](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/exploit_detection/syscalls/p_install.c#L92-L108). The notifier resumes and installs it. The later notifier unregistration does not repeat ED hook removal, leaving a registered probe pointing into LKRG after unload.

There is also a direct validator lifetime gap on preemptible x86 configurations. The notifier's [`p_lkrg_open_rw()` calls `p_ed_find_validate_current()`](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/wrap/p_struct_wrap.h#L507-L516), which [does not check `pint_validate`](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/exploit_detection/p_exploit_detection.c#L1481-L1491). Its [current-task lookup drops RCU before returning the pointer](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/exploit_detection/ed_task_tree.c#L207-L217). A notifier entering after the task-work Tasks RCU wait can be preempted after that lookup, then resume using an entry that teardown has [freed with the parked roots](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/exploit_detection/ed_task_tree.c#L383-L400). Draining probes does not necessarily drain this normal process-context caller.

Unregister and drain the module notifier before dismantling ED or its hooks. Its [current deregistration function also frees the module database arrays](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/kmod/p_kmod_notifier.c#L372-L383), so split those operations if the arrays must remain available to integrity workers. The direct validator also needs the gate and lifetime protection described in item 2.

### 4. A sysctl write can register integrity notifiers again during unload

**Severity:** major. **Confidence:** high.

**Origin: preexisting bug.** [The old unload path](https://github.com/kerneltoast/lkrg/blob/a357f1a0cd2cc42fdd8cce7858681eaa7b85056d/src/p_lkrg_main.c#L719-L729) also unregistered event notifiers **before** starting the userspace freeze, and removed the sysctls afterward. A writer could re-register a notifier in that earlier window. The later freeze did not undo the registration.

Normal unload [sets `kint_validate = 0`, unregisters event notifiers, and only then unregisters the sysctls](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/p_lkrg_main.c#L664-L669). In that interval, a sysctl writer can change `kint_validate` from 0 to 3. The handler explicitly [registers the notifiers for that transition at p_comm_channel.c:373–374](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/comm_channel/p_comm_channel.c#L364-L380). Draining the sysctl writer afterward does not undo those registrations, and unload has no second notifier-unregistration pass.

Those event sources can still invoke [`p_offload_work()`, which allocates and queues without a shutdown check](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/integrity_timer/p_integrity_timer.c#L102-L108). A [workqueue flush waits for work queued before the flush began](https://github.com/torvalds/linux/blob/v3.10/kernel/workqueue.c#L2550-L2557); it does not close the queue to later submissions. Thus moving [`p_offload_cache_delete()` and its flush](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/integrity_timer/p_integrity_timer.c#L55-L69) ahead of ED teardown does not establish that all integrity workers are gone. A later event can allocate through a destroyed/null cache, and the surviving notifier can eventually call unloaded module code.

There is a route into ED work too: even an integrity worker taking the disabled-integrity branch [still reaches `p_ed_enforce_validation_paranoid()` at p_integrity_timer.c:997](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/integrity_timer/p_integrity_timer.c#L995-L1009). `pint_validate = 0` skips its task walk, but [does not skip `p_ed_validate_globals()`](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/exploit_detection/p_exploit_detection.c#L1614-L1619). With `kint_validate` re-enabled, that function can [allocate and queue ED validation work](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/exploit_detection/p_exploit_detection.c#L1560-L1573) after its cache's final flush/destruction.

Unregister and drain the sysctl interface first, then disable validation, unregister event sources, stop the timer, and flush workers. The [error unwind already removes the sysctls first](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/p_lkrg_main.c#L604-L612), so this particular re-registration window concerns normal unload.

### 5. Seccomp enrollment has two races; only the snapshot race is fixed

**Severity:** major. **Confidence:** high. **Affected:** TSYNC-capable kernels, starting with 3.17.

**Origin: mixed.** The snapshot race already existed in the fork path after the old initial freeze; it is fixed in this tip. The first-entry race is exposed by asynchronous initial enrollment. Successful enrollment under the old systemwide freeze excluded that initial-enrollment overlap.

`p_insert_ed_task()` already [copies an enrolled sibling's TSYNC depth under the sighand lock](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/exploit_detection/p_exploit_detection.c#L998-L1034). That handles enrollment during an outstanding TSYNC **when a sibling entry exists to record it**. These are two different cases:

**Snapshot race — fixed in this tip.** In the pre-fix code, A takes its initial snapshot [before calling `p_insert_ed_task()`](https://github.com/kerneltoast/lkrg/blob/5e691de089d85a940545f10e3f15b38a61d59fd1/src/modules/exploit_detection/p_exploit_detection.c#L1031-L1043). Before A acquires the sighand lock and enters the tree, sibling B completes TSYNC and changes A's filter. Both TSYNC handlers miss A because it has no ED entry yet. A then inserts the old snapshot and inherits depth zero from B, enabling validation against stale state. The fix moves `p_update_ed_process()` inside the existing sighand-locked insertion section; allocation remains outside it.

**First-entry race — still open.** During asynchronous initial enrollment:

1. B enters TSYNC while no thread in the group has an ED entry. The [entry handler finds nothing to increment](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c#L55-L62).
2. A's task work inserts the group's first entry while B's TSYNC is still outstanding. No sibling shadow exists to copy, so A starts at depth zero with validation enabled. The snapshot fix does not recover the missing transaction count.
3. B changes A's filter and returns. The [return handler finds A, skips the refresh because its depth is not 1, then decrements 0 to -1](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/modules/exploit_detection/syscalls/p_seccomp/p_seccomp.c#L63-L69).

This can leave a stale snapshot, validation enabled during a legitimate filter change, and an invalid depth for later TSYNC calls. It needs a way to account for a TSYNC that began before the group's first ED entry, or an enrollment protocol that excludes that overlap. Simply refusing to decrement zero would not protect the filter change. This observation does not require an independent global hashtable as the solution.

The attribution matters: [freezing userspace protected the old **initial dump**](https://github.com/kerneltoast/lkrg/blob/a357f1a0cd2cc42fdd8cce7858681eaa7b85056d/src/p_lkrg_main.c#L561-L569), assuming enrollment succeeded. However, the [snapshot-before-lock gap](https://github.com/kerneltoast/lkrg/blob/a357f1a0cd2cc42fdd8cce7858681eaa7b85056d/src/modules/exploit_detection/p_exploit_detection.c#L1032-L1040) also existed in the old [`wake_up_new_task` insertion path](https://github.com/kerneltoast/lkrg/blob/a357f1a0cd2cc42fdd8cce7858681eaa7b85056d/src/modules/exploit_detection/syscalls/p_wake_up_new_task/p_wake_up_new_task.c#L24-L31) after userspace thawed; a running sibling could TSYNC a newborn thread. The first-entry overlap described above is exposed by asynchronously populating an initially untracked group. It should not be conflated with the smaller snapshot bug.

### 6. The iterator fallback does not compile against vanilla 3.10

**Severity:** major. **Confidence:** high.

**Origin: introduced by this series' iterator compatibility change.** [`14bedd6`](https://github.com/kerneltoast/lkrg/commit/14bedd67021f6b7bceb81ad01e93d7f2ceaa3905) adds the broken fallback. This is a compatibility regression separate from task-work execution.

The fallback at [p_lkrg_main.h:104–108](https://github.com/kerneltoast/lkrg/blob/c053dced4b9b0a9e000793f2c5dcea432630c95d/src/p_lkrg_main.h#L104-L108) defines `for_each_process_thread(p, t)` using `for_each_thread(p, t)`. [Upstream 3.10's sched.h](https://github.com/torvalds/linux/blob/v3.10/include/linux/sched.h#L2151) defines neither macro; it provides `do_each_thread`/`while_each_thread` instead. LKRG supplies no fallback for the missing inner macro, so the new loops do not compile against those headers.

The supplied 3.10 tree already had both modern iterators backported. The successful build against that tree therefore did not exercise this fallback and does not establish vanilla 3.10 compatibility.

Use the older iterator pair conditionally, or provide an iterator compatible with the old task-list representation. Copying only the modern macro definition is insufficient when its underlying structure members are also absent.

solardiz:
> 6. The iterator fallback does not compile against vanilla 3.10

This one may not be relevant as we're targeting EL 7 3.10'ish rather than vanilla 3.10 kernels, but the rest look like serious issues needing fixing. Overall, I am not sure where to go from here. Maybe we should prepare and merge a workaround first, which would keep task freezing for now, but would handle that on unload the same way we already do on load (that is, with dynamically adjusted timeouts).

kerneltoast:
> Overall, I am not sure where to go from here. Maybe we should prepare and merge a workaround first, which would keep task freezing for now, but would handle that on unload the same way we already do on load (that is, with dynamically adjusted timeouts).

After thinking about it for several hours, I came up with a plan for seccomp, and I'm now having Astra code it up. It's fairly complex but not too many new lines added, and there's pretty much zero runtime impact after init completes. I believe seccomp was worst of the bunch; will let you know how it goes.

PR #472, opened 2026-04-22
Make the dummy func noinline in case of LTO

kerneltoast:


solardiz (approved):
Thanks. I think I'll merge later today, once CI jobs complete.

PR #477, opened 2026-06-07
Rename local variable "true" to "one" for compatibility with gcc15

### LINKED ISSUE #473: logger.c:110:13: error: expected identifier or '(' before 'true'
Build failure on gcc15.
```
gcc -pipe -frecord-gcc-switches -Wall -g -O2 -flto=auto  -c logger.c
logger.c: In function 'main':
logger.c:110:13: error: expected identifier or '(' before 'true'
  110 |         int true = 1;
      |             ^~~~
logger.c:134:21: error: lvalue required as unary '&' operand
  134 |             (void *)&true, sizeof(true)))
      |                     ^
make: *** [Makefile:40: logger.o] Error 1
```

kotopesutility:
Closes #473

### Description
Just local variable replacement

### How Has This Been Tested?
Compiled it with gcc-15, then ```make```

### DIFF
--- a/logger/logger.c
+++ b/logger/logger.c
@@ -107,7 +107,7 @@ static int drop_root(void)
 
 int main(int argc, const char * const *argv)
 {
-	int true = 1;
+	int one = 1;
 	int sock, new;
 	struct sockaddr_in addr;
 	socklen_t addrlen;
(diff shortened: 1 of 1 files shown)


solardiz (approved):
We'd normally prefix the commit title with subsystem name, so "Logger: " in this case, but this isn't very important.

kotopesutility:
> We'd normally prefix the commit title with subsystem name, so "Logger: " in this case, but this isn't very important.

Sure, no problem!

PR #481, opened 2026-06-23
Fix build for Linux 7.2+

wladmis:
Linux Kernel 7.2 has removed `strncpy`, so it cannot be used.

### DIFF
--- a/src/modules/exploit_detection/p_exploit_detection.c
+++ b/src/modules/exploit_detection/p_exploit_detection.c
@@ -449,8 +449,12 @@ void p_update_ed_process(struct p_ed_process *p_source, char p_stack) {
    p_dump_addr_limit(&p_source->p_ed_task.p_addr_limit, p_task);
 #endif
    /* Name */
+#if LINUX_VERSION_CODE < KERNEL_VERSION(7,2,0)
    strncpy(p_source->p_ed_task.p_comm, p_task->comm, TASK_COMM_LEN);
    p_source->p_ed_task.p_comm[TASK_COMM_LEN] = 0;
+#else
+   strscpy(p_source->p_ed_task.p_comm, p_task->comm, TASK_COMM_LEN + 1);
+#endif
    rcu_read_unlock();
 }
 


solardiz:
Thank you @wladmis! I see you first used `sizeof`, was anything wrong with that?

`strscpy` is documented not to require the size argument at all when the destination is an array, so maybe we should omit?

solardiz:
```c
/**
 * strscpy - Copy a C-string into a sized buffer
 * @dst: Where to copy the string to
 * @src: Where to copy the string from
 * @...: Size of destination buffer (optional)
 *
 * Copy the source string @src, or as much of it as fits, into the
 * destination @dst buffer. The behavior is undefined if the string
 * buffers overlap. The destination @dst buffer is always NUL terminated,
 * unless it's zero-sized.
 *
 * The size argument @... is only required when @dst is not an array, or
 * when the copy needs to be smaller than sizeof(@dst).
 *
 * Preferred to strncpy() since it always returns a valid string, and
 * doesn't unnecessarily force the tail of the destination buffer to be
 * zero padded. If padding is desired please use strscpy_pad().
 *
 * Returns the number of characters copied in @dst (not including the
 * trailing %NUL) or -E2BIG if @size is 0 or the copy from @src was
 * truncated.
 */
#define strscpy(dst, src, ...)  \
        CONCATENATE(__strscpy, COUNT_ARGS(__VA_ARGS__))(dst, src, __VA_ARGS__)
```

solardiz:
```c
/*
 * The 2 argument style can only be used when dst is an array with a
 * known size.
 */
#define __strscpy0(dst, src, ...)       \
        sized_strscpy(dst, src, sizeof(dst) + __must_be_array(dst) +    \
                                __must_be_cstr(dst) + __must_be_cstr(src))
#define __strscpy1(dst, src, size)      \
        sized_strscpy(dst, src, size + __must_be_cstr(dst) + __must_be_cstr(src))
```

So by omitting the size argument, we make it use the `__strscpy0` macro, which would do the `sizeof` on its own.

solardiz:
The 2-argument form was introduced in:

```
commit e6584c3964f2ff76a9fb5a701e4a59997b35e547
Author: Kees Cook <kees@kernel.org>
Date:   Wed Sep 20 12:38:14 2023 -0700

    string: Allow 2-argument strscpy()
    
    Using sizeof(dst) for the "size" argument in strscpy() is the
    overwhelmingly common case. Instead of requiring this everywhere, allow a
    2-argument version to be used that will use the sizeof() internally. There
    are other functions in the kernel with optional arguments[1], so this
    isn't unprecedented, and improves readability. Update and relocate the
    kern-doc for strscpy() too, and drop __HAVE_ARCH_STRSCPY as it is unused.
    
    Adjust ARCH=um build to notice the changed export name, as it doesn't
    do full header includes for the string helpers.
    
    This could additionally let us save a few hundred lines of code:
     1177 files changed, 2455 insertions(+), 3026 deletions(-)
    with a treewide cleanup using Coccinelle:
    
    @needless_arg@
    expression DST, SRC;
    @@
    
            strscpy(DST, SRC
    -, sizeof(DST)
            )
    
    Link: https://elixir.bootlin.com/linux/v6.7/source/include/linux/pci.h#L1517 [1]
    Reviewed-by: Justin Stitt <justinstitt@google.com>
    Cc: Andy Shevchenko <andy@kernel.org>
    Cc: linux-hardening@vger.kernel.org
    Signed-off-by: Kees Cook <keescook@chromium.org>
```

A possible reason not to use it could be if we want to relax the version check such that we use `strscpy` on older kernels as well... but if we also support even older kernels where we can't use `strscpy`, this doesn't help much.

solardiz:
`strscpy` was introduced in:

```
commit 30035e45753b708e7d47a98398500ca005e02b86
Author: Chris Metcalf <cmetcalf@ezchip.com>
Date:   Wed Apr 29 12:52:04 2015 -0400

    string: provide strscpy()
    
    The strscpy() API is intended to be used instead of strlcpy(),
    and instead of most uses of strncpy().
    
    - Unlike strlcpy(), it doesn't read from memory beyond (src + size).
    
    - Unlike strlcpy() or strncpy(), the API provides an easy way to check
      for destination buffer overflow: an -E2BIG error return value.
    
    - The provided implementation is robust in the face of the source
      buffer being asynchronously changed during the copy, unlike the
      current implementation of strlcpy().
    
    - Unlike strncpy(), the destination buffer will be NUL-terminated
      if the string in the source buffer is too long.
    
    - Also unlike strncpy(), the destination buffer will not be updated
      beyond the NUL termination, avoiding strncpy's behavior of zeroing
      the entire tail end of the destination buffer.  (A memset() after
      the strscpy() can be used if this behavior is desired.)
    
    - The implementation should be reasonably performant on all
      platforms since it uses the asm/word-at-a-time.h API rather than
      simple byte copy.  Kernel-to-kernel string copy is not considered
      to be performance critical in any case.
    
    Signed-off-by: Chris Metcalf <cmetcalf@ezchip.com>
```

Maybe this is old enough we could use it even on CentOS 7?

solardiz:
Actually, in this specific place it may be better to do `memcpy` and to reduce the array size to be exactly `TASK_COMM_LEN` (which is 16) as the kernel has it and as defensive programming to zeroize the last element after the `memcpy`. The kernel itself ensures `comm` is NUL-terminated and NUL-padded, so actual string length is at most `TASK_COMM_LEN -  1`.

wladmis:
> Thank you @wladmis! I see you first used `sizeof`, was anything wrong with that?

No, just could not decided what is better. Probably `sizeof(...)` variant is more robust, as I already introduced off-by-one...

> `strscpy` is documented not to require the size argument at all when the destination is an array, so maybe we should omit?

As you have already mentioned, 2 args `strscpy` was introduced recently, but I'll check what the earliest kernel version does support 3 args variant.

P.S. For some reason I didn't receive GH mail notifications about your comments, I need to recheck my settings.

wladmis:
> Actually, in this specific place it may be better to do `memcpy` and to reduce the array size to be exactly `TASK_COMM_LEN` (which is 16) as the kernel has it and as defensive programming to zeroize the last element after the `memcpy`. The kernel itself ensures `comm` is NUL-terminated and NUL-padded, so actual string length is at most `TASK_COMM_LEN - 1`.

Ok. I'll see.

solardiz:
> Ok. I'll see.

Let me know if you'd like me to accept your PR as-is and then I'll make the cleanups including to the header file, or you want to include them right away. This can be viewed as one thing (7.2 build fix along with cleanup) or two separate things (first fix the build using the version check, then cleanup dropping the version check).

solardiz:
The kernel uses `memcpy` like this:

```c
void __set_task_comm(struct task_struct *tsk, const char *buf, bool exec)
{
        size_t len = strnlen(buf, sizeof(tsk->comm) - 1);

        trace_task_rename(tsk, buf);
        memcpy(tsk->comm, buf, len);
        memset(&tsk->comm[len], 0, sizeof(tsk->comm) - len);
        perf_event_comm(tsk, exec);
}
```

but I think we should simply `memcpy` the full size, then zeroize one last byte (fixed offset). Would be faster given the size is just 16.

wladmis:
> > Ok. I'll see.
> 
> Let me know if you'd like me to accept your PR as-is and then I'll make the cleanups including to the header file, or you want to include them right away. This can be viewed as one thing (7.2 build fix along with cleanup) or two separate things (first fix the build using the version check, then cleanup dropping the version check).

Ok, let's accept the PR, and then do the rest!

PR #482, opened 2026-06-24
Use `memcpy()` instead of `str(ns)cpy()`

wladmis:
As Solar Designer mentioned, it's better to use `memcpy()` here instead of `str(ns)cpy()`.


solardiz (approved):
Thanks. I already made and tested an identical commit in my private fork, but let's merge yours.

wladmis:
> Thanks. I already made and tested an identical commit in my private fork, but let's merge yours.

Thank you!

Issue #483, opened 2026-07-01
addr_limit verification false positive: KERNEL_DS baseline for compat-sleeper 32-bit tasks on pre-5.10 kernels


NeelimaGundu:
32-bit processes on x86_64 3.10 kernel that are sleeping in compat syscall wrappers (e.g. compat_sys_nanosleep, this does set_fs(KERNEL_DS) before sleeping) are sampled by the LKRG init walker while holding KERNEL_DS. Recording that as baseline produces DETECT alerts on the next VFS syscalls by these processes after wake. Proposed fix: at first-write only, when a non-kthread reads as KERNEL_DS, store USER_DS instead.

solardiz:
Thank you @NeelimaGundu for reporting this and suggesting a fix!

This makes me wonder, are we also at risk of false positives when a userspace task has KERNEL_DS during our verification? If not, why not?

NeelimaGundu:
The compat wrapper involved (compat_sys_nanosleep) doesn't trigger generic_permission() while KERNEL_DS is set i.e no VFS path lookup happens during those windows, so verify shouldn't fire from them. And from my experiment, tasks caught by the init walker in KDS produce DETECT on their next path-lookup syscall (sys32_stat64 → … → generic_permission) but not during the KDS window itself. Can't rule out some other kernel path doing set_fs(KERNEL_DS); filp_open(...) and tripping verify :( I haven't verified this exhaustively...happy to test any specific path you might have in mind

solardiz:
Thank you! So it appears that we're saved from this other kind of problem by only calling `p_verify_addr_limit` from two specific places, `generic_permission` and `security_bprm_committing_creds`.

I'm fine with us trying your proposed fix. Can you please contribute it via a pull request here?

PR #484, opened 2026-07-05
When a non-kthread reads as KERNEL_DS, store USER_DS instead

NeelimaGundu:
### Description
Only do this during the initial task dump. Otherwise, after an actual corruption of a task's addr_limit to KDS from UDS, log mode creates an infinite loop:  DETECT -> p_ed_kill_task -> p_dump_addr_limit to accept changes(normalises to USER_DS, i.e doesn't accept KERNEL_DS) -> another DETECT(because live is still KDS, whereas UDS is stored ) -> back to p_dump_addr_limit and so on.

### How Has This Been Tested?

Tested on my setup, normalises 4 different tasks that are sleeping in hrtimer_nanosleep(called from compat_sys_nanosleep) and now does not cause a DETECT on LKRG load


src/modules/exploit_detection/p_exploit_detection.c
```diff
@@ -388,12 +388,32 @@ void p_verify_addr_limit(struct p_ed_process *p_orig) {
 
 #if defined(P_VERIFY_ADDR_LIMIT)
 static inline void p_dump_addr_limit(mm_segment_t *p_addr_limit, struct task_struct *p_task) {
+unsigned long p_live = p_get_addr_limit(p_task);
+unsigned long p_user_ds, p_kernel_ds, p_prev;
```
solardiz: Missing indentation.

src/modules/exploit_detection/p_exploit_detection.c
```diff
@@ -388,12 +388,32 @@ void p_verify_addr_limit(struct p_ed_process *p_orig) {
...
+   p_user_ds   = (unsigned long)USER_DS.seg;
+   p_kernel_ds = (unsigned long)KERNEL_DS.seg;
+   p_prev      = (unsigned long)p_addr_limit->seg;
+#elif defined(CONFIG_ARM) || defined(CONFIG_ARM64)
+   p_user_ds   = (unsigned long)USER_DS;
+   p_kernel_ds = (unsigned long)KERNEL_DS;
+   p_prev      = (unsigned long)*p_addr_limit;
+#endif
+
+if (p_prev == 0 && !(p_task->flags & PF_KTHREAD) && p_live == p_kernel_ds) {
```
solardiz: FWIW, we check for (lack of) `p_task->mm` in `p_is_ed_task` to skip tracking kthreads. In this PR, we'd be checking the `PF_KTHREAD` flag for almost the same (perhaps except for obscure cases when a kthread does have an mm). I think this is fine, just mentioning it.

src/modules/exploit_detection/p_exploit_detection.c
```diff
@@ -388,12 +388,32 @@ void p_verify_addr_limit(struct p_ed_process *p_orig) {
...
+   p_print_log(P_LOG_DEBUG,
+      "addr_limit initial baseline normalised KERNEL_DS->USER_DS for pid=%u name=%s state=%ld",
+       p_task->pid, p_task->comm, p_task->state);
```
solardiz: Some CI builds failed because the `state` field doesn't always exist under that name - I think let's simply exclude it from this print.

solardiz (changes requested):
Thanks! I suggested a couple of minor changes. Please amend the commit and force-push (don't add more commits).

solardiz:
> Please amend the commit and force-push (don't add more commits).

and when you do, you can also add `Fixes #483` into the commit message.

src/modules/exploit_detection/p_exploit_detection.c
NeelimaGundu: Added indentation

src/modules/exploit_detection/p_exploit_detection.c
NeelimaGundu: Resolving the comment as no action is required

src/modules/exploit_detection/p_exploit_detection.c
NeelimaGundu: Removed state field.

NeelimaGundu:
@solardiz, I have addressed your comments.

solardiz:
> and when you do, you can also add `Fixes #483` into the commit message.

Oh, I didn't mean putting it into the commit _title_, but rather as a separate line further in the message "body". Like this:

```
pINT: When a non-kthread reads as KERNEL_DS, store USER_DS instead

Fixes #483
```

NeelimaGundu:
> > and when you do, you can also add `Fixes #483` into the commit message.
> 
> Oh, I didn't mean putting it into the commit _title_, but rather as a separate line further in the message "body". Like this:
> 
> ```
> pINT: When a non-kthread reads as KERNEL_DS, store USER_DS instead
> 
> Fixes #483
> ```

Done.

solardiz (approved):

